PR #24, opened 2020-10-24
feature: PnL API and wrapper binary compiler

c9s:


pkg/accounting/pnl/avg_cost.go
```diff
@@ -89,9 +86,9 @@ func (c *AverageCostCalculator) Calculate(symbol string, trades []types.Trade, c
 
 	return &AverageCostPnlReport{
 		Symbol:       symbol,
-		StartTime:    c.StartTime,
 		CurrentPrice: currentPrice,
 		NumTrades:    len(trades),
+		StartTime:    trades[0].Time,
```
gx578007, line 91: Need to check the length of `trades`?
If its length is zero, how about just early returning from this function?

c9s: good point, let me fix it.

PR #41, opened 2020-10-28
feature: support standard indicator sets

c9s:


pkg/bbgo/environment.go
```diff
@@ -139,8 +139,36 @@ func (environ *Environment) Connect(ctx context.Context) error {
...
+		}
+
+		// move market data store dispatch to here, use one callback to dispatch the market data
+		// session.Stream.OnKLineClosed(func(kline types.KLine) { })
+
+		now := time.Now()
+		for symbol := range loadedSymbols {
+			marketDataStore, ok := session.marketDataStores[symbol]
+			if !ok {
+				return errors.Errorf("symbol is not defined")
```
ycdesu, line 155: It would be better to include `symbol` in the error string. That makes debugging much easier.

pkg/bbgo/environment.go
```diff
@@ -139,8 +139,36 @@ func (environ *Environment) Connect(ctx context.Context) error {
...
+			}
+
+			for interval := range types.SupportedIntervals {
+				kLines, err := session.Exchange.QueryKLines(ctx, symbol, interval.String(), types.KLineQueryOptions{
+					EndTime: &now,
```
ycdesu, line 160: Do we really need the pointer of `now`?

pkg/bbgo/environment.go
c9s: make sense, thanks

pkg/bbgo/environment.go
c9s: yes, it's an optional parameter, we use nil to check if it's used.

pkg/bbgo/environment.go
c9s: actually we add the store by the symbol, here we are iterating the loadedSymbols again, so it should be fine.

pkg/indicator/ema.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (inc *EWMA) calculateAndUpdate(kLines []types.KLine) {
+	if len(kLines) < inc.Window {
+		// we can't calculate
+		return
+	}
+
+	var index = len(kLines) - 1
+	var lastK = kLines[index]
+	var multiplier = 2.0 / float64(inc.Window+1)
```
gx578007, line 25: Could you add ref link to this calculation settings?

gx578007 (approved):
Others LGTM

pkg/indicator/sma.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func calculateSMA(kLines []types.KLine) float64 {
+	sum := 0.0
+	length := len(kLines)
+	for _, k := range kLines {
+		sum += k.Close
+	}
+
+	avg := sum / float64(length)
```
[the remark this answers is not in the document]
c9s: fixed

pkg/indicator/ema.go
c9s: sure

PR #44, opened 2020-10-29
feature: add bollinger band indicator

c9s:
see #43 first.


c9s:
@gx578007 please take a look while you're available. I need to merge this first.

PR #45, opened 2020-10-29
strategy: grid trading strategy

c9s:


pkg/strategy/grid/strategy.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	orders, err := orderExecutor.SubmitOrders(context.Background(), submitOrders...)
+	if err != nil {
+		log.WithError(err).Error("submit bid order error")
+		return
+	}
+
+	s.mu.Lock()
+	for i := range orders {
```
gx578007, line 119: how about `for _, order := range orders`

pkg/strategy/grid/strategy.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	orders, err := orderExecutor.SubmitOrders(context.Background(), submitOrders...)
+	if err != nil {
+		log.WithError(err).Error("submit ask order error")
+		return
+	}
+
+	s.mu.Lock()
+	for i := range orders {
```
gx578007, line 166: ditto

pkg/strategy/grid/strategy.go
```diff
@@ -0,0 +1,274 @@
...
+			case types.SideTypeSell:
+				s.activeAskOrders[order.OrderID] = order
+			case types.SideTypeBuy:
+				s.activeBidOrders[order.OrderID] = order
+			}
+		}
+	})
+
+	go func() {
+		ticker := time.NewTicker(10 * time.Second)
```
gx578007, line 246: Could we increase this ticker period to be greater than normal network timeout, like 1 minute?
It's just a rare corner case. If an order update comes late due to network delay, it is likely to place duplicated orders.
However, setting to 1 minute does NOT completely fix this but it becomes much rarer :)

c9s: surely!

pkg/strategy/grid/strategy.go
c9s: it was because it's an iterator variable, i wrote this to avoid side effect

pkg/strategy/grid/strategy.go
c9s: fixed

PR #64, opened 2020-11-26
fix: backtest endTime

RicoToothless:
otherwise would occur error
```
panic: backtest.endTime must be defined
```


c9s (approved):
thanks!

PR #101, opened 2021-02-03
minor: extract SQL generator function of trades table

ycdesu:
hi, I think it's worth testing the SQL generation code. In order to do that, I extract the trading SQL generation logic into separated functions and test them.


c9s (approved):

c9s:
讚! 給我 metamask address 我發 BBG 給你

PR #103, opened 2021-02-03
minor: extract query order sql generator

ycdesu:
Test query order SQL.


pkg/service/order.go
```diff
@@ -58,6 +58,23 @@ type QueryOrdersOptions struct {
 }
 
 func (s *OrderService) Query(options QueryOrdersOptions) ([]AggOrder, error) {
+	sql := queryOrderSQL(options)
```
elena-ann, line 61: how about using `genOrderSQL` here?

elena-ann (approved):

PR #110, opened 2021-02-05
feat: add limit option for pnl command

jnlin:
Usage: `bbgo pnl --exchange max --symbol BTCUSDT --limit 7000`

To calculate profit and loss for specified limit of orders

### DIFF
--- a/pkg/cmd/pnl.go
+++ b/pkg/cmd/pnl.go
@@ -21,7 +21,7 @@ import (
 func init() {
 	PnLCmd.Flags().String("exchange", "", "target exchange")
 	PnLCmd.Flags().String("symbol", "BTCUSDT", "trading symbol")
-	PnLCmd.Flags().Int("limit", 500, "number of orders")
+	PnLCmd.Flags().Int("limit", 500, "number of trades")
 	RootCmd.AddCommand(PnLCmd)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/cmd/pnl.go
```diff
@@ -21,6 +21,7 @@ import (
 func init() {
 	PnLCmd.Flags().String("exchange", "", "target exchange")
 	PnLCmd.Flags().String("symbol", "BTCUSDT", "trading symbol")
+	PnLCmd.Flags().Int("limit", 500, "number of orders")
```
c9s, line 24: it's for trades.

c9s, line 24: I would like to know the purpose of this flag, could we set the default to 0? because we should have all trades to be calculated to get the correct result.

c9s:
Thanks for the contribution, please send me your metamask wallet address, I will transfer 3 bbgo to you.

c9s:
sharing some background for the PnL, the existing PnL needs some improvements. it should handle the position if position is closed, then we should reset the average cost, and start a new round of average cost calculation.

in this case, we can find the latest position = 0 trades, and start calculating PnL from the time frame.

pkg/cmd/pnl.go
jnlin: I think the pnl should be calculated in a specified timeframe. However, the trade count limit is a alternative method to set the timeframe if the user has some knowledge on databases.

Maybe we can improve it to an more user-friendly way.

c9s: ok, sounds great

c9s (approved):

jnlin:
> Thanks for the contribution, please send me your metamask wallet address, I will transfer 3 bbgo to you.

Please send to 0x3682F253B54E465b3Ff01640884251eC2580333A, thanks!

PR #113, opened 2021-02-05
minor: move maxapi.Response to util package

ycdesu:
The `Response` struct is general and could be shared with other exchange implementations, like FTX integration. I move it to util package and add unit test on it.


c9s (approved):
awesome! 👍

PR #126, opened 2021-02-07
fix(max): use global trade side here, not string

jnlin:
in pkg/exchange/max/stream.go:
```
func convertWebSocketTrade(t max.TradeUpdate) (*types.Trade, error) {
        // skip trade ID that is the same. however this should not happen
        var side = toGlobalSideType(t.Side)
```

We should use global side here, not string based side.


c9s (approved):
xddd

c9s:
looks great! thanks!

Issue #109, opened 2021-02-05
add ftx exchange support


c9s:

c9s:
closing this since it's almost completed.

PR #128, opened 2021-02-08
feature: query FTX balances

ycdesu:
ticket: #109

Hello @c9s , I have implemented the `QueryAccountBalances` of FTX exchange, and also extracted some helpers. I have a question about currency conversion (that's why I add a TODO in the comment). I wonder know is there any global currencies string list? I don't know how to convert the FTX currency into the global currencies. If there's a list, I can define a string map in the `convert.go` and do the conversion. Thanks for your reply.

### DIFF
--- a/pkg/exchange/ftx/exchange.go
+++ b/pkg/exchange/ftx/exchange.go
@@ -35,7 +35,7 @@ func NewExchange(key, secret string, subAccount string) *Exchange {
 }
 
 func (e *Exchange) Name() types.ExchangeName {
-	panic("implement me")
+	return types.ExchangeFTX
 }
 
 func (e *Exchange) PlatformFeeCurrency() string {
--- a/pkg/exchange/ftx/exchange.go
+++ b/pkg/exchange/ftx/exchange.go
@@ -1,0 +1,67 @@
+package ftx
+
+import (
+	"context"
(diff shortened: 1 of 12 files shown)


c9s:
we don't define global currency list, just pass the currency name in uppercase, so we don't have to maintain a list.

if they have special formatting, like XBTC something, we can just trim the prefix.

ycdesu:
Ok. Ftx seems return currency in uppercase. I will merge first and convert the currency if necessary in the future.

pkg/exchange/ftx/exchange.go
```diff
@@ -0,0 +1,107 @@
...
+const (
+	restEndpoint       = "https://ftx.com"
+	defaultHTTPTimeout = 15 * time.Second
+)
+
+type Exchange struct {
+	rest *restRequest
+}
+
+func NewExchange(key, secret string, subAccount string) *Exchange {
```
c9s, line 23: is subaccount optional?

ycdesu: Yes. If your credential is created for all accounts, you don't have to specify that. If it's only for subaccount, you just specify it.

c9s (approved):
looks good

PR #134, opened 2021-02-27
feature: balance testing cmd

ycdesu:
Sometimes the document isn't that clear, so it will be better to have a command to test the function. I add a cmd to print your balance of this account. In the future I will add websocket related commands.


cmd/ftx/main.go
```diff
@@ -0,0 +1,14 @@
+package main
+
+import (
+	log "github.com/sirupsen/logrus"
+
+	"github.com/c9s/bbgo/pkg/cmd/ftx"
+)
+
+// We could use this command to test FTX functions.
+func main() {
```
c9s, line 10: how about:

    bbgo balance --session ftx

ycdesu: Good idea. This PR is ready for another review, thanks

ycdesu:
Now we could print the balance by `godotenv -f .env.local go run ./cmd/bbgo balances --session=ftx`. Currently only support ftx.

c9s:
> Now we could print the balance by `godotenv -f .env.local go run ./cmd/bbgo balances --session=ftx`. Currently only support ftx.

actually we have embedded the godotenv in the root command, you can check that.

abd we also provide a builtin persistent option named `--dotenv`

c9s:
by default, bbgo looks for the ".env.local" file, so you can ignore the dotenv command wrapper now.

PR #135, opened 2021-02-27
Feature: base websocket client

ycdesu:
The base client could be embedded to your websocket implementation and manage the message reading and connectivity. Stream callbacks will be added in future pull requests.


pkg/exchange/ftx/exchange.go
```diff
@@ -16,8 +18,11 @@ const (
 	defaultHTTPTimeout = 15 * time.Second
 )
 
+var Logger = logrus.WithField("exchange", "ftx")
```
c9s, line 21: do we need to expose the logger here?

c9s (approved):

pkg/exchange/ftx/exchange.go
ycdesu: oh it's a mistake.

PR #136, opened 2021-02-27
feature: print ftx websocket subscribed message

ycdesu:
It's ready to send a subscription to FTX websocket server, and I've printed the subscribed message successfully. Later I will implement the orderbook snapshot first.

### DIFF
--- a/pkg/cmd/orderbook.go
+++ b/pkg/cmd/orderbook.go
@@ -1,0 +1,50 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"syscall"
+
+	"github.com/spf13/cobra"
+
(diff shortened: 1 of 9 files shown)


pkg/cmd/orderbook.go
```diff
@@ -0,0 +1,50 @@
...
+			return err
+		}
+
+		symbol, err := cmd.Flags().GetString("symbol")
+		if err != nil {
+			return fmt.Errorf("can't get the symbol from flags: %w", err)
+		}
+
+		s := ex.NewStream()
+		s.Subscribe(types.BookChannel, symbol, types.SubscribeOptions{})
```
c9s, line 34: see if we can reuse the binance book example here?

c9s (approved):
awesome! 😎😎

PR #140, opened 2021-03-02
fix: fix orderbook valudation logic 

ycdesu:
The original implementation always returns an error even if the orderbook prices are correct.


c9s (approved):

c9s:
thank you!!

PR #141, opened 2021-03-02
feature: emit ftx orderbook snapshot

ycdesu:
The orderbook snapshot is published to the standard stream. Now you could print orderbook snapshot of a specific market by `go run ./cmd/bbgo orderbook --session=ftx --symbol=btc/usdt`.

Note that I haven't implemented the checksum mechanism now. This will be a good improvement and may be contributed by someone who is interested.

screenshot:

image


c9s:
Awesome! can you make the command compatible with binance and max? meaning, use the symbol name in uppercase , thanks!

pkg/cmd/orderbook.go
```diff
@@ -29,13 +30,19 @@ var orderbookCmd = &cobra.Command{
 		if err != nil {
 			return fmt.Errorf("can't get the symbol from flags: %w", err)
 		}
+		if symbol == "" {
+			return fmt.Errorf("symbol is not found")
+		}
 
 		s := ex.NewStream()
 		s.Subscribe(types.BookChannel, symbol, types.SubscribeOptions{})
+		s.OnBookSnapshot(func(book types.OrderBook) {
```
c9s, line 39: you can use environment to load exchange session

ycdesu: @c9s Could you give me an example?

ycdesu:
> Awesome! can you make the command compatible with binance and max? meaning, use the symbol name in uppercase , thanks!

I've modified the comment. In the underlying implementation, I also convert the input symbol to uppercase string.

pkg/cmd/orderbook.go
c9s: oh but that way you will need to fulfill the implementation

c9s (approved):

PR #144, opened 2021-03-03
Remove unused conditions when generating orders

Larry850806:
In the two functions `generateGridSellOrders` and `generateGridBuyOrders`, the conditions `s.LowerPrice <= price` and `price <= s.UpperPrice` are always true. So I think they can be deleted.


c9s:
thanks!

PR #146, opened 2021-03-06
feature: orderbook snapshot checksum

ycdesu:
Verify the orderbook snapshot checksum. However there's a bug of order update checksum. I'm not sure if I have to maintain the whole orderbook or just verify the update message. But it's not an important issue. Let's do that next time.


c9s (approved):

c9s (approved):
this is awesome!

PR #147, opened 2021-03-07
feature: query FTX open orders

ycdesu:
Please verify the model conversion logic. I'm not confident about that. Note that the current open orders will NOT include the trigger orders, like stop limit order. We have to invoker another API to get that.


pkg/cmd/orders.go
```diff
@@ -0,0 +1,65 @@
...
+	// default is open which means we query open orders if you haven't provided args.
+	ValidArgs:    []string{"", "open", "closed"},
+	SilenceUsage: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		ctx := context.Background()
+		session, err := cmd.Flags().GetString("session")
+		if err != nil {
+			return fmt.Errorf("can't get session from flags: %w", err)
+		}
+		ex, err := newExchange(session)
```
c9s, line 26: you can try using the environment to get the exchange later (see pnl command for the example)

c9s (approved):

pkg/cmd/orders.go
ycdesu: Do you mean get credential? If that, ftx not only need key and secret. It may also need your account name if your credential is from subaccount.

PR #149, opened 2021-03-13
feature: place FTX order

ycdesu:
Now we support limit/market orders on FTX. Note that I've only tested limit order.


c9s (approved):
👍👍

pkg/exchange/ftx/convert.go
```diff
@@ -21,10 +21,14 @@ func TrimUpperString(original string) string {
 	return strings.ToUpper(strings.TrimSpace(original))
 }
 
+func TrimLowerString(original string) string {
```
c9s, line 24: make this private or move this function to an util package?

PR #150, opened 2021-03-13
feature: add deposit service and withdraw service for sync

c9s:


migrations/mysql/20210301140656_add_withdraws_table.sql
```diff
@@ -0,0 +1,28 @@
...
+    `network`          VARCHAR(32)     NOT NULL DEFAULT '',
+
+    `amount`           DECIMAL(16, 8)  NOT NULL,
+    `txn_id`           VARCHAR(256)    NOT NULL,
+    `txn_fee`          DECIMAL(16, 8)  NOT NULL DEFAULT 0,
+    `txn_fee_currency` VARCHAR(32)     NOT NULL DEFAULT '',
+    `time`             DATETIME(3)     NOT NULL,
+
+    PRIMARY KEY (`gid`),
+    UNIQUE KEY `txn_id` (`exchange`, `txn_id`)
```
ycdesu, line 21: Is the unique constraint name `txn_id` must be unique to all tables? If it is, we should append a prefix to it to prevent naming collision, such as `withdraw_txn_id`.

ycdesu (commented):
Leave one comment for the sql. The PR is too big to review for me, and I just go through very quickly. It will be better to split it to smaller ones.

migrations/mysql/20210301140656_add_withdraws_table.sql
c9s: the key name is table wide, so it's fine. ;-)

image

c9s:
@ycdesu I might not be albe to split the PR LOL, it's all together

PR #158, opened 2021-03-15
fix: fix FTX configs

ycdesu:
In order to support FTX in config, we have to add FTX to the enabled exchanges list. I also modify the readme and default config path. According to `.gitignore`, the default config path should be `config/bbgo.yaml`. Then user can follow the readme to run bbgo without specifying `--config` explicitly.


README.md
```diff
@@ -1,4 +1,4 @@
-# bbgo
+bbgo
```
c9s, line 1: I think this is the title?

README.md
```diff
@@ -67,13 +67,19 @@ MAX_API_SECRET=
 FTX_API_KEY=
 FTX_API_SECRET=
 # specify it if credentials are for subaccount
-FTX_SUBACCOUNT_NAME=
+FTX_SUBACCOUNT=
```
c9s, line 70: thanks~

c9s: oh it should be `FTX_API_SUBACCOUNT`

pkg/cmd/root.go
```diff
@@ -54,7 +54,7 @@ var RootCmd = &cobra.Command{
 
 func init() {
 	RootCmd.PersistentFlags().Bool("debug", false, "debug flag")
-	RootCmd.PersistentFlags().String("config", "bbgo.yaml", "config file")
+	RootCmd.PersistentFlags().String("config", "config/bbgo.yaml", "config file")
```
c9s, line 57: would prefer `bbgo.yaml` it simplifies the installation :-)

README.md
ycdesu: It should be `FTX_SUBACCOUNT` actually https://github.com/c9s/bbgo/blob/main/pkg/cmd/cmdutil/exchange.go#L51

c9s: oh you are right

c9s: do you think we should the API to the prefix?

c9s (approved):

PR #163, opened 2021-03-16
feature: support grid backup and restore

c9s:


pkg/strategy/grid/strategy.go
```diff
@@ -431,6 +461,21 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
...
 
+		if s.Persistence != nil {
+			log.Infof("backing up active orders...")
+			submitOrders := s.activeOrders.Backup()
+			snapshot := Snapshot{
+				Orders: submitOrders,
+				Position: &s.position,
+			}
+
+			if err := s.Persistence.Save(&snapshot, ID, s.Symbol, "snapshot"); err != nil {
```
ycdesu, line 472: It might be better to pass ctx into the Save and Load method of Persistence.

ycdesu (approved):

pkg/strategy/grid/strategy.go
c9s: let's fix this later ;)

PR #165, opened 2021-03-17
Remove --config argument in balances command

Larry850806:
The current `balances` command needs to be executed with the `--config <config_file>`, but the config file is NOT needed to create the session. So I delete the code which reads the config file. In this way, you can get the balance by running `bbgo balances --session=max`.

### DIFF
--- a/pkg/cmd/balances.go
+++ b/pkg/cmd/balances.go
@@ -43,9 +43,6 @@ var balancesCmd = &cobra.Command{
 		}
 
 		environ := bbgo.NewEnvironment()
-		if err := environ.ConfigureDatabase(ctx); err != nil {
-			return err
-		}
 
 		if err := environ.ConfigureExchangeSessions(userConfig); err != nil {
 			return err
--- a/pkg/cmd/balances.go
+++ b/pkg/cmd/balances.go
@@ -33,12 +33,20 @@ var balancesCmd = &cobra.Command{
 			return errors.New("--config option is required")
 		}
 
-		if _, err := os.Stat(configFile); os.IsNotExist(err) {
-			return err
-		}
-
-		userConfig, err := bbgo.Load(configFile, false)
-		if err != nil {
+		// if config file exists, use the config loaded from the config file.
+		// otherwise, use a empty config object
+		var userConfig *bbgo.Config
+		if _, err := os.Stat(configFile); err == nil {
+			// load successfully
+			userConfig, err = bbgo.Load(configFile, false)
+			if err != nil {
+				return err
+			}
+		} else if os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


c9s (commented):
the config is still needed when multiple sessions are configured.

eg, you can have 2 max exchange sessions configured and 4 binance isolated margin sessions

c9s:
the default sessions are configured in the config when the session section is empty.

however, it's possible to provide a default config skeleton when the config file doesn't exist

Larry850806:
Oh, I see. Thank you for the explanation. It seems that I don't know enough about bbgo.

BTW, will the `balances` command use the database? If it doesn't need to configure the database, I will keep the first commit. Otherwise, I just close this PR.

c9s:
> Oh, I see. Thank you for the explanation. It seems that I don't know enough about bbgo.


here is what we can do:

if config file exists, the use the config loaded from the config file.

if not, use the default config object.

> 
> BTW, will the `balances` command use the database? If it doesn't need to configure the database, I will keep the first commit. Otherwise, I just close this PR.

ConfigureDatabase() checks the environment automatically, it's not required by the balances command.

yes I think you can remove it.

c9s:
i added ConfigureDatabase method for consistency

Larry850806:
I just pushed a new commit, please help me review it. Thanks.

Current approach is to check the result of `os.Stat(configFile)`, if the user provides config, I will use it. Otherwise, I will use an empty config object.

pkg/cmd/balances.go
```diff
@@ -33,19 +33,20 @@ var balancesCmd = &cobra.Command{
...
+		// otherwise, use a empty config object
+		var userConfig *bbgo.Config
 		if _, err := os.Stat(configFile); os.IsNotExist(err) {
-			return err
-		}
-
-		userConfig, err := bbgo.Load(configFile, false)
-		if err != nil {
-			return err
+			log.Info("config file doesn't exist, use the default config object")
```
c9s, line 40: remove this log?

pkg/cmd/balances.go
```diff
@@ -33,19 +33,20 @@ var balancesCmd = &cobra.Command{
...
+			userConfig = &bbgo.Config{}
+		} else {
```
c9s, line 42: `err` here could be other error, inexist is just one kind of error, we should exclude this kind of error.

```
if err == nil { // load succeeded
} else if IsNotExists(err) { // does not exist
} else { // other errors
}
```

Larry850806: You are right, I should take other situation into consideration.

c9s (approved):
awesome!

PR #167, opened 2021-03-17
feature: query closed orders

ycdesu:
The time range precision is second, so I will always -1 second to query next page data. I dedupe by created-at and order ID. Note that if the order status is not `closed`, I will skip the order. Please help me to ensure the logic. Thanks.


c9s (approved):
🅰️

PR #179, opened 2021-03-21
feature: get num of fractional parts

ycdesu:
I will use the util to calculate the price/volume precision for market.


pkg/fixedpoint/convert.go
```diff
@@ -263,3 +263,15 @@ func Max(a, b Value) Value {
 
 	return b
 }
+
+func NumFractionalDigits(a Value) int {
```
c9s, line 267: i checked, it's called number of decimal points

c9s (approved):

PR #181, opened 2021-03-21
feature: support query FTX markets

ycdesu:
Some fields are not supported by FTX, so I assign the zero value to them. Please help me verify each field carefully, thanks.

### DIFF
--- a/pkg/cmd/market.go
+++ b/pkg/cmd/market.go
@@ -1,0 +1,74 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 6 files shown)


pkg/cmd/market.go
```diff
@@ -0,0 +1,72 @@
...
+		if !ok {
+			return fmt.Errorf("session %s not found", sessionName)
+		}
+
+		markets, err := session.Exchange.QueryMarkets(ctx)
+		if err != nil {
+			return err
+		}
+
+		log.Infof("markets: %+v", markets)
```
c9s, line 69: iterate the markets and print the market one by one? so that the format could be easier for reading.

c9s (approved):
awesome 😎

PR #182, opened 2021-03-21
feature: ftx deposit histories

ycdesu:


pkg/cmd/deposit.go
```diff
@@ -0,0 +1,92 @@
...
+		if err != nil {
+			return fmt.Errorf("can't get the asset from flags: %w", err)
+		}
+		if asset == "" {
+			return fmt.Errorf("asset is not found")
+		}
+
+		until := time.Now()
+		since := until.Add(-7 * 24 * time.Hour)
+		histories, err := session.Exchange.QueryDepositHistory(ctx, asset, since, until)
```
c9s, line 82: just to make sure the deposit history is in the ascending order.

pkg/cmd/deposit.go
```diff
@@ -0,0 +1,92 @@
...
+		if err != nil {
+			return err
+		}
+
+		for _, h := range histories {
```
c9s, line 87: FYI, the transfer-history command also collect deposits and withdraws

pkg/exchange/ftx/exchange.go
```diff
@@ -149,7 +149,24 @@ func (e *Exchange) QueryTrades(ctx context.Context, symbol string, options *type
...
+	resp, err := e.newRest().DepositHistory(ctx)
+	if err != nil {
+		return nil, err
+	}
+	for _, r := range resp.Result {
+		d, err := toGlobalDeposit(r)
+		if err != nil {
+			return nil, err
+		}
+		if d.Asset == asset && !since.After(d.Time.Time()) && !until.Before(d.Time.Time()) {
```
c9s, line 165: filter out the deposits for the duplicated tx ID here? check the Binance QueryDepositHistory?

pkg/types/exchange.go
```diff
@@ -65,6 +65,8 @@ type Exchange interface {
 	ExchangeMarketDataService
 
 	ExchangeTradingService
+
+	ExchangeTransferService
```
c9s, line 69: I want to make the ExchangeTransferService interface as an optional interface, that way we can add new exchange faster and easier, what do you think?

check the `transfer-history` command to see that implementation checking?

c9s (approved):
others LGTM

pkg/exchange/ftx/exchange.go
ycdesu: By the way, the txid, address are optional because the api also returns internal transfers/airdrop histories. I only can dedupe records from chain.

pkg/cmd/deposit.go
ycdesu: ok will sort later

pkg/types/exchange.go
ycdesu: I think you should ask everyone implement all interfaces because bbgo is a really general framework. When you publish a new release and say `bbgo support the new exchange`. Everyone thinks all functions are supported, and they just have to create a new config. Most people may not read all your release notes. They just want you finish the whole job.

PR #185, opened 2021-03-25
feature: query ftx trades

ycdesu:
FTX trades API doesn't support pagination natively, so we have to split the time range into smaller slices. Please help me verify the logic. I hope there's no infinite loop XD


c9s (approved):
looks good! now you can try to sync your trades?

PR #186, opened 2021-03-27
feature: authenticate ftx websocket

ycdesu:
If you don't invoke `SetPublicOnly`,  I will login when connecting. Note that I haven't subscribe orders and trades yet, so the `orderupdate` command doesn't work yet.

### DIFF
--- a/pkg/cmd/orderbook.go
+++ b/pkg/cmd/orderbook.go
@@ -3,11 +3,14 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"os"
 	"syscall"
 
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
+	"github.com/c9s/bbgo/pkg/bbgo"
 	"github.com/c9s/bbgo/pkg/cmd/cmdutil"
 	"github.com/c9s/bbgo/pkg/types"
 )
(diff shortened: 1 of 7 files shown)


pkg/cmd/orderbook.go
```diff
@@ -62,9 +66,76 @@ var orderbookCmd = &cobra.Command{
 	},
 }
 
+// go run ./cmd/bbgo orderupdate
```
c9s, line 69: do we really need this command? looks like only for testing and development

ycdesu: Yes it's for testing only. Do I have to remove them?

c9s: move to examples?

pkg/exchange/ftx/stream.go
```diff
@@ -28,18 +29,39 @@ func NewStream(key, secret string) *Stream {
 }
 
 func (s *Stream) Connect(ctx context.Context) error {
-	return s.wsService.Connect(ctx)
+	// If it's not public only, let's do the authentication.
+	if atomic.LoadInt32(&s.publicOnly) == 0 {
+		logger.Infof("subscribe private events")
```
c9s, line 34: remove debug log?

pkg/exchange/ftx/stream.go
```diff
@@ -28,18 +29,39 @@ func NewStream(key, secret string) *Stream {
...
 
 func (s *Stream) SetPublicOnly() {
 	atomic.StoreInt32(&s.publicOnly, 1)
 }
 
 func (s *Stream) Subscribe(channel types.Channel, symbol string, _ types.SubscribeOptions) {
-	if err := s.wsService.Subscribe(channel, symbol); err != nil {
-		logger.WithError(err).Errorf("subscribe failed, should never happen")
+	if channel != types.BookChannel {
+		// TODO: return err
```
c9s, line 53: you can panic here, since it's logical exception

c9s (approved):

PR #187, opened 2021-03-29
feature: order and trade update

ycdesu:
:P the testing cmd is still in this PR. I've tested it can receive trade and order update.


c9s (approved):

pkg/cmd/trades.go
```diff
@@ -96,3 +92,77 @@ var tradesCmd = &cobra.Command{
 		return nil
 	},
 }
+
+// go run ./cmd/bbgo tradeupdate --session=ftx
+var tradeUpdateCmd = &cobra.Command{
```
c9s, line 97: how about a `private` command to show both trade updates and order updates? because once your order is executed, you would get these updates (trades, orders, balances) ... etc

PR #188, opened 2021-03-31
ftx: query kline using rest api

ycdesu:
FTX only supports REST Kline api

### DIFF
--- a/pkg/exchange/ftx/convert.go
+++ b/pkg/exchange/ftx/convert.go
@@ -17,7 +17,7 @@ func toGlobalCurrency(original string) string {
 }
 
 func toGlobalSymbol(original string) string {
-	return TrimUpperString(original)
+	return strings.ReplaceAll(TrimUpperString(original), "/", "")
 }
 
 func TrimUpperString(original string) string {
@@ -133,7 +133,7 @@ func toGlobalTrade(f fill) (types.Trade, error) {
 func toGlobalKLine(symbol string, interval types.Interval, h Candle) (types.KLine, error) {
 	return types.KLine{
 		Exchange:  types.ExchangeFTX.String(),
-		Symbol:    symbol,
+		Symbol:    toGlobalSymbol(symbol),
 		StartTime: h.StartTime.Time,
 		EndTime:   h.StartTime.Add(interval.Duration()),
 		Interval:  interval,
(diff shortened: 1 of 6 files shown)


pkg/exchange/ftx/convert.go
```diff
@@ -129,3 +129,19 @@ func toGlobalTrade(f fill) (types.Trade, error) {
 		IsIsolated:    false,
 	}, nil
 }
+
+func toGlobalKLine(symbol string, interval types.Interval, h Candle) (types.KLine, error) {
+	return types.KLine{
+		Exchange:  types.ExchangeFTX.String(),
+		Symbol:    symbol,
```
c9s, line 136: ToGlobalSymbol? because these klines will be stored in the db

pkg/exchange/ftx/exchange.go
```diff
@@ -142,7 +142,51 @@ func (e *Exchange) QueryAccountBalances(ctx context.Context) (types.BalanceMap,
...
+		if err != nil {
+			return nil, err
+		}
+		kline = append(kline, globalKline)
+	}
+	return kline, nil
+}
+
+func isIntervalSupportedInKLine(interval types.Interval) bool {
+	_, ok := map[int]struct{}{
```
c9s, line 180: pull out the predefined map?

pkg/exchange/ftx/exchange.go
```diff
@@ -142,7 +142,51 @@ func (e *Exchange) QueryAccountBalances(ctx context.Context) (types.BalanceMap,
...
+		15:    {},
```
c9s, line 181: we need at least 1m and 5m for back testing

c9s (approved):

pkg/exchange/ftx/convert.go
ycdesu: How to convert to global symbol, such as btc-perp?

pkg/exchange/ftx/exchange.go
ycdesu: Yes it's supported by ftx.

pkg/exchange/ftx/convert.go
c9s: perp is for future, we don't support futures for now. will need the design for futures

c9s: simply convert BTC/USDT to BTCUSDT ?

ycdesu: done

c9s (approved):

PR #189, opened 2021-04-02
Add validation interface to strategy

### LINKED ISSUE #161: add parameter validation interface to strategy
strategies need to validate the given parameters from the YAML config, currently we can only do this in the Run method. we can probably add a new method called `Validate` for strategy developer to validate the inputs:

```
func (s *Strategy) Validate() error {
   // check
}
```

LarryLuTW:
The PR adds a `Validator` interface for validating the given parameters in the YAML configuration. And I also moved some code from `Run()` method to `Validate()`.

Resolve #161


c9s (approved):
awesome!

PR #190, opened 2021-04-03
feat: allow telegram bot send to group

jnlin:
This patch allows telegram bot to send message to a telegram group, to aggregate all notification in 1 place

### DIFF
--- a/pkg/notifier/telegramnotifier/interaction.go
+++ b/pkg/notifier/telegramnotifier/interaction.go
@@ -15,12 +15,14 @@ var log = logrus.WithField("service", "telegram")
 
 type Session struct {
 	Owner              *telebot.User `json:"owner"`
+	Chat               *telebot.Chat `json:"chat"`
 	OneTimePasswordKey *otp.Key      `json:"otpKey"`
 }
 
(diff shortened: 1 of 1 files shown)


pkg/notifier/telegramnotifier/interaction.go
```diff
@@ -60,14 +62,14 @@ func (it *Interaction) Session() *Session {
 }
 
 func (it *Interaction) HandleInfo(m *telebot.Message) {
-	if it.session.Owner == nil {
+	if it.session.Owner == nil || it.session.Chat == nil {
 		return
 	}
 
 	if m.Sender.ID != it.session.Owner.ID {
```
c9s, line 67: you need to check Owner == nil here i guess?

c9s (approved):
awesome!

pkg/notifier/telegramnotifier/interaction.go
jnlin: Yes. If (Owner == nil OR Chat == nil), then stop handling request

c9s: got it, thanks~

PR #199, opened 2021-04-13
binance: set TimeInForce to GTC by default

aitjcize:
Binance does not allow submitting order without TimeInForce set for
certain order types. Set TimeInforce to GTC (Good-Til-Cancel) by
default.

### DIFF
--- a/pkg/exchange/binance/exchange.go
+++ b/pkg/exchange/binance/exchange.go
@@ -538,6 +538,11 @@ func (e *Exchange) submitMarginOrder(ctx context.Context, order types.SubmitOrde
 	if len(order.TimeInForce) > 0 {
 		// TODO: check the TimeInForce value
 		req.TimeInForce(binance.TimeInForceType(order.TimeInForce))
+	} else {
+		switch order.Type {
+		case types.OrderTypeLimit, types.OrderTypeStopLimit:
+			req.TimeInForce(binance.TimeInForceTypeGTC)
+		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/exchange/binance/exchange.go
```diff
@@ -538,6 +538,8 @@ func (e *Exchange) submitMarginOrder(ctx context.Context, order types.SubmitOrde
 	if len(order.TimeInForce) > 0 {
 		// TODO: check the TimeInForce value
 		req.TimeInForce(binance.TimeInForceType(order.TimeInForce))
+	} else {
+		req.TimeInForce(binance.TimeInForceTypeGTC)
```
c9s, line 542: TimeInForce is only valid for limit order, you might need to check the order type.

If you set timeInForce for market order, the market order submission will fail.

aitjcize: Thanks, you are right. Fixed.

c9s (approved):
Awesome! thanks!

c9s:
To get some BBG, you can send me your eth address

aitjcize:
ah cool, here is my address: `0x9064ffc8e6b85ed1d96b01976a4875a46c105ef6`

elena-ann:
it’s on the way! with a low wei gas, might need to wait for few days ~

Wei-Ning Huang ***@***.***>於 2021年4月14日 週三，14:45寫道：

PR #201, opened 2021-04-16
migration: extend symbol column length to 10

jnlin:
Originally the length of symbol column in MySQL is 7 (VARCHAR(7)). However, we have the SANDUSDT and LINKUSDT trading pairs, which is longer than 7 characters.


c9s (approved):

c9s:
awesome! thanks!

PR #203, opened 2021-04-21
Extend volume and symbol size for DOGEUSDT in binance

zenixls2:
The volume amount between 4/12 to 4/20 has been increased above the original design of 16 digits. Increase it to 20.

Also symbol length might not be able to hold some new currency pairs like ALICEUSDT or AUCTIONBUSD(11). Increase the length to 12 to be able to hold longer symbol pairs.


zenixls2:
Also upload the forgotten part of code in previous mr.

c9s (approved):

c9s:
thanks!

PR #204, opened 2021-04-21
Fix: add hard start time for binance

zenixls2:
When since becomes time.Time{}, it will do the api call to get data from 1971, which is unreasonable and waste a lot of time in sync command.

Also ran go fmt on several files.


pkg/exchange/binance/exchange.go
```diff
@@ -24,6 +24,8 @@ var log = logrus.WithFields(logrus.Fields{
 	"exchange": "binance",
 })
 
+var binanceStartDate = time.Date(2017, 8, 1, 0, 0, 0, 0, time.UTC)
```
c9s, line 27: we can reuse the launchTime from the binance exchange

zenixls2: Is launch time reasonable? I set this because it was about the time when binance was founded.

c9s: yeah, I mean I have declared the same var, maybe we should use it

zenixls2: Done. Plz check

c9s:
if you're interested in BBG token, please give me your ETH address so that I can send you some BBG

zenixls2:
Eth address: 0x047d65Fca36C06Bdd9Fb2ae809732eC85d525a12

c9s:
sent~ txn : https://etherscan.io/tx/0xd1dfbde2c6b57b56d21bc3dd961b0537475da4b1dcd1bdb284957cdae5c301ad

PR #230, opened 2021-05-11
bollgrid: generate the last order if balance is not enough

LarryLuTW:
In bollgrid strategy, The original behavior is that bbgo will stop generating new orders when the balance is insufficient.

Although the left balance is not much, I think it's wasteful to be idle there. Therefore, I use the remaining balance to send the last order when the balance is insufficient. I hope this will increase the utilization of assets.

### DIFF
--- a/pkg/strategy/bollgrid/strategy.go
+++ b/pkg/strategy/bollgrid/strategy.go
@@ -156,17 +156,20 @@ func (s *Strategy) generateGridBuyOrders(session *bbgo.ExchangeSession) ([]types
 		if price >= currentPrice {
 			continue
 		}
+		// adjust buy quantity using current quote balance
+		quantity := bbgo.AdjustQuantityByMaxAmount(s.Quantity, price, quoteBalance.Float64())
 		order := types.SubmitOrder{
 			Symbol:      s.Symbol,
 			Side:        types.SideTypeBuy,
 			Type:        types.OrderTypeLimit,
 			Market:      s.Market,
-			Quantity:    s.Quantity,
+			Quantity:    quantity,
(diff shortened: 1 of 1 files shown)


zenixls2:
The total value of the order must be over 10usd in most platforms.  
Be sure to check the remaining balance is above that.

c9s:
btw there is a bbgo.AdjustQuantityXXXX function you can use

pkg/strategy/bollgrid/strategy.go
```diff
@@ -226,7 +228,9 @@ func (s *Strategy) generateGridSellOrders(session *bbgo.ExchangeSession) ([]type
 		}
 		baseQuantity := fixedpoint.NewFromFloat(order.Quantity)
 		if baseBalance < baseQuantity {
-			log.Infof("base balance %f is not enough, stop generating sell orders", baseBalance.Float64())
+			log.Infof("base balance %f is not enough, generating the last sell orders", baseBalance.Float64())
+			order.Quantity = baseBalance.Float64()
```
c9s, line 232: try:
```
quantity = bbgo.AdjustQuantityByMaxAmount(quantity, closePrice, b.Available.Float64())
```

LarryLuTW:
@c9s Please help me review this commit. My current approach is to adjust the quantity before constructing the order. If the adjusted quantity is too small, I will not send the order.

c9s (approved):
looks great! thanks!

c9s:
thank you for your contribution! 😎 if you're interested in BBG, please check the link below:

http://bit.ly/get-your-bbg

PR #242, opened 2021-05-14
add twap order execution and related command

c9s:


c9s:
@ycdesu I changed placeOrder to submit-order in order to make the interface consistent.

PR #246, opened 2021-05-17
fix(ftxExchange): setup a symbol mapping table

jnlin:
ftx uses BTC/USDT symbol styles, however bbgo uses the BTCUSDT style
We setup a mapping table in Markets() to make conversion

### DIFF
--- a/pkg/exchange/ftx/convert.go
+++ b/pkg/exchange/ftx/convert.go
@@ -20,6 +20,10 @@ func toGlobalSymbol(original string) string {
 	return strings.ReplaceAll(TrimUpperString(original), "/", "")
 }
 
+func toLocalSymbol(original string) string {
+	return symbolMap[original]
+}
+
 func TrimUpperString(original string) string {
 	return strings.ToUpper(strings.TrimSpace(original))
 }
--- a/pkg/exchange/ftx/exchange.go
+++ b/pkg/exchange/ftx/exchange.go
@@ -21,6 +21,7 @@ const (
 )
 
 var logger = logrus.WithField("exchange", "ftx")
+var symbolMap map[string]string
 
 type Exchange struct {
 	key, secret  string
@@ -33,6 +34,7 @@ func NewExchange(key, secret string, subAccount string) *Exchange {
 	if err != nil {
 		panic(err)
 	}
+	symbolMap = make(map[string]string)
 	return &Exchange{
 		restEndpoint: u,
 		key:          key,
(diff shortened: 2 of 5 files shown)


pkg/exchange/ftx/exchange.go
```diff
@@ -33,6 +34,7 @@ func NewExchange(key, secret string, subAccount string) *Exchange {
 	if err != nil {
 		panic(err)
 	}
+	ftxExchangeSymbolMap = make(map[string]string)
```
c9s, line 37: how about putting this in the init function?

pkg/exchange/ftx/exchange.go
```diff
@@ -21,6 +21,7 @@ const (
 )
 
 var logger = logrus.WithField("exchange", "ftx")
+var ftxExchangeSymbolMap map[string]string
```
c9s, line 24: we can remove the ftx prefix since it is already in the ftx package

pkg/exchange/ftx/exchange.go
```diff
@@ -302,7 +305,7 @@ func (e *Exchange) SubmitOrders(ctx context.Context, orders ...types.SubmitOrder
 			return createdOrders, fmt.Errorf("unsupported TimeInForce %s. only support GTC", so.TimeInForce)
 		}
 		or, err := e.newRest().PlaceOrder(ctx, PlaceOrderPayload{
-			Market:     TrimUpperString(so.Symbol),
+			Market:     ftxExchangeSymbolMap[TrimUpperString(so.Symbol)],
```
c9s, line 308: put the conversion in the toLocalSymbol function?

c9s (commented):
awesome! thanks!

pkg/exchange/ftx/exchange.go
jnlin: I did not found the init function of ftx exchange, do you have any recommendation?

c9s (approved):
thanks!

pkg/exchange/ftx/exchange.go
c9s: you can add one?

pkg/exchange/ftx/convert.go
```diff
@@ -20,6 +20,10 @@ func toGlobalSymbol(original string) string {
 	return strings.ReplaceAll(TrimUpperString(original), "/", "")
 }
 
+func toLocalSymbol(original string) string {
```
c9s, line 23: what if `toLocalSymbol` is called before QueryMarkets is executed?

pkg/exchange/ftx/exchange.go
```diff
@@ -33,6 +34,7 @@ func NewExchange(key, secret string, subAccount string) *Exchange {
 	if err != nil {
 		panic(err)
 	}
+	symbolMap = make(map[string]string)
```
c9s, line 37: Can be in the `init` function. (go's init)

PR #247, opened 2021-05-18
ftx: handle errors of toLocalSymbol

jnlin:


pkg/exchange/ftx/convert.go
```diff
@@ -21,6 +22,9 @@ func toGlobalSymbol(original string) string {
 }
 
 func toLocalSymbol(original string) string {
+	if "" == symbolMap[original] {
```
c9s, line 25: I wonder if we can use go:generate to generate a static lookup map here

pkg/exchange/ftx/convert.go
```diff
@@ -21,6 +22,9 @@ func toGlobalSymbol(original string) string {
...
+		panic(errors.New("Fatal Error: Can not find FTX's symbol mapping table"))
```
c9s, line 26: errors use lower cases

jnlin: fixed

pkg/exchange/ftx/convert.go
jnlin: We should handle symbols without local symbols (ex. BTC-PERP)

jnlin:
Will send another PR

PR #249, opened 2021-05-21
Implement kline stream and subaccount feature for FTX exchange

jnlin:


pkg/exchange/ftx/stream.go
```diff
@@ -102,15 +117,69 @@ func (s *Stream) SetPublicOnly() {
...
+	}
+}
+
+func (s *Stream) subscribeKLine(symbol string, option types.SubscribeOptions) {
+	interval := types.Interval(option.Interval)
+	if !isIntervalSupportedInKLine(interval) {
+		logger.Errorf("not supported kline interval %s", option.Interval)
+		return
+	}
+
```
c9s, line 155: how about this pattern:

```
ticker := time.NewTicker(interval.Duration()) 
defer ticker.Stop()

for {
    select {
          case <-ctx.Done():
             ..
          case <-ticker.C:
    }

}
```

pkg/exchange/ftx/stream.go
```diff
@@ -102,15 +117,69 @@ func (s *Stream) SetPublicOnly() {
...
+			time.Sleep(time.Second * 2)
+			continue
+		}
+
+		// get the last kline
+		since := time.Now().Add(time.Duration(-1*(interval.Minutes())) * time.Minute)
+		kline, err := s.exchange.QueryKLines(s.ctx, symbol, interval, types.KLineQueryOptions{
+			StartTime: &since,
+		})
+		if err != nil {
```
c9s, line 167: wrap this with retry?

pkg/exchange/ftx/stream.go
```diff
@@ -17,36 +17,48 @@ const endpoint = "wss://ftx.com/ws/"
...
+
+		subs := []websocketRequest{newLoginRequest(s.key, s.secret, time.Now(), s.subAccount)}
 		subs = append(subs, s.subscriptions...)
 		for _, sub := range subs {
 			if err := conn.WriteJSON(sub); err != nil {
 				s.ws.EmitError(fmt.Errorf("failed to send subscription: %+v", sub))
 			}
 		}
 	})
+	go s.handleChannelKlineMessage()
```
c9s, line 61: we should start the go routine when we call Connect, not in the constructor.

pkg/exchange/ftx/stream.go
```diff
@@ -60,6 +72,9 @@ func (s *Stream) Connect(ctx context.Context) error {
 	if err := s.ws.Connect(ctx); err != nil {
 		return err
 	}
+	s.ctx = ctx
+	s.isConnected = true
+	s.EmitStart()
```
c9s, line 77: here you have the context, you can pass the context to the k-line worker

pkg/exchange/ftx/stream.go
```diff
@@ -102,15 +117,69 @@ func (s *Stream) SetPublicOnly() {
...
-	s.addSubscription(websocketRequest{
-		Operation: subscribe,
-		Channel:   orderBookChannel,
-		Market:    toLocalSymbol(TrimUpperString(symbol)),
-	})
+}
+
+func (s *Stream) handleChannelKlineMessage() {
+	for {
+		kline := <-s.klineMessage
```
c9s, line 138: use select { } clause to avoid channel blocking and add context deadline checking here.

pkg/exchange/ftx/stream.go
```diff
@@ -102,15 +117,69 @@ func (s *Stream) SetPublicOnly() {
...
+
+	for {
+		if !s.isConnected {
+			time.Sleep(time.Second * 2)
+			continue
+		}
+
+		// get the last kline
+		since := time.Now().Add(time.Duration(-1*(interval.Minutes())) * time.Minute)
+		kline, err := s.exchange.QueryKLines(s.ctx, symbol, interval, types.KLineQueryOptions{
```
c9s, line 164: not sure if FTX api returns kline that is not closed yet?

jnlin: It is always closed.

pkg/exchange/ftx/stream.go
jnlin: fixed

pkg/exchange/ftx/stream.go
jnlin: fixed

pkg/exchange/ftx/stream.go
jnlin: I found pkg/util/retry.go but can not find anywhere used it. Should we use this or use https://pkg.go.dev/gopkg.in/retry.v1 ?

pkg/exchange/ftx/stream.go
jnlin: The subscribeKLine() worker is not called in the Connect() function, (it is called in Subscribe()). Maybe we can move the query from subscribeKLine() to handleChannelKlineMessage() ?

pkg/exchange/ftx/stream.go
jnlin: fixed

pkg/exchange/ftx/stream.go
jnlin: fixed in b7bf89f

pkg/exchange/ftx/rest.go
```diff
@@ -80,13 +83,23 @@ func (r *restRequest) ReferenceURL(refURL string) *restRequest {
...
+	}
+	refURL, err := url.Parse(u)
 	if err != nil {
 		return nil, err
 	}
+
 	return r.baseURL.ResolveReference(refURL), nil
 }
 
+func (r *restRequest) Id(id string) *restRequest {
```
c9s, line 98: prefer using ID here

pkg/exchange/ftx/stream.go
```diff
@@ -102,15 +117,69 @@ func (s *Stream) SetPublicOnly() {
...
+			klines := getLastKLine(s.exchange, ctx, km.symbol, km.interval)
+			if len(klines) >= 0 {
+				// handle mutiple klines, get the latest one
+				s.EmitKLineClosed(klines[len(klines)-1])
+			}
+		}
+	}
+}
+
+func (s *Stream) subscribeKLine(symbol string, option types.SubscribeOptions) {
```
c9s, line 150: can we rename this method to pollKLines?

pkg/exchange/ftx/stream.go
```diff
@@ -102,15 +117,69 @@ func (s *Stream) SetPublicOnly() {
...
+	}
+
+	// get current kline candle
+	s.km <- klineMessage{symbol: symbol, interval: interval}
+
+	ticker := time.NewTicker(interval.Duration())
+	defer ticker.Stop()
+
+	for {
+		select {
```
c9s, line 164: you need to add: case <-ctx.Done(): here

pkg/exchange/ftx/stream.go
```diff
@@ -102,15 +117,69 @@ func (s *Stream) SetPublicOnly() {
...
+		case <-ticker.C:
+			s.km <- klineMessage{symbol: symbol, interval: interval}
+		}
+	}
+}
+
+func getLastKLine(e *Exchange, ctx context.Context, symbol string, interval types.Interval) []types.KLine {
+	// set since to more 30s ago to avoid getting no kline candle
+	since := time.Now().Add(time.Duration(-1*(interval.Minutes()*60+30)) * time.Second)
```
c9s, line 173: how about using time.Truncate and add -time.Millisecond

pkg/exchange/ftx/stream.go
```diff
@@ -102,15 +117,69 @@ func (s *Stream) SetPublicOnly() {
...
+	if channel == types.BookChannel {
+		s.addSubscription(websocketRequest{
+			Operation: subscribe,
+			Channel:   orderBookChannel,
+			Market:    toLocalSymbol(TrimUpperString(symbol)),
+		})
+
+	} else if channel == types.KLineChannel {
+		// FTX does not support kline channel, do polling
+		go s.subscribeKLine(symbol, option)
```
c9s, line 130: move this to Connect? when Connect is called you can use the subscriptions to start the goroutine

jnlin: I did not get your point. Do you mean start the gorouting in Connect() ?

pkg/exchange/ftx/stream.go
jnlin: fixed

pkg/exchange/ftx/rest.go
jnlin: fixed

pkg/exchange/ftx/stream.go
```diff
@@ -17,35 +17,48 @@ const endpoint = "wss://ftx.com/ws/"
 type Stream struct {
 	*types.StandardStream
 
-	ws *service.WebsocketClientBase
+	ws       *service.WebsocketClientBase
+	km       chan klineMessage
```
c9s, line 21: would prefer using `klineMessageC` or `klineC` here

pkg/exchange/ftx/stream.go
c9s: when user calls Subscribe() we will push the subscription to the subscription array.

we can start the workers when the Connect() method is called.

pkg/exchange/ftx/stream.go
```diff
@@ -102,15 +117,69 @@ func (s *Stream) SetPublicOnly() {
...
+	if channel == types.BookChannel {
+		s.addSubscription(websocketRequest{
+			Operation: subscribe,
+			Channel:   orderBookChannel,
+			Market:    toLocalSymbol(TrimUpperString(symbol)),
+		})
+
+	} else if channel == types.KLineChannel {
+		// FTX does not support kline channel, do polling
+		go s.pollKLines(symbol, option)
```
c9s, line 130: we should still add the kline subscription to the subscription list.

and move `go s.pollKLines(symbol, option)` to the Connect() method.

pkg/exchange/ftx/stream.go
```diff
@@ -102,15 +117,69 @@ func (s *Stream) SetPublicOnly() {
...
 	}
-	s.addSubscription(websocketRequest{
-		Operation: subscribe,
-		Channel:   orderBookChannel,
-		Market:    toLocalSymbol(TrimUpperString(symbol)),
+}
+
+func (s *Stream) handleChannelKlineMessage(ctx context.Context) {
+	for {
+		select {
```
c9s, line 138:
```
case <-ctx.Done():
    return
```

is missing.

pkg/exchange/ftx/stream.go
```diff
@@ -102,15 +117,69 @@ func (s *Stream) SetPublicOnly() {
...
+	// get current kline candle
+	s.km <- klineMessage{symbol: symbol, interval: interval}
+
+	ticker := time.NewTicker(interval.Duration())
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			s.km <- klineMessage{symbol: symbol, interval: interval}
```
c9s, line 166: just curious, why not call `s.EmitKLineClosed` here?

jnlin:
rewrite to add the kline subscription to the subscription list.

pkg/exchange/ftx/websocket_messages.go
```diff
@@ -61,14 +61,15 @@ type loginArgs struct {
 	SubAccount string `json:"subaccount"`
 }
 
-func newLoginRequest(key, secret string, t time.Time) websocketRequest {
+func newLoginRequest(key, secret string, t time.Time, subsaccount string) websocketRequest {
```
c9s, line 64: `subsaccount` -> `subaccount`?

pkg/exchange/ftx/stream.go
```diff
@@ -102,15 +116,77 @@ func (s *Stream) SetPublicOnly() {
...
+		interval := types.Interval(option.Interval)
+		ks := klineSubscription{symbol: symbol, interval: interval}
+		s.klineSubscriptions = append(s.klineSubscriptions, ks)
+	} else {
+		panic("only support book/kline channel now")
 	}
-	s.addSubscription(websocketRequest{
-		Operation: subscribe,
-		Channel:   orderBookChannel,
-		Market:    toLocalSymbol(TrimUpperString(symbol)),
```
c9s, line 112: I wonder why can't we use the above method call to add subscription?

jnlin: The definition of websocketRequest is:
`type websocketRequest struct {
	Operation operation `json:"op"`

	// {'op': 'subscribe', 'channel': 'trades', 'market': 'BTC-PERP'}
	Channel channel `json:"channel,omitempty"`
	Market  string  `json:"market,omitempty"`

	Login loginArgs `json:"args,omitempty"`
}`

I suppose that It should be mapped to the json format of request to websocket. Since the kline is not supported by FTX, I decided not to modify the structure.

Or, the structure can be:

`type websocketRequest struct {
	Operation operation `json:"op"`

	// {'op': 'subscribe', 'channel': 'trades', 'market': 'BTC-PERP'}
	Channel channel `json:"channel,omitempty"`
	Market  string  `json:"market,omitempty"`

	Login loginArgs `json:"args,omitempty"`
+.     Interval types.Interval	
}`

pkg/exchange/ftx/stream.go
jnlin: Because the current timestamp may be `interval+N ms`, and we need klines from the previous time slot, not the current one. The formula should be:

`since := time.Now().Truncate(interval.Duration()).Add(time.Duration(-1*(interval.Minutes()*60*1000)-1) * time.Millisecond)`

I guess it is not simpler than original one

pkg/exchange/ftx/stream.go
c9s: oh you're right, then I have no question :D

c9s (approved):

c9s:
thank you for the contributions!

PR #256, opened 2021-05-26
Fix: nil pointer exception in indicator creation, add stoch util func

zenixls2:
if not ok, `inc` will be `nil`. Should not use `:=` in this case, otherwise the original `inc` will still be `nil` and returned back.  
add stoch indicator creating function.
also fix the error format when depth is not loaded


c9s (approved):
awesome! thanks!

PR #264, opened 2021-06-22
fix max order delete url

LarryLuTW:
The cached URL for deleting max order is `v2/orders/delete`, but it should be `v2/order/delete`.


c9s:
thanks! I have fixed this issue

PR #292, opened 2021-12-04
Bug/288 taker buy base volumn

### LINKED ISSUE #288: backtest sync data fail
Got error messages. QQ

`[0000] FATAL cannot execute command error=Error 1264: Out of range value for column 'taker_buy_base_volume' at row 1
`

Enviroment: 

`bbgo backtest --exchange binance -v --sync --sync-only --sync-from 2021-11-01 --config bbgo.yaml 
`
```
backtest: 
  startTime: "2021-11-25" 
  endTime: "2022-11-30" 
  symbols: 
  - XRPUSDT 
  account: 
    balances: 
      XRP: 0.0 
      BTC: 0.0 
      USDT: 10000.0
```

Tested it will work if i modify the column size manually. (from decimal(16,8) to decimal(32,8) )
`
ALTER TABLE binance_klines CHANGE  taker_buy_base_volume  taker_buy_base_volume   decimal(32,8) NOT NULL

`

tony1223:
1. fix #288 , update the taker_buy_base_volumn.
Since sqlite3 didn't have the precision config, so it's only updating the mysql part.

2. Adding sample reference for the migration section in readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -583,7 +583,7 @@ or
 bash utils/generate-new-migration.sh add_pnl_column
 ```
 
-Be sure to edit both sqlite3 and mysql migration files. ( [Sample] (migrations/mysql/20210531234123_add_kline_taker_buy_columns.sql) )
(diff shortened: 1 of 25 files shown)


c9s:
thanks!

PR #335, opened 2021-12-10
ftx : fix #334 rate limit

### LINKED ISSUE #334: ftx rate limit issue
it shows following error when submit too much order request  in short time.  need to use a ratelimiter to handle this. 

`// POST https://ftx.com/api/orders 429, Success: false, err: Do not send more than 2 orders on this market per 200ms
`

tony1223:
add ratelimit 
see #334 for more details


c9s (approved):
awesome

Issue #343, opened 2021-12-11
Data too long for column 'fee_currency' at row 1


tony1223:
it occusing when I fetch ftx exchange order to db
```

time="2021-12-11T10:33:52+08:00" level=info msg="inserting trade: ftx 2759882393 MATICUSD BUY  price: 2.080000      volume: 40.000000   MAKER 2021-12-11 01:36:13.465512 +0000 +0000"
time="2021-12-11T10:33:52+08:00" level=fatal msg="cannot execute command" error="Error 1406: Data too long for column 'fee_currency' at row 1"

```

PR #346, opened 2021-12-11
db: 343 fee currency length 

tony1223:
PR for #343

### DIFF
--- a/migrations/sqlite3/20211211103657_update_fee_currency_length.sql
+++ b/migrations/sqlite3/20211211103657_update_fee_currency_length.sql
@@ -1,0 +1,8 @@
+-- +up
+-- +begin
+-- +end
+
+-- +down
+
+-- +begin
+-- +end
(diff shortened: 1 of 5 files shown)


migrations/sqlite3/20211211103657_update_fee_currency_length.sql
```diff
@@ -0,0 +1,8 @@
+-- +up
+-- +begin
+-- +end
```
c9s, line 3: remove this file?

pkg/migrations/sqlite3/20211211103657_update_fee_currency_length.go
```diff
@@ -0,0 +1,34 @@
...
+		return err
+	}
+
+	return err
+}
+
+func downUpdateFeeCurrencyLength(ctx context.Context, tx rockhopper.SQLExecutor) (err error) {
+	// This code is executed when the migration is rolled back.
+
+	_, err = tx.ExecContext(ctx, "")
```
c9s, line 28: it might return error here?

c9s (approved):

Issue #318, opened 2021-12-08
pricealert fail on ftx exchange


tony1223:
The pricealert strategy only subcribe OnKLine , and ftx exchange only fire onKlineClosed,so the pricealert wont trigger for the case.

as discussion,  we should also fire a onKline before  when ftx onKlineClosed to fix this.

tony1223:
There's still one more place to fix

PR #353, opened 2021-12-12
ftx: update kline event handling for #318

tony1223:
Last time i only handle the event in the begining, it only works in the first interval. this one fix them all.


c9s (approved):
great!

PR #362, opened 2021-12-13
backtest: fix backtest for ftx

tony1223:


tony1223:
there's a bug in interval , will submit PR later.

PR #380, opened 2021-12-20
backtest : finetune for kline scan logic to prevent hanging for

tony1223:
query


pkg/service/backtest.go
```diff
@@ -211,24 +204,33 @@ func (s *BacktestService) QueryKLinesCh(since, until time.Time, exchange types.E
...
-			close(errC)
-		}()
-		return nil, errC
+		return s.returnError(err)
 	}
 
 	return s.scanRowsCh(rows)
 }
 
+func (s *BacktestService) returnError(err error) (chan types.KLine, chan error) {
```
c9s, line 220: this can be a simple function?

tony1223: ok

c9s (approved):

PR #383, opened 2021-12-21
strategy: implment record strategy

tony1223:


pkg/types/account.go
```diff
@@ -41,25 +41,36 @@ func (b Balance) String() string {
...
-	Total    fixedpoint.Value `json:"total" db:"total"`
-	InUSD    fixedpoint.Value `json:"inUSD" db:"inUSD"`
-	InBTC    fixedpoint.Value `json:"inBTC" db:"inBTC"`
-	Time     time.Time        `json:"time" db:"time"`
+	Currency  string           `json:"currency" db:"currency"`
+	Total     fixedpoint.Value `json:"total" db:"total"`
+	InUSD     fixedpoint.Value `json:"inUSD" db:"inUSD"`
+	InBTC     fixedpoint.Value `json:"inBTC" db:"inBTC"`
+	Time      time.Time        `json:"time" db:"time"`
+	Lock      fixedpoint.Value `json:"lock" db:"lock" `
```
c9s, line 49: to `Locked` ?

c9s:
Would you consider integrating this feature into bbgo as a core feature? so that we can render the NAV (net asset value) chart in the web dashboard.

pkg/types/account.go
tony1223: ok

zenixls2:
@tony1223 @c9s  this won't compile with `-tags dnum`.  
There were new methods used for fixedpoint computation:
- Compare
- Add    (ex: a := a.Add(b))
to make sure that the strategy implementation works with all kinds of decimal lib.

pkg/strategy/record/strategy.go
```diff
@@ -0,0 +1,238 @@
...
+			if tb, ok := totalBalances[b.Currency]; ok {
+				tb.Available += b.Available
+				tb.Locked += b.Locked
+				totalBalances[b.Currency] = tb
+			} else {
+				totalBalances[b.Currency] = b
+			}
+		}
+
+		prices := session.LastPrices()
```
zenixls2, line 119: `session.LastPrices` now returns `map[string]fixedpoint.Value`.  
try to define `lastPrices` of the same type.

c9s:
already integrated. thanks!

Issue #388, opened 2021-12-21
onBookTickerUpdate & bookticker


tony1223:
create a new StandardStream event for booktricker.

c9s:
booktricker? do you mean bookTicker?

c9s:
please add ftx (I know you will do it), max, binance, okex to support this~ thanks~

PR #393, opened 2021-12-22
exchange/okex : implement bookticker for okex  , #388

tony1223:
implement okex bookticker event for #388

### DIFF
--- a/pkg/exchange/okex/parse.go
+++ b/pkg/exchange/okex/parse.go
@@ -60,6 +60,21 @@ type BookData struct {
 	Asks                 []BookEntry
 	MillisecondTimestamp int64
 	Checksum             int
+	channel              string
+}
+
+func (data *BookData) BookTicker() types.BookTicker {
+
+	var askBookData BookEntry = data.Asks[0]
+	var bidBookData BookEntry = data.Bids[0]
+
+	return types.BookTicker{
+		Symbol:   data.Symbol,
+		Buy:      bidBookData.Price,
+		BuySize:  bidBookData.Price,
+		Sell:     askBookData.Price,
+		SellSize: askBookData.Volume,
+	}
 }
 
 func (data *BookData) Book() types.SliceOrderBook {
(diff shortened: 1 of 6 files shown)


pkg/exchange/okex/stream.go
```diff
@@ -343,8 +342,11 @@ func (s *Stream) read(ctx context.Context) {
 					s.EmitEvent(*et)
 
 				case *BookData:
-					s.EmitBookData(*et)
-
+					//there's "books" for 400 depth and books5 for 5 depth
+					if channel != "books5" {
+						s.EmitBookData(*et)
+					}
+					s.EmitBookTickerUpdate(et.BookTicker())
```
c9s, line 349: should EmitBookTickerUpdate only for `books5`?

pkg/exchange/okex/parse.go
```diff
@@ -38,7 +38,7 @@ type WebSocketEvent struct {
 	Arg     interface{} `json:"arg,omitempty"`
 }
 
-func parseEvent(v *fastjson.Value) (*WebSocketEvent, error) {
+func parseEvent(v *fastjson.Value) (*WebSocketEvent, string, error) {
```
c9s, line 41: why returning the event separately here? it's already included in the event object.

pkg/exchange/okex/parse.go
```diff
@@ -306,26 +320,30 @@ func parseOrder(v *fastjson.Value) ([]okexapi.OrderDetails, error) {
...
 }
 
-func parseData(v *fastjson.Value) (interface{}, error) {
+func parseData(v *fastjson.Value) (interface{}, string, error) {
 
 	channel := string(v.GetStringBytes("arg", "channel"))
 
 	switch channel {
+	case "books5":
+		data, err := parseBookData(v)
```
c9s, line 329: I think you can include the channel in the BookData struct? so that we can keep this parse function simple.

c9s:
Hi @tony1223, you made a merge point in your pull request, I think you need to add `--rebase` flag for your `git pull` command.

Please rebase the branch against the main branch again to remove the merge point commit~ thanks!

pkg/exchange/okex/stream.go
tony1223: it should all fulfill the rule of bookticker in my though,but i could change it .

pkg/exchange/okex/parse.go
tony1223: ok

tony1223:
ok will do the rebase ~

pkg/exchange/okex/parse.go
tony1223: i was thinking to make the interface consistency, haha.  i have implemented this in another way now.

c9s (approved):

c9s:
awesome!

PR #432, opened 2022-01-11
ftx: Support LIMIT_MAKER and IOC_LIMIT order type

jessy1092:
According [FTX document](https://docs.ftx.com/#place-order), it should use `postOnly` and `ioc` to support `LIMIT_MAKER` and 
 `IOC_LIMIT `

I create the `toLocalOrderType` method for mapping ftx order.

PS. Still not support `STOP_LIMIT` and `STOP_MARKET`


c9s (approved):
awesome!

PR #437, opened 2022-01-14
indicator: Support difference bandwidth on boll indicator and can dynamic create BOLL

jessy1092:
Could create bollinger band with interval, the window and bandwidth

### DIFF
--- a/pkg/bbgo/session.go
+++ b/pkg/bbgo/session.go
@@ -36,7 +36,7 @@ type StandardIndicatorSet struct {
 	// interval -> window
 	sma        map[types.IntervalWindow]*indicator.SMA
 	ewma       map[types.IntervalWindow]*indicator.EWMA
-	boll       map[types.IntervalWindow]*indicator.BOLL
+	boll       map[types.IntervalWindowBandWidth]*indicator.BOLL
 	stoch      map[types.IntervalWindow]*indicator.STOCH
 	volatility map[types.IntervalWindow]*indicator.VOLATILITY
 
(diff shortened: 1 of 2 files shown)


pkg/bbgo/session.go
```diff
@@ -81,22 +81,37 @@ func NewStandardIndicatorSet(symbol string, store *MarketDataStore) *StandardInd
...
+
+	iwb := types.IntervalWindowBandWidth{IntervalWindow: iw, BandWidth: bandWidth}
+	inc, ok := set.boll[iwb]
+
 	if !ok {
 		inc = &indicator.BOLL{IntervalWindow: iw, K: bandWidth}
+
+		// Bind tmp data store
+		tmpDataStore := NewMarketDataStore("TMP")
+		inc.Bind(tmpDataStore)
```
c9s, line 104: since it's tmp, you don't need to bind it?

pkg/bbgo/session.go
```diff
@@ -81,22 +81,37 @@ func NewStandardIndicatorSet(symbol string, store *MarketDataStore) *StandardInd
...
+
+		// Trigger tmp data store to calculate indicator
+		Klines := set.store.KLineWindows[iw.Interval]
```
c9s, line 107: you can see how did we handle EWMA, I think they are similiar.

c9s:
Hi @jessy1092 could you please take a look? thanks~

pkg/bbgo/session.go
jessy1092: @c9s But EWMA only support `7, 25, 99`, original bollinger band only support 20, 2

I just want bollinger support dynamic interval and band width at run time.

Currently if I want to setup bollinger bands like 12, 1 at [bollmaker strategy](https://github.com/c9s/bbgo/blob/main/pkg/strategy/bollmaker/strategy.go#L511-L512), I need to change code at `NewStandardIndicatorSet` method and compile again.

Because The bollinger is initialized at [NewStandardIndicatorSet](https://github.com/c9s/bbgo/blob/main/pkg/bbgo/session.go#L47) when session setup, and then update kline to calculate the bollinger value. 

When run bollmaker  strategy, I found the session was already initialized, so the bollinger 12, 1 creator [BOLL](https://github.com/c9s/bbgo/blob/main/pkg/bbgo/session.go#L95) cannot calculate value with the old kline data. The dynamic bollinger  12, 1 that do not match 20, 2 have zero value at beginning.

c9s (approved):

Issue #423, opened 2022-01-09
binance query trades more than 24 hours between startTime and endTime


c9s:
Error: <APIError> code=-1127, msg=More than 24 hours between startTime and endTime.
[0000] FATAL cannot execute command error=<APIError> code=-1127, msg=More than 24 hours between startTime and endTime.

the `bbgo trades` command

c9s:
I have fixed this in v1.27.0

PR #444, opened 2022-01-26
fix: #423 

zenixls2:
limit startTime and endTime in query for /api/v3/myTrades endpoint  
may have the same issue in futures and margin endpoint, need further check  
@c9s if you have time, could you check for these two?

### DIFF
--- a/pkg/exchange/binance/exchange.go
+++ b/pkg/exchange/binance/exchange.go
@@ -33,6 +33,7 @@ const BinanceUSBaseURL = "https://api.binance.us"
 const BinanceUSWebSocketURL = "wss://stream.binance.us:9443"
 const WebSocketURL = "wss://stream.binance.com:9443"
 const FuturesWebSocketURL = "wss://fstream.binance.com"
+const ONE_DAY_MILLISEC = 86400000
 
 // 5 per second and a 2 initial bucket
 var orderLimiter = rate.NewLimiter(5, 2)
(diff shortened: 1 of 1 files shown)


pkg/exchange/binance/exchange.go
```diff
@@ -1190,30 +1191,41 @@ func (e *Exchange) QueryTrades(ctx context.Context, symbol string, options *type
...
+			endTime = options.EndTime.UnixNano() / int64(time.Millisecond)
+		}
+
+		for t = startTime; t < endTime; t += ONE_DAY_MILLISEC {
+			req.StartTime(t)
+			et := t + ONE_DAY_MILLISEC
+			if et > endTime {
+				et = endTime
+			}
+			req.EndTime(et)
```
c9s, line 1215: If you used both startTime and endTime, you can not query the time range more than 24 hours.

c9s:
I have fixed this in v1.27.0

zenixls2:
> I have fixed this in v1.27.0

I don't think it's a perfect solution.  
Cli interface should provide same result no matter which exchange it connects to.  
Also the deduction of 1 millisecond seems not necessary

c9s:
> > I have fixed this in v1.27.0
> 
> I don't think it's a perfect solution. Cli interface should provide same result no matter which exchange it connects to. Also the deduction of 1 millisecond seems not necessary

What do you want to fix?

c9s:
You need to consider a situation -- when there is no trade between the given start time and end time, the sync will stop even if you have trade data in the later time range.

c9s:
This is the reason that StartTime and EndTime can not be given at the same time.

pkg/exchange/binance/exchange.go
```diff
@@ -33,6 +33,7 @@ const BinanceUSBaseURL = "https://api.binance.us"
 const BinanceUSWebSocketURL = "wss://stream.binance.us:9443"
 const WebSocketURL = "wss://stream.binance.com:9443"
 const FuturesWebSocketURL = "wss://fstream.binance.com"
+const ONE_DAY_MILLISEC = 86400000
```
c9s, line 36: With the Go convention, for constant please use camel case~

c9s:
add link #423

c9s:
oh! I guess I fixed the QueryClosedOrders but I didn't fix the QueryTrade .....

c9s:
https://binance-docs.github.io/apidocs/spot/en/#account-trade-list-user_data 

looks like the 24 hours limit only happens to the closed orders?

Issue #240, opened 2021-05-14
refactor: convert market minQuantity, tickSize .. etc to fixed point type


c9s:

LarryLuTW:
I also want to do this recently, maybe I can take this issue.

Issue #400, opened 2021-12-24
Profit spread should be bigger than 0


arthurwolf:
I get the following error:

`Error: failed to validate the config: profit spread should bigger than 0`

Or in detail:

[...]

PR #446, opened 2022-01-27
fixedpoint improvement and code fixes

zenixls2:
legacy (low precision to 8)
```bash
goos: linux
goarch: amd64
pkg: github.com/c9s/bbgo/pkg/fixedpoint
cpu: Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
BenchmarkMul/mul-float64-8         	281725761	         4.052 ns/op
BenchmarkMul/mul-float64-large-numbers-8         	295906885	         4.049 ns/op
BenchmarkMul/mul-big-small-numbers-8             	19299628	       104.5 ns/op
BenchmarkMul/mul-big-large-numbers-8             	19619779	        92.10 ns/op
PASS
ok  	github.com/c9s/bbgo/pkg/fixedpoint	7.144s
```

dnum (high precision to 16)
```bash
goos: linux
goarch: amd64
pkg: github.com/c9s/bbgo/pkg/fixedpoint
cpu: Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
BenchmarkMul/mul-float64-8         	71511472	        16.91 ns/op
BenchmarkMul/mul-float64-large-numbers-8         	31729623	        36.73 ns/op
BenchmarkMul/mul-big-small-numbers-8             	14251694	       110.3 ns/op
BenchmarkMul/mul-big-large-numbers-8             	10869277	        96.74 ns/op
PASS
ok  	github.com/c9s/bbgo/pkg/fixedpoint	6.110s
```


zenixls2:
also changed  the behavior of floatpoint presentation.  
use 'g' instead of 'f' to present price for BTC pairs.

c9s:
it's faster because it now uses float64 instead of int64?

zenixls2:
> it's faster because it now uses float64 instead of int64?

Exactly

zenixls2:
update: change float64 in types/kline to fixedpoint.Value

zenixls2:
based on https://github.com/apmckinlay/gsuneido/blob/master/util/dnum/dnum.go
with some minor optimization
```
goos: linux
goarch: amd64
pkg: github.com/c9s/bbgo/pkg/fixedpoint
cpu: Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
BenchmarkMul/mul-float64-8         	69861969	        17.11 ns/op
BenchmarkMul/mul-float64-large-numbers-8         	16960190	        70.20 ns/op
BenchmarkMul/mul-big-small-numbers-8             	14283976	       106.1 ns/op
BenchmarkMul/mul-big-large-numbers-8             	12499324	       111.6 ns/op
PASS
ok  	github.com/c9s/bbgo/pkg/fixedpoint	5.576s
```

zenixls2:
changes:
- add dnum implementation (to build, set -tags=dnum) (without dnum flag, use legacy convert.go implementation)
- add more tests, fix direct operations on fixedpoint.Value. 
- fix accounting bug in backtest (startPrice => lastPrice in finalQuoteAsset)
- add skeleton as a buy-and-hold strategy

pkg/bbgo/session.go
```diff
@@ -173,8 +173,8 @@ type ExchangeSession struct {
 
 	// Withdrawal is used for enabling withdrawal functions
 	Withdrawal   bool             `json:"withdrawal,omitempty" yaml:"withdrawal,omitempty"`
-	MakerFeeRate fixedpoint.Value `json:"makerFeeRate,omitempty" yaml:"makerFeeRate,omitempty"`
-	TakerFeeRate fixedpoint.Value `json:"takerFeeRate,omitempty" yaml:"takerFeeRate,omitempty"`
+	MakerFeeRate fixedpoint.Value `json:"makerFeeRate" yaml:"makerFeeRate"`
+	TakerFeeRate fixedpoint.Value `json:"takerFeeRate" yaml:"takerFeeRate"`
```
zenixls2, line 177: there were some differences on json and yaml for detecting "empty" columns.  
the behavior on primitive type and struct is different on json and yaml. One will be nil while others will be 0.

pkg/fixedpoint/convert.go
```diff
@@ -164,47 +225,22 @@ func (v Value) MarshalJSON() ([]byte, error) {
 }
 
 func (v *Value) UnmarshalJSON(data []byte) error {
-	var a interface{}
-	var err = json.Unmarshal(data, &a)
-	if err != nil {
-		return err
+	if bytes.Compare(data, []byte{'n', 'u', 'l', 'l'}) == 0 {
```
zenixls2, line 228: from FTX

zenixls2:
This pr also resolves #240, #396 and #400   
for #400,  example:  
```bash
go run -tags dnum ./cmd/bbgo backtest --exchange binance --config config/abc.yaml -v 0
```
only workable when using dnum tag.

pkg/fixedpoint/convert.go
c9s: XD

pkg/fixedpoint/convert.go
```diff
@@ -164,47 +225,22 @@ func (v Value) MarshalJSON() ([]byte, error) {
...
-
-	return nil
-}
-
-func Must(v Value, err error) Value {
-	if err != nil {
-		panic(err)
+	var err error
+	if data[0] == '"' {
+		data = data[1 : len(data)-1]
```
c9s, line 238: You can do this

```
var a interface{}
err := json.Unmarshal(data, &a)
...
switch a.(type) {
    case string:
    case int64:
}
```

pkg/strategy/schedule/strategy_back
```diff
@@ -0,0 +1,192 @@
+package schedule
```
c9s, line 1: back? backup file?

pkg/fixedpoint/convert.go
zenixls2: json.Unmarshal will accidentally convert numbers to float32, and lost the precision.  
directly parse from string would be better (well, maybe not in the legacy fixedpoint, just to make sure it doesn't fail)

pkg/strategy/schedule/strategy_back
zenixls2: removed

pkg/bbgo/session.go
c9s: 0 is ok here for the fee rate.

pkg/fixedpoint/convert.go
c9s: json.Unmarshal now cast number to float64, so it should be fine. if we really need high precision, we can write the number in the string form.

pkg/service/trade.go
```diff
@@ -118,7 +118,7 @@ func (s *TradeService) Sync(ctx context.Context, exchange types.Exchange, symbol
 
 		tradeKeys[key] = struct{}{}
 
-		log.Infof("inserting trade: %s %d %s %-4s price: %-13f volume: %-11f %5s %s",
+		log.Infof("inserting trade: %s %d %s %-4s price: %-13v volume: %-11v %5s %s",
```
c9s, line 121: wow, `%-13v` go supports this?!

c9s (approved):
Awesome!

PR #459, opened 2022-03-01
add binance paper trade endpoint

zenixls2:
options:
1. use environment variable
2. define it in config yaml

also, what is the influence of our db data once this option is enabled?  
anyway to separate them?

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,15 @@ bbgo pnl --exchange binance --asset BTC --since "2019-01-01"
 
 ## Advanced Configuration
 
+### Testnet (Paper Trading)
+
+Currently only supports binance testnet.
+To run bbgo in testnet, apply new API keys from [Binance Test Network](https://testnet.binance.vision), and set the following env before you start bbgo:
+```bash
+export PAPER_TRADE=1
+export DISABLE_MARKET_CACHE=1 # the symbols supported in testnet is far less than the mainnet
+```
+
 ### Notification
 
 - [Setting up Telegram notification](./doc/configuration/telegram.md)
--- a/pkg/service/sync.go
+++ b/pkg/service/sync.go
@@ -3,6 +3,8 @@ package service
 import (
 	"context"
 	"errors"
+	"os"
+	"strconv"
 	"time"
 
 	"github.com/c9s/bbgo/pkg/cache"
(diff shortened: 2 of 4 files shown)


zenixls2:
choose to support this feature from environment variable.
ready to be reviewed.

README.md
```diff
@@ -173,6 +173,15 @@ bbgo pnl --exchange binance --asset BTC --since "2019-01-01"
 
 ## Advanced Configuration
 
+### Testnet (Paper Trading)
+
+Currently only supports binance testnet.
+To run bbgo in testnet, apply new API keys from [Binance Test Network](https://testnet.binance.vision), and set the following env before you start bbgo:
+```bash
+export PAPER_TRADE=1
```
c9s, line 181: Add `BINANCE_` as the prefix?

pkg/service/sync.go
```diff
@@ -44,27 +51,28 @@ func (s *SyncService) SyncSessionSymbols(ctx context.Context, exchange types.Exc
 		}
 	}
 
-	log.Infof("syncing %s deposit records...", exchange.Name())
-	if err := s.DepositService.Sync(ctx, exchange); err != nil {
-		if err != ErrNotImplemented {
-			return err
+	if !paperTrade() {
```
c9s, line 54: early return?

```
if paperTrade() {
    return nil
}
```

pkg/service/sync.go
```diff
@@ -23,6 +25,11 @@ type SyncService struct {
 	DepositService  *DepositService
 }
 
+func paperTrade() bool {
+	v, err := strconv.ParseBool(os.Getenv("PAPER_TRADE"))
```
c9s, line 29: you can use something like this:

```
if b, ok := util.GetEnvVarBool("DEBUG_SLACK"); ok { ... }
```

README.md
zenixls2: okex also has paper trade endpoints. keep this to avoid additional modification.

pkg/service/sync.go
zenixls2: done

pkg/service/sync.go
zenixls2: done

c9s (approved):

PR #460, opened 2022-03-01
feature: add multiple exchange support in backtest

zenixls2:


config/bollgrid.yaml
```diff
@@ -52,11 +52,12 @@ backtest:
...
   - BTCUSDT
   account:
-    makerCommission: 15
-    takerCommission: 15
-    balances:
-      BTC: 0.0
-      USDT: 10000.0
+    max:
+      makerCommission: 15
+      takerCommission: 15
```
c9s, line 57: makerCommission and takerCommission is not used anymore, we now use the makerFeeRate and takerFeeRate field in the exchange session struct.

config/bollmaker.yaml
```diff
@@ -16,9 +21,10 @@ backtest:
   symbols:
   - ETHUSDT
   account:
```
c9s, line 23: maybe call this key `sessions` so we can map to the sessions?

pkg/bbgo/testdata/backtest.yaml
```diff
@@ -14,13 +14,14 @@ backtest:
...
-    buyerCommission: 0
-    sellerCommission: 0
-    balances:
-      BTC: 1.0
-      USDT: 5000.0
+    binance:
+      makerCommission: 15
+      takerCommission: 15
+      buyerCommission: 0
+      sellerCommission: 0
```
c9s, line 21: ditto

c9s (approved):

PR #461, opened 2022-03-02
REFACTOR: re-implement ftxapi with requestgen and support query order status API

c9s:


zenixls2:
Re-generate the command doc?
In cmd there's a update-doc

zenixls2:
Also the ftx unit tests are failing

pkg/cmd/balances.go
```diff
@@ -19,41 +17,19 @@ func init() {
...
-			if err != nil {
-				return err
-			}
-		} else if os.IsNotExist(err) {
-			// config file doesn't exist
-			userConfig = &bbgo.Config{}
-		} else {
-			// other error
-			return err
+		if userConfig == nil {
```
zenixls2, line 31: instead of checking here, maybe add the following line in `init()`:
```go
balancesCmd.MarkFlagRequired("config")
```

zenixls2: also according to the logic in `root.go`
if config is given, userConfig won't be nil (if file not exists, will only be an empty `bbgo.Config`)

pkg/cmd/orders.go
```diff
@@ -19,11 +19,60 @@ import (
...
 )
 
+var getOrderCmd = &cobra.Command{
+	Use:          "get-order --session SESSION --order-id ORDER_ID",
+	Short:        "Get order status",
+	SilenceUsage: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		ctx := context.Background()
+
+		if userConfig == nil {
```
zenixls2, line 29: here also, check comments in `balances.go`

pkg/cmd/orders.go
```diff
@@ -39,22 +88,10 @@ var listOrdersCmd = &cobra.Command{
...
-			if err != nil {
-				return err
-			}
-		} else if os.IsNotExist(err) {
-			// config file doesn't exist
-			userConfig = &bbgo.Config{}
-		} else {
-			// other error
-			return err
+		if userConfig == nil {
```
zenixls2, line 91: here also, check comments in balances.go

pkg/exchange/ftx/exchange.go
```diff
@@ -78,7 +82,10 @@ func NewExchange(key, secret string, subAccount string) *Exchange {
 		panic(err)
 	}
 
+	client := ftxapi.NewClient()
+	client.Auth(key, secret, subAccount)
```
zenixls2, line 86: Just an TODO item in the future:  Change the name to `SetAuth` instead of `Auth` will be more descriptive.
Also we should move out the fields to some global struct inside `types` like `types.KeyHolder` or something.

pkg/exchange/ftx/exchange.go
```diff
@@ -510,73 +519,59 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
...
 func (e *Exchange) QueryClosedOrders(ctx context.Context, symbol string, since, until time.Time, lastOrderID uint64) (orders []types.Order, err error) {
-	if until == (time.Time{}) {
-		until = time.Now()
+	symbol = TrimUpperString(symbol)
+
+	req := e.client.NewGetOrderHistoryRequest(toLocalSymbol(symbol))
+
+	if since != (time.Time{}) {
+		req.StartTime(since)
+	} else if until != (time.Time{}) {
```
zenixls2, line 528: can since and util both exist?

pkg/exchange/ftx/exchange.go
```diff
@@ -340,74 +344,55 @@ func isIntervalSupportedInKLine(interval types.Interval) bool {
...
+	req := e.client.NewGetFillsRequest()
+	req.Market(toLocalSymbol(symbol))
 
-	if options.Limit == 1 {
-		// FTX doesn't provide pagination api, so we have to split the since/until time range into small slices, and paginate ourselves.
-		// If the limit is 1, we always get the same data from FTX.
-		return nil, fmt.Errorf("limit can't be 1 which can't be used in pagination")
+	if options.StartTime != nil {
+		req.StartTime(*options.StartTime)
+	} else if options.EndTime != nil {
```
zenixls2, line 355: can StartTime and EndTime both exist?

pkg/cmd/balances.go
c9s: good idea

pkg/exchange/ftx/exchange.go
```diff
@@ -510,73 +519,59 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
...
 	for _, o := range orders {
-		rest := e.newRest()
 		if err := requestLimit.Wait(ctx); err != nil {
 			logrus.WithError(err).Error("rate limit error")
 		}
+
 		if len(o.ClientOrderID) > 0 {
-			if _, err := rest.CancelOrderByClientID(ctx, o.ClientOrderID); err != nil {
+			req := e.client.NewCancelOrderByClientOrderIdRequest(o.ClientOrderID)
+			_, err := req.Do(ctx)
```
zenixls2, line 565:
```go
if _, err := req.Do(ctx); err != nil {
    return err
}
```

pkg/exchange/ftx/exchange.go
```diff
@@ -510,73 +519,59 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
...
+			if err != nil {
+				return err
+			}
+		} else {
+			req := e.client.NewCancelOrderRequest(strconv.FormatUint(o.OrderID, 10))
+			_, err := req.Do(ctx)
```
zenixls2, line 571: same here

pkg/exchange/ftx/exchange.go
c9s: usually no

pkg/exchange/ftx/exchange.go
c9s: usually no

pkg/exchange/ftx/exchange.go
c9s: we can do that later~

PR #462, opened 2022-03-03
refactor: refactor ftx implementation and clean up legacy methods

c9s:


pkg/exchange/ftx/rest_order_request.go
```diff
@@ -1,153 +1,6 @@
...
-	"context"
-	"encoding/json"
-	"fmt"
-	"strconv"
-	"time"
-	"github.com/c9s/bbgo/pkg/fixedpoint"
-)
-
 type orderRequest struct {
 	*restRequest
```
zenixls2, line 4: remove this file? ordeRequest definition here seems redundant

c9s: yeah

pkg/exchange/ftx/exchange.go
```diff
@@ -560,18 +561,26 @@ func (e *Exchange) CancelOrders(ctx context.Context, orders ...types.Order) erro
...
-			_, err := req.Do(ctx)
+			resp, err = req.Do(ctx)
+		}
+
+		if err != nil {
+			return err
+		}
+
+		if !resp.Success {
+			v, err := unmarshalResult(resp.Result)
```
zenixls2, line 579: what about just 
```go
return fmt.Errorf("cancel order failed %s", resp.Result)
```

zenixls2 (approved):

PR #469, opened 2022-03-08
WIP: add heikin ashi implementation

zenixls2:


ruvr:
Cool stuff, since this PR could pave the way for further feature enhancements like support for range bars, renko, etc. They ease strategies that focus purely on price action.

PR #470, opened 2022-03-08
grpc: add exchange field to QueryOrderRequest

narumiruna:


narumiruna:
@c9s do we need both `id` and `client_order_id` fields in cancel/query order request?

narumiruna:
@c9s do we need both `id` and `client_order_id` fields in cancel/query order request?

c9s:
@narumiruna yes, we need both fields, the exchange implementation will choose order id first (if given), and then the client order id.

pkg/pb/bbgo.proto
```diff
@@ -150,8 +150,9 @@ message CancelOrderResponse {
 }
 
 message QueryOrderRequest {
-    int64 id = 1;
-    int64 client_order_id = 2;
```
c9s, line 154: client order id is usually a string.

pkg/pb/bbgo.proto
```diff
@@ -150,8 +150,9 @@ message CancelOrderResponse {
 }
 
 message QueryOrderRequest {
-    int64 id = 1;
```
c9s, line 153: would prefer using string here, because some exchange use UUID (string).

we can parse the integer from the server side.

narumiruna: ok

c9s (approved):

PR #475, opened 2022-03-14
fix support strategy

andycheng123:
do not submit order if current position < market.MinQuantity

### DIFF
--- a/pkg/strategy/support/strategy.go
+++ b/pkg/strategy/support/strategy.go
@@ -207,6 +207,10 @@ func (s *Strategy) Subscribe(session *bbgo.ExchangeSession) {
 	}
 }
 
+func (s *Strategy) CurrentPosition() *types.Position {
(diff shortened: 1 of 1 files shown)


pkg/strategy/support/strategy.go
```diff
@@ -428,7 +428,12 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
 					if err != nil {
 						log.WithError(err).Error("submit profit trailing stop order error")
 					} else {
-						s.trailingStopControl.OrderID = orders.IDs()[0]
+						if len(orders.IDs()) > 0 {
```
c9s, line 431: you can cache the ID list here?

```
ids := orders.IDs()
```

andycheng123: cached

c9s (approved):

PR #484, opened 2022-03-16
backtest: fix order cancel fail when run order cancel on the filled e…

ankion:
Use order cancel in OnFilled() backtest will fail.

log:
[0000] ERROR LocalActiveOrderBook: can not cancel ATOMUSDT order error=cancel order failed, order 10 not found: ORDER backtest 10 ATOMUSDT SELL 0/0.5 @ 29.1 -> NEW


c9s:
seems ok, but why does this change fix this issue?

you should avoid canceling orders inside the OnFilled callback.

c9s (approved):

PR #489, opened 2022-03-18
feature: add market trade subscription in binance

zenixls2:
and update session.lastPrice accordingly.  

in backtest still doesn't work.  
also the field in market trade is very different from private trade. 
some fields in types.Trade is filled with dummy messages.

one missing part in types/exchange.go is added (the kucoin verification)

### DIFF
--- a/pkg/exchange/binance/parse.go
+++ b/pkg/exchange/binance/parse.go
@@ -480,6 +480,7 @@ type MarketTradeEvent struct {
 	TradeId        int64 `json:"t"`
 
 	IsMaker bool `json:"m"`
+	Dummy   bool `json:"M"`
 }
 
 /*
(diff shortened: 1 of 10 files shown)


pkg/exchange/binance/parse.go
```diff
@@ -462,6 +467,60 @@ func parseDepthEvent(val *fastjson.Value) (*DepthEvent, error) {
...
+type MarketTradeEvent struct {
+	EventBase
+	Symbol   string           `json:"s"`
+	Quantity fixedpoint.Value `json:"q"`
+	Price    fixedpoint.Value `json:"p"`
+
+	BuyerOrderId  int64 `json:"b"`
+	SellerOrderId int64 `json:"a"`
+
+	OrderTradeTime int64 `json:"T"`
```
c9s, line 479: I guess maybe you can use `types.MillisecondTimestamp` here? not quiet sure

pkg/exchange/binance/parse.go
```diff
@@ -462,6 +467,60 @@ func parseDepthEvent(val *fastjson.Value) (*DepthEvent, error) {
...
+
+*/
+
+func (e *MarketTradeEvent) Trade() types.Trade {
+	tt := time.Unix(0, e.Time*int64(time.Millisecond))
+	return types.Trade{
+		ID:            uint64(e.TradeId),
+		Exchange:      types.ExchangeBinance,
+		Symbol:        e.Symbol,
+		OrderID:       uint64(e.BuyerOrderId),
```
c9s, line 511: use taker's order ID here?

pkg/exchange/binance/parse.go
```diff
@@ -462,6 +467,60 @@ func parseDepthEvent(val *fastjson.Value) (*DepthEvent, error) {
...
+		Side:          types.SideTypeBoth,
```
c9s, line 512: Both is used for self trade type

zenixls2: shouldn't self trade be `SideTypeSelf`?  
for this case, we are not either buyer or seller, and we don't take our own order.
so should be `SideTypeBoth`.

pkg/exchange/binance/parse.go
zenixls2: int64 is used elsewhere in the parse. Follow the same parsing interface of other events.

c9s: ok

pkg/exchange/binance/parse.go
zenixls2: done

pkg/exchange/binance/parse.go
```diff
@@ -462,6 +467,72 @@ func parseDepthEvent(val *fastjson.Value) (*DepthEvent, error) {
...
+	Quantity fixedpoint.Value `json:"q"`
+	Price    fixedpoint.Value `json:"p"`
+
+	BuyerOrderId  int64 `json:"b"`
+	SellerOrderId int64 `json:"a"`
+
+	OrderTradeTime int64 `json:"T"`
+	TradeId        int64 `json:"t"`
+
+	IsMaker bool `json:"m"`
```
c9s, line 482: you need to define the upper case `M` to fix the parsing.

zenixls2: done

c9s (approved):

PR #498, opened 2022-03-28
indicator: add SMA min() max()

austin362667:
in order to do some calculation which using previous values.
e.g., price normalization

### DIFF
--- a/pkg/indicator/sma.go
+++ b/pkg/indicator/sma.go
@@ -30,6 +30,33 @@ func (inc *SMA) Last() float64 {
 	return inc.Values[len(inc.Values)-1]
 }
 
+func (inc *SMA) Max() float64 {
+	if len(inc.Values) == 0 {
+		return 0.0
+	}
+	max := inc.Values[len(inc.Values)-1]
+	for _, value := range inc.Values {
+		if max < value {
+			max = value
+		}
+	}
+	return max
+}
+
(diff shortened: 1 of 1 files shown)


pkg/indicator/sma.go
```diff
@@ -30,6 +30,33 @@ func (inc *SMA) Last() float64 {
 	return inc.Values[len(inc.Values)-1]
 }
 
+func (inc *SMA) Max() float64 {
+	if len(inc.Values) == 0 {
+		return 0.0
+	}
+	max := inc.Values[len(inc.Values)-1]
```
zenixls2, line 37: Why not pick the first as the default number

pkg/indicator/sma.go
```diff
@@ -30,6 +30,33 @@ func (inc *SMA) Last() float64 {
...
+		}
+	}
+	return max
+}
+
+func (inc *SMA) Min() float64 {
+	if len(inc.Values) == 0 {
+		return 0.0
+	}
+	min := inc.Values[len(inc.Values)-1]
```
zenixls2, line 50: Same here

pkg/indicator/sma.go
austin362667: both are ok
which one is better?

zenixls2: It will be faster if you could pick up the first

PR #504, opened 2022-03-30
update script version

tony1223:


c9s:
conflict, we have 1.29.0, let me close it. but thanks!

PR #505, opened 2022-03-31
feature: add pinescript series interface

zenixls2:
add pinescript series interface and several functions

### DIFF
--- a/pkg/cmd/backtest.go
+++ b/pkg/cmd/backtest.go
@@ -365,7 +365,7 @@ var BacktestCmd = &cobra.Command{
 
 				startPrice, ok := session.StartPrice(symbol)
 				if !ok {
-					return fmt.Errorf("start price not found: %s, %s. Run --sync first", symbol, exchangeName)
+					return fmt.Errorf("start price not found: %s, %s. run --sync first", symbol, exchangeName)
 				}
 
 				lastPrice, ok := session.LastPrice(symbol)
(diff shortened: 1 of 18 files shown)


zenixls2:
Will add the README related to `Series` interface in the next pr.  
Let's merge this pr first.

pkg/cmd/backtest.go
```diff
@@ -373,7 +373,7 @@ var BacktestCmd = &cobra.Command{
 
 				startPrice, ok := session.StartPrice(symbol)
 				if !ok {
-					return fmt.Errorf("start price not found: %s, %s", symbol, exchangeName)
+					return fmt.Errorf("start price not found: %s, %s. Run --sync first", symbol, exchangeName)
```
c9s, line 376: log message are all in lower case~

zenixls2: done

pkg/types/kline.go
```diff
@@ -508,3 +508,96 @@ func (k KLineWindow) SlackAttachment() slack.Attachment {
 }
 
 type KLineCallback func(kline KLine)
+
+type KValueType int
+
+const (
+	kOpenValue KValueType = iota
```
c9s, line 515: if the field is not assigned, the default value will be `kOpenValue` (since it's 0) is that what you want?

c9s (approved):

pkg/types/kline.go
zenixls2: fixed

PR #523, opened 2022-04-14
fix: grpc: fix user data stream subscribe

c9s:


pkg/grpc/server.go
```diff
@@ -25,54 +25,59 @@ type UserDataService struct {
...
+		return fmt.Errorf("session name can not be empty")
+	}
+	
+	session, ok := s.Environ.Session(sessionName)
+	if !ok {
+		return fmt.Errorf("session %s not found", sessionName)
 	}
 
-	<-server.Context().Done()
+	userDataStream := session.Exchange.NewStream()
```
narumiruna, line 39: Use `session.UserDataStream`?

pkg/grpc/server.go
```diff
@@ -25,54 +25,59 @@ type UserDataService struct {
...
+			Channel:  pb.Channel_BALANCE,
+			Event:    pb.Event_UPDATE,
+			Balances: transBalances(session, balances),
+		})
+		if err != nil {
+			log.WithError(err).Errorf("grpc: can not send user data")
+		}
+	}
+	userDataStream.OnBalanceUpdate(balanceHandler)
+	userDataStream.OnBalanceSnapshot(balanceHandler)
```
narumiruna, line 72:
```
	balanceMap, err := session.Exchange.QueryAccountBalances(server.Context())
	if err != nil {
		log.WithError(err).Errorf("grpc: query balance error")
	}

	err = server.Send(
		&pb.UserData{
			Channel:  pb.Channel_BALANCE,
			Event:    pb.Event_SNAPSHOT,
			Balances: transBalances(session, balanceMap),
		},
	)
	if err != nil {
		log.WithError(err).Errorf("grpc: send balance snapshot error")
	}
```

pkg/grpc/server.go
c9s: we need to avoid that, because when this grpc request is closed, we should not close the main user data stream. (we have registered our callbacks)

pkg/grpc/server.go
c9s: so you need an initial balance ? sure

pkg/grpc/server.go
narumiruna: ok

narumiruna: It seems that we have to add `Connect` method after `NewSteam`.
```
userDataStream := session.Exchange.NewStream()
userDataStream.Connect(server.Context())
```

narumiruna (approved):

Issue #509, opened 2022-04-05
Incorrect takerfeerate and makerfeerate calculation in bactesting


vinceyap88:
bollmaker.yaml
—————————-
backtest:
  startTime: "2021-08-01"
  endTime: "2021-08-30"
  symbols:
  - DOTUSDT
  account:
    binance:
      makerFeeRate: **0.001**
      takerFeeRate: **0.001**
      balances:
        DOT: 0.0
        USDT: 400.0

bollmaker.log
-------------
time="2022-04-05T22:26:53+08:00" level=info msg="TRADE backtest DOTUSDT  BUY **11.92803862** @ 16.76721599 amount 199.99999987 **fee 0.00894602** DOT orderID 2 Aug  1 00:00:59.999" session=binance
time="2022-04-05T22:26:53+08:00" level=info msg="TRADE backtest DOTUSDT SELL 11.87943396 @ 16.83581899 amount 199.99999985 fee 0.14999999 USDT orderID 3 Aug  1 00:20:59.999" session=binance

Manual calculation
--------------------
fee 0.00894602/11.92803862= 0.00074999924 is not matched with the setting 0.001

zenixls2:
Maybe you need to set the following in backtest:
```
binance:
      makerCommission: 15
      takerCommission: 15
      buyerCommission: 0
      sellerCommission: 0
      balances:
        BTC: 1.0
        USDT: 5000.0
```
And try again

PR #529, opened 2022-04-15
fix: rename EVWMP to VWEMP, fix backtesting fee

zenixls2:
resolve https://github.com/c9s/bbgo/issues/509  
also try to deduct the fee from trade to prevent backtesting balance issue


zenixls2:
also fix the issue when order price is too extreme. The traded price will be far away from the kline price range.

c9s (approved):

PR #542, opened 2022-04-18
feature: strategy controller

andycheng123:
Users can use callback-style functions to register controller callbacks in strategy


pkg/bbgo/interact.go
```diff
@@ -13,35 +12,9 @@ import (
...
-}
-
-type StrategyController interface {
-	StrategyStatusProvider
-	Suspender
-}
-
-type EmergencyStopper interface {
-	EmergencyStop(ctx context.Context) error
-}
```
c9s, line 40: these interfaces are good, I think we don't need to remove them?

pkg/bbgo/interact.go
```diff
@@ -130,11 +113,11 @@ func (it *CoreInteraction) Commands(i *interact.Interact) {
 	i.PrivateCommand("/position", "Show Position", func(reply interact.Reply) error {
 		// it.trader.exchangeStrategies
 		// send symbol options
-		if strategies, found := it.FilterStrategyByInterface((*PositionReader)(nil)); found {
+		if strategies, found := it.FilterStrategyByCallBack("GetPositionCallback"); found {
```
c9s, line 116: I think the FilterStrategyByInterface method already make sense.

pkg/bbgo/interact.go
andycheng123: I thought callback style is better?

pkg/bbgo/interact.go
andycheng123: but why do we need those interfaces if we want to move to callback style?

pkg/bbgo/strategy_controller.go
```diff
@@ -0,0 +1,119 @@
...
+	EmitGetPosition() (position *types.Position, err error)
+	OnClosePosition(cb func(percentage fixedpoint.Value) error)
+	EmitClosePosition(percentage fixedpoint.Value) error
+	OnSuspend(cb func() error)
+	EmitSuspend() error
+	OnResume(cb func() error)
+	EmitResume() error
+	OnEmergencyStop(cb func() error)
+	EmitEmergencyStop() error
+}
```
c9s, line 119: ideally, in the strategy, we use it like:

c := strategyController{}
c.OnSuspend(func() {  do something before suspend })

and from the interaction, we do:

```
when suspend is called
if the strategy has embedded a Controller type and field
    call Suspend on the controller instance
    the controller instance emit suspend to trigger OnSuspend callbacks
    call suspend
end if
```

pkg/bbgo/strategy_controller.go
```diff
@@ -0,0 +1,119 @@
...
+	Status types.StrategyStatus
+
+	// Callbacks
+	GetStatusCallback     func() types.StrategyStatus
+	GetPositionCallback   func() *types.Position
+	ClosePositionCallback func(percentage fixedpoint.Value) error
+	SuspendCallback       func() error
+	ResumeCallback        func() error
+	EmergencyStopCallback func() error
+}
```
c9s, line 20: OnSuspend, OnResume can be done by the callbackgen.

pkg/bbgo/strategy_controller.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/c9s/bbgo/pkg/fixedpoint"
+	"github.com/c9s/bbgo/pkg/types"
+	"reflect"
+)
+
+type StrategyController struct {
+	Status types.StrategyStatus
+
+	// Callbacks
+	GetStatusCallback     func() types.StrategyStatus
```
c9s, line 14: I think we don't need this GetStatusCallback~ because we don't need to override it or register a callback on it.

pkg/bbgo/interact.go
c9s: nop~ I mean PositionCloser is for closing the position, it's implemented, we don't need a callback for it.

If we have a position object and know how much we want to close, we can close it directly without implementing the method. It's better to share the position closing code between different strategies, since they do the same thing.

pkg/bbgo/interact.go
c9s: we need the interface to filter the strategies, so I think the original implementation is good. it does the interface checking and filtering.

andycheng123: will change it back

pkg/bbgo/strategy_controller.go
andycheng123: done

andycheng123:
mix embeded struct and callback in strategy controller

pkg/bbgo/strategy_controller.go
andycheng123: the controller is simplified now
we dont need to use calbackgen?

c9s: use callbackgen thanks 😊 more consistent

pkg/strategy/support/strategy.go
```diff
@@ -439,7 +390,48 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
 	s.orderExecutor = orderExecutor
 
 	// StrategyController
-	s.status = types.StrategyStatusRunning
+	s.Status = types.StrategyStatusRunning
+
+	s.OnSuspend(func() error {
+		var err error = nil
+		// Cancel all order
+		for _, order := range s.orderStore.Orders() {
```
c9s, line 398: orderStore includes closed orders, I think you need to check the open orders~

pkg/strategy/support/strategy.go
```diff
@@ -439,7 +390,48 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
...
+			err = s.cancelOrder(order.OrderID, ctx, s.orderExecutor)
+		}
+		if err != nil {
+			errMsg := "Not all orders are cancelled! Please check again."
+			log.WithError(err).Errorf(errMsg)
+			s.Notify(errMsg)
+		} else {
+			s.Notify("All orders cancelled.")
```
c9s, line 406: `All {symbol} orders are cancelled.`

pkg/strategy/support/strategy.go
```diff
@@ -439,7 +390,48 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
...
+		} else {
+			log.Infof("%s position is saved.", s.Symbol)
+		}
+
+		return err
+	})
+
+	s.OnEmergencyStop(func() error {
+		// Close 100% position
+		percentage, _ := fixedpoint.NewFromString("100%")
```
c9s, line 421: you can do: fixedpoint.NewFromtFloat(1.0)

pkg/bbgo/interact.go
```diff
@@ -74,6 +56,15 @@ func (it *CoreInteraction) FilterStrategyByInterface(checkInterface interface{})
 	return strategies, found
 }
 
+func (it *CoreInteraction) GetStrategyController(strategy SingleExchangeStrategy) (controller StrategyControllerInterface, found bool) {
```
c9s, line 59: this function does not need the CoreInteraction instance, so you can make it a simple function. (without receiver)

pkg/bbgo/interact.go
```diff
@@ -74,6 +56,15 @@ func (it *CoreInteraction) FilterStrategyByInterface(checkInterface interface{})
 	return strategies, found
```
c9s, line 56: also, make FilterStrategyByInterface as pure simple function and you can pass `it.exchangeStrategies` as an input to it. 

make it a simple filter function, which can be easier for testing.

pkg/bbgo/strategy_controller.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (s *StrategyController) EmitEmergencyStop() error {
+	if s.EmergencyStopCallback != nil {
+		return s.EmergencyStopCallback()
+	} else {
+		return nil
+	}
+}
+
+type StrategyControllerInterface interface {
```
c9s, line 74: In go, we don't use the suffix `Interface` to denote an interface. maybe use other name?

pkg/strategy/support/strategy.go
andycheng123: is there a function to check this, or a function to remove filled and cancelled orders in orderStore?

pkg/bbgo/interact.go
c9s: something like:

sc, ok := toStrategyController(instance)

instead of:

sc, ok := it.GetStrategyController(instance)

pkg/bbgo/strategy_controller.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"github.com/c9s/bbgo/pkg/types"
+)
+
+type StrategyController struct {
+	Status types.StrategyStatus
+
+	// Callbacks
+	SuspendCallback       func() error
```
c9s, line 11: name it as `suspendCallbacks []func() error` then you can use callbackgen

pkg/strategy/support/strategy.go
c9s: Cancelled orders (and not filled) are automatically removed from the order store once you bind it to the stream.

OrderStore is used for tracking order ids for the trades.

You need to call ex.QueryOpenOrders() to get these pending orders (orders still on the book)

c9s: In the bollmaker strategy, we use activeMakerOrders to track these open orders:

	s.activeMakerOrders = bbgo.NewLocalActiveOrderBook(s.Symbol)
	s.activeMakerOrders.BindStream(session.UserDataStream)

pkg/strategy/support/strategy.go
c9s: Because you might run multiple support strategies with different symbols.

pkg/strategy/support/strategy.go
```diff
@@ -439,7 +390,48 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
...
+		err := s.ClosePosition(context.Background(), percentage)
```
c9s, line 422: if we failed to close the position, we may log the error and continue

if err != nil {
    log.WithError(err).Error("failed to close the position)
}

if err := s.Suspend(); err != nil {
  ... 
}

pkg/bbgo/strategy_controller.go
andycheng123: got it

pkg/bbgo/interact.go
andycheng123: this function is no longer needed

pkg/bbgo/interact.go
andycheng123: done
is it because we can use FilterStrategyByInterface independently?

pkg/bbgo/strategy_controller.go
andycheng123: Split this interface into several smaller ones

pkg/bbgo/strategy_controller.go
andycheng123: done

pkg/strategy/support/strategy.go
andycheng123: done

pkg/strategy/support/strategy.go
andycheng123: done

pkg/bbgo/strategy_controller.go
andycheng123: done

pkg/strategy/support/strategy.go
andycheng123: done

pkg/strategy/support/strategy.go
andycheng123: use LocalActiveOrderBook to achieve this

pkg/bbgo/interact.go
```diff
@@ -61,10 +43,10 @@ func NewCoreInteraction(environment *Environment, trader *Trader) *CoreInteracti
 	}
 }
 
-func (it *CoreInteraction) FilterStrategyByInterface(checkInterface interface{}) (strategies []string, found bool) {
+func FilterStrategyByInterface(checkInterface interface{}, exchangeStrategies map[string]SingleExchangeStrategy) (strategies []string, found bool) {
```
c9s, line 46: can you return map[string]SingleExchangeStrategy ? so that we can keep the object in the map.

c9s: you can add another function called `strategySignatures(map[string]SingleExchangeStrategy) []string` to get signatures from the map.

pkg/bbgo/strategy_controller.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+//go:generate callbackgen -type StrategyController -interface
+type StrategyController struct {
+	Status types.StrategyStatus
+
+	// Callbacks
+	SuspendCallbacks       []func()
+	ResumeCallbacks        []func()
+	EmergencyStopCallbacks []func()
```
c9s, line 14: make these fields start with lower case?

pkg/strategy/support/strategy.go
```diff
@@ -439,7 +393,41 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
 	s.orderExecutor = orderExecutor
 
 	// StrategyController
-	s.status = types.StrategyStatusRunning
+	s.Status = types.StrategyStatusRunning
+
+	s.OnSuspend(func() {
+		// Cancel all order
+		if err := s.activeOrders.GracefulCancel(ctx, session.Exchange); err != nil {
+			errMsg := "Not all {s.Symbol} orders are cancelled! Please check again."
```
c9s, line 401: we can use `{s.Symbol}` in go?

pkg/strategy/support/strategy.go
```diff
@@ -439,7 +393,41 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
...
+			log.WithError(err).Errorf(errMsg)
+			s.Notify(errMsg)
+		} else {
+			s.Notify("All {s.Symbol} orders are cancelled.")
```
c9s, line 405: I guess you need to write:

    Notify("All %s orders are cancelled.", s.Symbol)

pkg/strategy/support/strategy.go
andycheng123: oh...
i thought we could
sorry, mislead by previous comments

pkg/bbgo/interact.go
andycheng123: So we use strategy map obj to pass in/out functions for convenience in situations like we want to access the strategy objects. and with strategySignatures() we can easily get the string list of strategies
am i right?

c9s: yes~

andycheng123: done

pkg/bbgo/strategy_controller.go
andycheng123: done

pkg/strategy/support/strategy.go
andycheng123: done

pkg/strategy/support/strategy.go
andycheng123: done

pkg/bbgo/interact.go
```diff
@@ -61,12 +43,21 @@ func NewCoreInteraction(environment *Environment, trader *Trader) *CoreInteracti
 	}
 }
 
-func (it *CoreInteraction) FilterStrategyByInterface(checkInterface interface{}) (strategies []string, found bool) {
+func GetStrategySignatures(exchangeStrategies map[string]SingleExchangeStrategy) []string {
```
c9s, line 46: since it's not used from other package, make it a private function?

pkg/bbgo/interact.go
```diff
@@ -61,12 +43,21 @@ func NewCoreInteraction(environment *Environment, trader *Trader) *CoreInteracti
...
+	var strategies []string
+	for signature := range exchangeStrategies {
+		strategies = append(strategies, signature)
+	}
+
+	return strategies
+}
+
+func FilterStrategyByInterface(checkInterface interface{}, exchangeStrategies map[string]SingleExchangeStrategy) (strategies map[string]SingleExchangeStrategy, found bool) {
```
c9s, line 55: since it's not used from other package, make it a private function?

c9s (approved):

c9s (approved):

PR #546, opened 2022-04-19
feature: add max order history api

c9s:


narumiruna (approved):
Very beautiful, very powerful.

PR #553, opened 2022-04-21
refactor: rewrite max private trade query request with requestgen

c9s:


pkg/exchange/max/maxapi/trade.go
```diff
@@ -111,129 +116,36 @@ func (options *QueryTradeOptions) Params() url.Values {
...
 
 type PrivateRequestParams struct {
 	Nonce int64  `json:"nonce"`
 	Path  string `json:"path"`
 }
 
-type PrivateTradeRequest struct {
-	client *RestClient
+//go:generate GetRequest -url "v2/trades/my" -type GetPrivateTradesRequest -responseType []Trade
+type GetPrivateTradesRequest struct {
```
narumiruna, line 129: no page number here?

narumiruna (approved):

pkg/exchange/max/maxapi/trade.go
c9s: yeah, we don't use pagination for this request

PR #593, opened 2022-05-04
glassnode: simplify NewAuthenticatedRequest

narumiruna:


pkg/datasource/glassnode/glassnodeapi/client.go
```diff
@@ -43,11 +43,6 @@ func (c *RestClient) Auth(apiKey string) {
 }
 
 func (c *RestClient) NewRequest(ctx context.Context, method string, refURL string, params url.Values, payload interface{}) (*http.Request, error) {
```
c9s, line 45: requestgen now provides a BaseApiClient, it implements the NewRequest method, so you can reduce your code.

c9s (approved):

pkg/datasource/glassnode/glassnodeapi/client.go
c9s: check this https://github.com/c9s/requestgen/blob/main/example/api/client.go#L27

narumiruna: done

PR #617, opened 2022-05-16
feature: bollmaker dynamic spread

andycheng123:
Dynamic spread is an experimental feature.
Enables the automatic adjustment to bid and ask spread.

### DIFF
--- a/config/bollmaker.yaml
+++ b/config/bollmaker.yaml
@@ -68,7 +68,7 @@ exchangeStrategies:
     #     # exp means we want to use exponential scale, you can replace "exp" with "linear" for linear scale
     #     exp:
     #       # amplitude from low to high
-    #       domain: [ 0.0005, 0.005 ]
+    #       domain: [ 0.0001, 0.005 ]
     #       # when amplitude is low, minimum ask is spread 0.05%
     #       # when amplitude is high, maximum ask is spread 0.2%
     #       range: [ 0.001, 0.002 ]
@@ -77,7 +77,7 @@ exchangeStrategies:
     #       # exp means we want to use exponential scale, you can replace "exp" with "linear" for linear scale
     #       exp:
     #         # amplitude from low to high
-    #         domain: [ 0.0005, 0.005 ]
+    #         domain: [ 0.0001, 0.005 ]
     #         # when amplitude is low, minimum bid is spread 0.05%
     #         # when amplitude is high, maximum bid is spread 0.2%
     #         range: [ 0.001, 0.002 ]
--- a/config/bollmaker.yaml
+++ b/config/bollmaker.yaml
@@ -58,15 +58,29 @@ exchangeStrategies:
 
     # EXPERIMENTAL
     #   Dynamic spread is an experimental feature. Use at your own risk!
-    #   DynamicSpreadWindow enables the automatic adjustment to bid and ask spread.
+    #
+    #   dynamicSpreadWindow enables the automatic adjustment to bid and ask spread.
     #   When DynamicSpreadWindow is set and larger than 0, the spreads are calculated based on the SMA of amplitude of
-    #   previous [DynamicSpreadWindow] K-lines. Max and min spreads limit the maximum and minimum of spreads. Set them
-    #   to 0 to disable the limitation.
+    #   previous [DynamicSpreadWindow] K-lines.
     # dynamicSpreadWindow: 1
-    # minAskSpread: 0.05%
-    # minBidSpread: 0.05%
-    # maxAskSpread: 1
-    # maxBidSpread: 1
+    # dynamicAskSpreadScale:
+    #   byPercentage:
+    #     # exp means we want to use exponential scale, you can replace "exp" with "linear" for linear scale
+    #     exp:
+    #       # amplitude from low to high
+    #       domain: [ 0.001, 0.01 ]
+    #       # when amplitude is low, minimum ask is spread 0.05%
+    #       # when amplitude is high, maximum ask is spread 0.2%
+    #       range: [ 0.0005, 0.002 ]
+    #   dynamicBidSpreadScale:
+    #     byPercentage:
+    #       # exp means we want to use exponential scale, you can replace "exp" with "linear" for linear scale
+    #       exp:
+    #         # amplitude from low to high
+    #         domain: [ 0.001, 0.01 ]
+    #         # when amplitude is low, minimum bid is spread 0.05%
+    #         # when amplitude is high, maximum bid is spread 0.2%
+    #         range: [ 0.0005, 0.002 ]
 
     # maxExposurePosition is the maximum position you can hold
     # +10 means you can hold 10 ETH long position by maximum
--- a/config/bollmaker.yaml
+++ b/config/bollmaker.yaml
@@ -68,19 +68,19 @@ exchangeStrategies:
     #     # exp means we want to use exponential scale, you can replace "exp" with "linear" for linear scale
     #     exp:
     #       # amplitude from low to high
-    #       domain: [ 0.001, 0.01 ]
+    #       domain: [ 0.0005, 0.005 ]
     #       # when amplitude is low, minimum ask is spread 0.05%
     #       # when amplitude is high, maximum ask is spread 0.2%
-    #       range: [ 0.0005, 0.002 ]
+    #       range: [ 0.001, 0.002 ]
     #   dynamicBidSpreadScale:
     #     byPercentage:
     #       # exp means we want to use exponential scale, you can replace "exp" with "linear" for linear scale
     #       exp:
     #         # amplitude from low to high
-    #         domain: [ 0.001, 0.01 ]
+    #         domain: [ 0.0005, 0.005 ]
     #         # when amplitude is low, minimum bid is spread 0.05%
     #         # when amplitude is high, maximum bid is spread 0.2%
-    #         range: [ 0.0005, 0.002 ]
+    #         range: [ 0.001, 0.002 ]
 
     # maxExposurePosition is the maximum position you can hold
     # +10 means you can hold 10 ETH long position by maximum
--- a/config/bollmaker.yaml
+++ b/config/bollmaker.yaml
@@ -56,6 +56,19 @@ exchangeStrategies:
     # For short position, you will only place buy order below the price (= average cost * (1 - minProfitSpread))
     minProfitSpread: 0.1%
 
+    # EXPERIMENTAL
+    # Dynamic spread is an experimental feature. Use at your own risk!
+    # DynamicSpreadWindow enables the automatic adjustment to bid and ask spread.
+    # When DynamicSpreadWindow is set and is larger than 0, the spreads are calculated based on the SMA of amplitude of
+    # previous [DynamicSpreadWindow] K-lines. Max and min spreads limit the maximum and minimum of spreads. Set them to
+    # 0 to disable the limitation
+    #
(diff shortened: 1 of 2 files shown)


pkg/strategy/bollmaker/strategy.go
```diff
@@ -653,6 +671,36 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
 			return
 		}
 
+		// Update spreads
```
zenixls2, line 674: please filter kline.Interval.  
In backtest, 1m is automatically subscribed (however, in runtime, only s.Interval is used)

zenixls2: cc @c9s known differences between runtime and backtest.

pkg/strategy/bollmaker/strategy.go
```diff
@@ -653,6 +671,36 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
...
+		if s.DynamicSpreadWindow > 0 && kline.Direction() == types.DirectionUp {
+			s.DynamicAskSpread.Update(kline.GetHigh().Sub(kline.GetOpen()).Div(kline.GetOpen()).Float64())
```
zenixls2, line 676: maybe create variables in float64 first, and then do the calculation?  
ex: 
```go
high := kline.High.Float64()
open := kline.Open.Float64()
```

zenixls2: so that in line 682 we don't need to convert the float64 back to fixedpoint and do slow comparison.

pkg/strategy/bollmaker/strategy.go
```diff
@@ -653,6 +671,36 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
...
+		}
+		if s.DynamicSpreadWindow > 0 && kline.Direction() == types.DirectionDown {
+			s.DynamicBidSpread.Update(kline.GetOpen().Sub(kline.GetLow()).Div(kline.GetOpen()).Float64())
+		}
+		if s.DynamicSpreadWindow > 0 && s.DynamicBidSpread.Length() >= s.DynamicSpreadWindow {
+			dynamicBidSpread := fixedpoint.NewFromFloat(s.DynamicBidSpread.Last())
+			if !s.MaxBidSpread.IsZero() && dynamicBidSpread.Compare(s.MaxBidSpread) > 0 {
+				s.BidSpread = s.MaxBidSpread
```
zenixls2, line 684: what's the default value of BidSpread and AskSpread when window dynamic bid spread's length is less than the window size?

config/bollmaker.yaml
```diff
@@ -56,6 +56,19 @@ exchangeStrategies:
     # For short position, you will only place buy order below the price (= average cost * (1 - minProfitSpread))
     minProfitSpread: 0.1%
 
+    # EXPERIMENTAL
```
c9s, line 59: can you add a block comment for EXPERIMENTAL? or we might think all the following ocnfig are experimental.

pkg/strategy/bollmaker/strategy.go
andycheng123: This the intended behavior. I use the same interval as the strategy

pkg/strategy/bollmaker/strategy.go
andycheng123: askSpread and bidSpread
Or spread

pkg/strategy/bollmaker/strategy.go
```diff
@@ -653,6 +671,36 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
 			return
 		}
 
+		// Update spreads
+		if s.DynamicSpreadWindow > 0 && kline.Direction() == types.DirectionUp {
+			s.DynamicAskSpread.Update(kline.GetHigh().Sub(kline.GetOpen()).Div(kline.GetOpen()).Float64())
+		}
+		if s.DynamicSpreadWindow > 0 && kline.Direction() == types.DirectionDown {
+			s.DynamicBidSpread.Update(kline.GetOpen().Sub(kline.GetLow()).Div(kline.GetOpen()).Float64())
```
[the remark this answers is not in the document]
andycheng123: But that might make it less likely to make the trade?

pkg/strategy/bollmaker/strategy.go
```diff
@@ -653,6 +671,36 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
...
+			log.Infof("new bid spread: %v", s.BidSpread.Percentage())
+		}
+		if s.DynamicSpreadWindow > 0 && s.DynamicAskSpread.Length() >= s.DynamicSpreadWindow {
+			dynamicAskSpread := fixedpoint.NewFromFloat(s.DynamicAskSpread.Last())
+			if !s.MaxAskSpread.IsZero() && dynamicAskSpread.Compare(s.MaxAskSpread) > 0 {
+				s.AskSpread = s.MaxAskSpread
+			} else if !s.MinAskSpread.IsZero() && dynamicAskSpread.Compare(s.MinAskSpread) < 0 {
+				s.AskSpread = s.MinAskSpread
+			} else {
+				s.AskSpread = dynamicAskSpread
```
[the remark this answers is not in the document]
andycheng123: The max spread is to avoid this. The dynamic spread cannot be higher than it if it's not 0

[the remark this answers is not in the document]
andycheng123: Do you meaning scaling between max and min spread?

pkg/strategy/bollmaker/strategy.go
zenixls2: I mean, in backtest, 1m and s.Interval are both subscribed. 1m is subscribed by the system automatically, and s.Interval is subscribed by the strategy.  

while in runtime, only s.Interval is subscribed.

zenixls2: so better to move this after line 709

andycheng123: I see what you mean. That might explain something

pkg/strategy/bollmaker/strategy.go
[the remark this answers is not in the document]
c9s: @andycheng123 yes~ so I can map amplification to a spread range:

AMPL 0.1% -> Spread 0.075%
(in the middle, use exponential scale)
AMPL 1% -> Spread 0.1%

something like this

pkg/strategy/bollmaker/strategy.go
```diff
@@ -276,6 +290,53 @@ func (s *Strategy) getCurrentAllowedExposurePosition(bandPercentage float64) (fi
...
+func (s *Strategy) updateDynamicSpread(kline types.KLine) {
+	if s.DynamicSpreadWindow > 0 {
+		high := kline.GetHigh().Float64()
+		open := kline.GetOpen().Float64()
+		low := kline.GetLow().Float64()
+
+		if kline.Direction() == types.DirectionUp {
+			s.DynamicAskSpread.Update((high - open) / open)
+		}
+		if kline.Direction() == types.DirectionDown {
```
c9s, line 303: use switch case to make it cleaner?

pkg/strategy/bollmaker/strategy.go
```diff
@@ -276,6 +290,53 @@ func (s *Strategy) getCurrentAllowedExposurePosition(bandPercentage float64) (fi
 	return s.MaxExposurePosition, nil
 }
 
+// Update dynamic spreads
+func (s *Strategy) updateDynamicSpread(kline types.KLine) {
+	if s.DynamicSpreadWindow > 0 {
```
c9s, line 295: return directly if it's 0?

pkg/strategy/bollmaker/strategy.go
```diff
@@ -657,6 +724,19 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
 			return
 		}
 
+		// Update spreads
```
c9s, line 727: if dynamic spread window > 0 then do the follow override?

c9s: or maybe:
```
if s.IsDynamicSpreadEnabled() {

 // override
}
```

config/bollmaker.yaml
```diff
@@ -56,6 +56,32 @@ exchangeStrategies:
...
+    #       range: [ 0.001, 0.002 ]
+    #   dynamicBidSpreadScale:
+    #     byPercentage:
+    #       # exp means we want to use exponential scale, you can replace "exp" with "linear" for linear scale
+    #       exp:
+    #         # amplitude from low to high
+    #         domain: [ 0.0001, 0.005 ]
+    #         # when amplitude is low, minimum bid is spread 0.05%
+    #         # when amplitude is high, maximum bid is spread 0.2%
+    #         range: [ 0.001, 0.002 ]
```
c9s, line 83: can you restructure this config:

```
dynamicSpread:
   enabled: true
   window: 1
   askSpreadScale:
     ....
   bidSpreadScale:
      .....
```

pkg/strategy/bollmaker/strategy.go
```diff
@@ -78,6 +78,20 @@ type Strategy struct {
 	// AskSpread overrides the spread setting, this spread will be used for the sell order
 	AskSpread fixedpoint.Value `json:"askSpread,omitempty"`
 
+	// DynamicSpreadWindow enables the automatic adjustment to bid and ask spread.
+	// When DynamicSpreadWindow is set and is larger than 0, the spreads are calculated based on the SMA of amplitude of
+	// [DynamicSpreadWindow] K-lines
+	DynamicSpreadWindow int `json:"dynamicSpreadWindow,omitempty"`
```
c9s, line 84: then, to unmarshal the dynamic spread settings, you can define a new struct called `DynamicSpreadSettings`

pkg/strategy/bollmaker/strategy.go
```diff
@@ -276,6 +290,75 @@ func (s *Strategy) getCurrentAllowedExposurePosition(bandPercentage float64) (fi
...
+func (s *Strategy) IsDynamicSpreadEnabled() bool {
+	if s.DynamicSpreadWindow > 0 && s.DynamicAskSpreadScale != nil && s.DynamicBidSpreadScale != nil {
+		return true
+	}
+
+	return false
+}
+
+// Update dynamic spreads
+func (s *Strategy) updateDynamicSpread(kline types.KLine) {
```
c9s, line 303: this method then could be moved to the DynamicSpreadSettings struct.

pkg/strategy/bollmaker/strategy.go
```diff
@@ -276,6 +290,75 @@ func (s *Strategy) getCurrentAllowedExposurePosition(bandPercentage float64) (fi
...
+		s.DynamicBidSpread.Update(ampl)
+		s.DynamicAskSpread.Update(0)
+	default:
+		s.DynamicAskSpread.Update(0)
+		s.DynamicBidSpread.Update(0)
+	}
+}
+
+// Get current dynamic ask spread
+func (s *Strategy) getDynamicAskSpread() (dynamicAskSpread float64, err error) {
```
c9s, line 324: ditto, to the DynamicSpreadSettings

c9s (approved):

PR #629, opened 2022-05-20
quick fix to filter duplicated spot of cost and base line series. but…

snakewa:
… still need to find out why positionAverageCostHistoryToLineData() and positionBaseHistoryToLineData() will produce thses dots

### DIFF
--- a/apps/backtest-report/components/TradingViewChart.js
+++ b/apps/backtest-report/components/TradingViewChart.js
@@ -398,14 +398,18 @@ const TradingViewChart = (props) => {
       if (chartData.positionHistory) {
         const lineSeries = chart.current.addLineSeries();
         const costLine = positionAverageCostHistoryToLineData(currentInterval, chartData.positionHistory);
-        lineSeries.setData(costLine);
+        const costLineIdx = {}
+        const costLineDeduplicated = costLine.filter( v => { const exists = costLineIdx[v.time]; costLineIdx[v.time] = v.value; return !exists  });
+        lineSeries.setData(costLineDeduplicated);
 
         const baseLineSeries = chart.current.addLineSeries({
           priceScaleId: 'left',
           color: '#98338C',
         });
         const baseLine = positionBaseHistoryToLineData(currentInterval, chartData.positionHistory)
-        baseLineSeries.setData(baseLine);
+        const baseLineIdx = {}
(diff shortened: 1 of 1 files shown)


snakewa:
polygon:0x11AE25B81A681479A0BeA0e32a22505214F9F4c2

c9s:
these data are triggered by the klines, some klines are duplicated, we need to fix it from the golang side.

c9s:
these position history data are huge, would suggest not to iterate all of them. (we have already iterate it once 😂

apps/backtest-report/components/TradingViewChart.js
```diff
@@ -398,14 +398,18 @@ const TradingViewChart = (props) => {
       if (chartData.positionHistory) {
         const lineSeries = chart.current.addLineSeries();
         const costLine = positionAverageCostHistoryToLineData(currentInterval, chartData.positionHistory);
-        lineSeries.setData(costLine);
+        const costLineIdx = {}
+        const costLineDeduplicated = costLine.filter( v => { const exists = costLineIdx[v.time]; costLineIdx[v.time] = v.value; return !exists  });
```
c9s, line 402: move this check to the positionAverageCostHistoryToLineData function? there is already a for loop.

apps/backtest-report/components/TradingViewChart.js
```diff
@@ -398,14 +398,18 @@ const TradingViewChart = (props) => {
...
+        lineSeries.setData(costLineDeduplicated);
 
         const baseLineSeries = chart.current.addLineSeries({
           priceScaleId: 'left',
           color: '#98338C',
         });
         const baseLine = positionBaseHistoryToLineData(currentInterval, chartData.positionHistory)
-        baseLineSeries.setData(baseLine);
+        const baseLineIdx = {}
+        const baseLineDeduplicated = baseLine.filter( v => { const exists = baseLineIdx[v.time]; baseLineIdx[v.time] = v.value; return !exists  });
```
c9s, line 411: move this check to the positionBaseHistoryToLineData function? there is already a for loop.

apps/backtest-report/components/TradingViewChart.js
snakewa: sure. let me look into the function

zenixls2:
any update or action required for this pr? @c9s @snakewa 
if not, maybe we could close the pr.

PR #634, opened 2022-05-26
rebalance: place limit orders

narumiruna:


pkg/strategy/rebalance/strategy.go
```diff
@@ -79,13 +81,24 @@ func (s *Strategy) Subscribe(session *bbgo.ExchangeSession) {
 }
 
 func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, session *bbgo.ExchangeSession) error {
+	s.orderStore = bbgo.NewOrderStore("")
+	s.orderStore.RemoveCancelled = true
+	s.orderStore.RemoveFilled = true
```
c9s, line 86: order store is used for collect the trades, sometimes the order get filled before you receive trade, so RemoveFilled needs to be false.

narumiruna: Is ok to use OrderStore here?

I just want to 
1. store created orders after submitting orders
2. cancel all active orders before the next rebalancing

c9s: use maintain the active orders, use LocalActiveOrderBook ~

c9s (approved):

PR #655, opened 2022-06-02
fix: improve and fix kline sync

c9s:


pkg/service/backtest.go
```diff
@@ -23,16 +24,43 @@ type BacktestService struct {
...
+	// override symbol if isolatedSymbol is not empty
+	if isIsolated && len(isolatedSymbol) > 0 {
+		symbol = isolatedSymbol
+	}
+
+	tasks := []SyncTask{
+		{
+			Type: types.KLine{},
+			Time: func(obj interface{}) time.Time {
+				return obj.(types.KLine).StartTime.Time()
```
zenixls2, line 38: Can we set this to UTC()?
Refer: my latest pr

c9s: Sure

PR #662, opened 2022-06-02
WIP: strategy: add mirror

austin362667:
panic: reflect: call of reflect.Value.NumField on zero Value


austin362667:
#663 resolved

PR #670, opened 2022-06-05
Improve backtest report ui

ksidenko:
Updated visualization

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@
 
 bbgo.sqlite3
 node_modules
+output
 
 otp*png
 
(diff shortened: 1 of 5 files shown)


ksidenko:
@c9s hi!
In orders data there are order dublicates. I don't why. Can you check it?

ksidenko:
polygon:0xB947e83C893F115037557d559dCA76aF6d612271

ksidenko:
@c9s another thing, that is important, but I don't know how I can do it, is pass custom strategy params to backtest report. Have you any advise?

c9s:
> @c9s hi! In orders data there are order dublicates. I don't understand why. Can you check it?

Are these order ids different? or they have different status?

I will also check this later.

c9s:
> 

You can check pkg/cmd/backtest.go LINE 505, I think we can render the config file into a JSON file.

c9s:
> another thing, that is important, but I don't know how I can do it, is pass custom strategy params to backtest report. Have you any advise?

You can check pkg/cmd/backtest.go LINE 505, I think we can render the config file into a JSON file.

c9s (approved):

c9s:
neat!

PR #678, opened 2022-06-06
interact: fix missing make()

andycheng123:
Fix missing make()

### DIFF
--- a/pkg/bbgo/interact.go
+++ b/pkg/bbgo/interact.go
@@ -413,7 +413,7 @@ func getStrategySignature(strategy SingleExchangeStrategy) (string, error) {
 
 	var signature = path.Base(rv.Type().PkgPath())
 
-	id := callID(strategy)
+	var id = strategy.ID()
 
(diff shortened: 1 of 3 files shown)


c9s (approved):

pkg/bbgo/interact.go
```diff
@@ -410,12 +411,12 @@ func getStrategySignature(strategy SingleExchangeStrategy) (string, error) {
...
-	var signature = path.Base(rv.Type().PkgPath())
-
-	var id = strategy.ID()
-
-	if !strings.EqualFold(id, signature) {
-		signature += "." + strings.ToLower(id)
+	var signature string
+	var id = callID(strategy)
+	if id != "" {
+		signature = id
```
c9s, line 417: you can return directly? since the instance ID should already be unique identifier

andycheng123: So we don't need interval in the signature?

c9s: if we have instance id, we don't need to build the signature.

PR #703, opened 2022-06-10
fix: syncing goroutine leak

ycdesu:
The syncing operation is protected by the mutex, so only one goroutine can sync the history. We shouldn't start new goroutines when our previous syncing operation is not done yet. It causes a memory leak.


c9s (approved):
Thanks!!!

ycdesu:
polygon:0x131D4f07574Dabd3125ffCaa6A57A50Ec8034B11

PR #707, opened 2022-06-10
feature: add basic implementation of alma indicator

zenixls2:
refer: https://capital.com/arnaud-legoux-moving-average
also: https://github.com/DaveSkender/Stock.Indicators/blob/main/src/a-d/Alma/Alma.cs


pkg/indicator/alma.go
```diff
@@ -0,0 +1,95 @@
...
+	Sum                  float64
+	input                []float64
+	Values               types.Float64Slice
+	UpdateCallbacks      []func(value float64)
+}
+
+const MaxNumOfALMA = 5_000
+const MaxNumOfALMATruncateSize = 100
+
+func (inc *ALMA) Update(value float64) {
```
c9s, line 29: call this `update` and rename `calculateAndUpdate` to `Update` as an exported method?

c9s (approved):

PR #714, opened 2022-06-14
improve: support specifying session in the sync symbol

c9s:


pkg/bbgo/config.go
```diff
@@ -208,12 +208,53 @@ func GetNativeBuildTargetConfig() BuildTargetConfig {
...
+		}
+	}
+	return m, rest
+}
+
 type SyncConfig struct {
 	// Sessions to sync, if ignored, all defined sessions will sync
 	Sessions []string `json:"sessions,omitempty" yaml:"sessions,omitempty"`
 
 	// Symbols is the list of symbol to sync, if ignored, symbols wlll be discovered by your existing crypto balances
```
zenixls2, line 256: the list of {symbol, session} pairs to sync

c9s: 👍

pkg/cmd/sync.go
```diff
@@ -89,7 +89,9 @@ var SyncCmd = &cobra.Command{
 
 		if len(symbol) > 0 {
 			if userConfig.Sync != nil && len(userConfig.Sync.Symbols) > 0 {
-				userConfig.Sync.Symbols = []string{symbol}
+				userConfig.Sync.Symbols = []bbgo.SyncSymbol{
+					{Symbol: symbol},
```
zenixls2, line 93: what about sessionNames?

c9s: session name is optional, when session name is empty, the symbol will be applied to all sessions.

PR #716, opened 2022-06-14
Do not block whole page while syncing

ycdesu:
Currently we block the whole page when the server is syncing. But actually we don't have to invoke sync API for some tabs, like strategy. Therefore, I remove the blocking dialog. We should be able to open the strategy page even if the syncing operation is not done yet.

The first time we enter the bbgo frontend page, it invokes the asynchronous sync API first. When we click the order or trade tab, it will query the orders/trades from bbgo API. It's reasonable not to receive whole data since the orders and trades are inserted anytime. it's not possible to always have latest data if bbgo only provides REST API.

So the logic is the same as previous implementation. 

Also, you can trigger the syncing anytime by click the sync button #715

### DIFF
--- a/frontend/api/bbgo.ts
+++ b/frontend/api/bbgo.ts
@@ -15,18 +15,19 @@ export function queryOutboundIP(cb) {
   });
 }
 
-const triggerSync = async () => {
+export async function triggerSync() {
   return axios.post<any>(baseURL + '/api/environment/sync');
 };
 
-export { triggerSync };
+export enum SyncStatus {
+  SyncNotStarted = 0,
+  Syncing = 1,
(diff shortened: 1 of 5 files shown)


c9s:
@ycdesu can we show the sync status somewhere?

c9s (approved):

ycdesu:
> @ycdesu can we show the sync status somewhere?

@c9s if it's syncing, I can disable the sync button and modify the button text to syncing. How about that?

ycdesu:
@c9s could you review it again? I add the sync status to the button, and will fetch the status every 2 seconds.

c9s (approved):
thanks!!

PR #719, opened 2022-06-15
optimizer: bool type parameter

andycheng123:


pkg/optimizer/grid.go
```diff
@@ -115,6 +115,35 @@ func (o *GridOptimizer) buildOps() []OpFunc {
 				return nil
 			}
 			ops = append(ops, f)
+		case "bool":
+			values := []bool{true, false}
+			f := func(configJson []byte, next func(configJson []byte) error) error {
+				for _, val := range values {
+					log.Debugf("%d %s: %v of %v", ii, path, val, values)
+
+					jsonOp := []byte(reformatJson(fmt.Sprintf(`[{"op": "replace", "path": "%s", "value": %v}]`, path, val)))
```
c9s, line 124: %v is the only difference from the iterate case?

andycheng123: line 119 is also the difference
and users don't have to set values in config

c9s (approved):

PR #720, opened 2022-06-15
Fix/supertrend

andycheng123:


pkg/strategy/supertrend/strategy.go
```diff
@@ -295,7 +295,8 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
 		// TP/SL if there's non-dust position
 		if !s.Market.IsDustQuantity(base.Abs(), kline.GetClose()) {
 			if s.StopLossByTriggeringK && !s.currentStopLossPrice.IsZero() && ((baseSign < 0 && kline.GetClose().Compare(s.currentStopLossPrice) > 0) || (baseSign > 0 && kline.GetClose().Compare(s.currentStopLossPrice) < 0)) {
-				// SL by triggered Kline low
+				// SL by triggering Kline low
+				log.Infof("SL by triggering Kline low")
```
c9s, line 299: would suggest to show the symbol name in the log

pkg/strategy/supertrend/strategy.go
```diff
@@ -344,10 +347,18 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
 		}
 
 		if side == types.SideTypeSell || side == types.SideTypeBuy {
```
c9s, line 349: You don't need this condition?

c9s: since It's always true?

pkg/strategy/supertrend/strategy.go
```diff
@@ -344,10 +347,18 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
...
+			log.Infof("open position for signal %v", side)
 			// Close opposite position if any
-			if !s.Market.IsDustQuantity(base.Abs(), kline.GetClose()) && ((side == types.SideTypeSell && baseSign > 0) || (side == types.SideTypeBuy && baseSign < 0)) {
-				if err := s.ClosePosition(ctx, fixedpoint.One); err != nil {
-					s.Notify("can not place close position order")
+			if !s.Market.IsDustQuantity(base.Abs(), kline.GetClose()) {
+				if (side == types.SideTypeSell && baseSign > 0) || (side == types.SideTypeBuy && baseSign < 0) {
```
c9s, line 353: you can use the position method to check,  IsShort, IsLong

PR #727, opened 2022-06-15
strategy: add marketcap strategy

narumiruna:


pkg/fixedpoint/value_map.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (m ValueMap) Sub(n ValueMap) ValueMap {
+	if len(m) != len(n) {
+		panic("unequal length")
+	}
+
+	o := ValueMap{}
+
+	for k, v := range m {
+		o[k] = v.Sub(n[k])
```
zenixls2, line 46:
```go
nv, ok := n[k]
if !ok {
    panic("key not exists")
}
```

pkg/fixedpoint/value_map.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (m ValueMap) Mul(n ValueMap) ValueMap {
+	if len(m) != len(n) {
+		panic("unequal length")
+	}
+
+	o := ValueMap{}
+
+	for k, v := range m {
+		o[k] = v.Mul(n[k])
```
zenixls2, line 60: same as line 46

pkg/fixedpoint/value_map.go
```diff
@@ -0,0 +1,135 @@
...
+	o := ValueMap{}
+
+	for k, v := range m {
+		o[k] = v.Div(n[k])
+	}
+
+	return o
+}
+
+func (m ValueMap) AddScalar(x Value) ValueMap {
```
zenixls2, line 80: I recommend using `Add(interface{})` similar to `types.Add` in `types/indicator.go`.
Users could pass in either a `ValueMap` or a `fixedpoint.Value` or even a `float64` to do the job.

pkg/strategy/marketcap/types.go
```diff
@@ -0,0 +1,42 @@
+package marketcap
```
zenixls2, line 1: move this to types?

pkg/fixedpoint/value_map.go
narumiruna: Ok, maybe we can create another PR to do this.

c9s (approved):

PR #757, opened 2022-06-22
totp-user: add default user 'bbgo'

iamken1204:
There's no `USER` env var in alpine image, causes the program throws error 'No USER or USERNAME' in containers.

* Create and assign `USER` env var in bbgo image
* Fallback to use the default user 'bbgo' when env `USER` or `USERNAME` was unassigned


c9s (approved):
thanks!

PR #817, opened 2022-07-13
optimizer: prepare database before executing backtests

COLDTURNIP:
In current design, optimizer launches `bbgo backtest` as subprocesses, and these processes will apply DB migration during `DatabaseService` initialization. However, if the optimizer is launched with an empty database, the migrations in these subprocesses would apply their own migration and conflict. There are some ways to prevent such situation:

1. Check the DB before launching backtests, and tip user to execute `bbgo backtest --sync --sync-only` before grid searching logic.
2. Trigger DB sync before backtests. Implemented by this PR.

In addition, to the second approach, maybe it would be more convenient to provide `--sync-from` option to user?

### DIFF
--- a/pkg/cmd/optimize.go
+++ b/pkg/cmd/optimize.go
@@ -101,6 +101,10 @@ var optimizeCmd = &cobra.Command{
 			Config: optConfig,
 		}
 
+		if err := executor.Prepare(configJson); err != nil {
+			return err
+		}
+
 		metrics, err := optz.Run(executor, configJson)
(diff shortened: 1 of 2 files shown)


c9s (approved):
neat!

COLDTURNIP:
polygon:0x1787877702357690b10FE25C092A3E5Aa2322567

COLDTURNIP:
polygon:0x1787877702357690b10FE25C092A3E5Aa2322567

PR #818, opened 2022-07-13
backtest: correct final asset calculation

COLDTURNIP:
The final asset should be

> quote + balance * _last_price_

Instead of

> quote + balance * _start_price_


c9s (approved):
oh yes, thanks!

PR #813, opened 2022-07-12
feature: drift study

zenixls2:
add drift strategy for study purpose.
- plot series
- dynamic kline (result is bad, remove related implementation from strategy)
- fix bug for SeriesExtend.Array/Reverse
image

PR #819, opened 2022-07-13
fix limit not working

kfrico:


c9s:
TestLogisticRegression test failed, could you please take a look? thanks

kfrico:
this should be due to the use of the "Reverse" method
The number of A slice and B sliece in the test case is not equal ?
@zenixls2

pkg/types/indicator.go
```diff
@@ -133,13 +133,14 @@ type BoolSeries interface {
 // if limit is given, will only sum first limit numbers (a.Index[0..limit])
 // otherwise will sum all elements
 func Sum(a Series, limit ...int) (sum float64) {
-	l := -1
+	l := a.Length()
+
 	if len(limit) > 0 {
```
zenixls2, line 138:
```golang
if len(limit) > 0 && limit[0] < l {
    l = limit[0]
}
```

zenixls2:
have fixed the issue in another pr.
Just wait to be merged.
but once you fix the syntax that I suggested in this pr, I think we could merge this first.

zenixls2:
refer: https://github.com/c9s/bbgo/pull/813/commits/66a61e6f5f91ef36cd513b5e555df1817448c1e3
refer: https://github.com/c9s/bbgo/pull/813/commits/f8608125c9910ed806aef4c35aae252f6f90a715

zenixls2:
no side effect on this pr, so we don't need all tests to be passed.

PR #828, opened 2022-07-15
optimizer: prevent deadlock for huge number of grid points

COLDTURNIP:
Currently we queue the backtest tasks in a buffered channel, and enqueue all the tasks before execution. This will cause deadlock if the number of grid points more than 10,000. In this PR, it will tip user to reduce the number of tests.

TODO: generate the tasks on another goroutine


COLDTURNIP:
This PR is replaced by #829 .

PR #829, opened 2022-07-15
optimizer: eliminate limitation of number of grid point

COLDTURNIP:
Currently we queue the backtest tasks in a buffered channel, and enqueue all the tasks before execution. This will cause deadlock if the number of grid points more than 10,000. In this PR, it generate tasks asynchronously on another goroutine, hence the deadlock won't happen.

This PR replaces #828


c9s (approved):
Awesome thanks!

PR #830, opened 2022-07-17
backtest: resolve data race on index.json

COLDTURNIP:
Before this change, each time when backtest tries to read/write/append output/index.json, it will create an empty .report.lock lock file, acquire the file lock, I/O on index.json, unlock, then delete the lock file.

However, this does not prevent the data race between backtests:

1. backtest A acquire the lock, and finish the works on index.json
2. backtest B wait for the lock
3. backtest A release the lock
4. backtest B acquire the lock
5. backtest A delete .report.lock
6. backtest C want to modify index.json, hence it creates a new .report.lock at the same path, lock on the new lock file
7. index.json is modified by B and C simultaneously since there are multiple lock instances

In POSIX, file deletion is not protected by file locks. While performing optimizer, this problem is very likely to occur when we increase the number of grid points.

In this PR, we apply the file lock on index.json itself.

### DIFF
--- a/pkg/backtest/report.go
+++ b/pkg/backtest/report.go
@@ -149,18 +149,89 @@ func FormatSessionName(sessions []string, symbols []string, startTime, endTime t
 }
 
 func WriteReportIndex(outputDirectory string, reportIndex *ReportIndex) error {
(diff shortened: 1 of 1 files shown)


c9s (approved):
Awesome!

PR #869, opened 2022-08-09
Revert "feature: add smart cancel to drift"

c9s:
Reverts c9s/bbgo#853


c9s:
This PR requires go 1.18, which is not backward compatible.

PR #870, opened 2022-08-10
drift improvement

zenixls2:
feature: smartcancel for drift
feature: rebalance on internal position
feature: dump config
feature: dump parameter to tg, esp series
fix: order tag, position calculation and bp/sp of drift
fix: empty internal position on start if there's already stored data in redis


c9s:
conflicts?

austin362667 (approved):

PR #881, opened 2022-08-17
feature: drift & print strategy config

zenixls2:
also fix `exits` printing
re-implement `CanInt` which doesn't exist in go1.17
call TruncateQuantity on the SubmitOrders and new Trades in backtest

### DIFF
--- a/pkg/backtest/matching.go
+++ b/pkg/backtest/matching.go
@@ -148,11 +148,13 @@ func (m *SimplePriceMatching) PlaceOrder(o types.SubmitOrder) (*types.Order, *ty
 		price = o.Price
 	}
 
+	o.Quantity = m.Market.TruncateQuantity(o.Quantity)
+
 	if o.Quantity.Compare(m.Market.MinQuantity) < 0 {
 		return nil, nil, fmt.Errorf("order quantity %s is less than minQuantity %s, order: %+v", o.Quantity.String(), m.Market.MinQuantity.String(), o)
 	}
 
-	quoteQuantity := o.Quantity.Mul(price)
+	quoteQuantity := m.Market.TruncateQuantity(o.Quantity.Mul(price))
 	if quoteQuantity.Compare(m.Market.MinNotional) < 0 {
 		return nil, nil, fmt.Errorf("order amount %s is less than minNotional %s, order: %+v", quoteQuantity.String(), m.Market.MinNotional.String(), o)
 	}
@@ -297,7 +299,7 @@ func (m *SimplePriceMatching) newTradeFromOrder(order *types.Order, isMaker bool
 	// BINANCE uses 0.1% for both maker and taker
 	// MAX uses 0.050% for maker and 0.15% for taker
 	var feeRate = m.getFeeRate(isMaker)
-	var quoteQuantity = order.Quantity.Mul(price)
+	var quoteQuantity = m.Market.TruncateQuantity(order.Quantity.Mul(price))
 	var fee fixedpoint.Value
 	var feeCurrency string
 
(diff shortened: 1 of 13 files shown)


pkg/bbgo/exit.go
```diff
@@ -35,6 +37,35 @@ type ExitMethod struct {
 	TrailingStop              *TrailingStop2             `json:"trailingStop"`
 }
 
+func (e ExitMethod) String() string {
```
c9s, line 40: since it's String(), maybe use YAML format for more readability ?

pkg/strategy/drift/output.go
```diff
@@ -26,6 +22,16 @@ func (a jsonArr) Len() int           { return len(a) }
 func (a jsonArr) Less(i, j int) bool { return a[i].key < a[j].key }
 func (a jsonArr) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
 
+func canInt(v reflect.Value) bool {
```
c9s, line 25: Rename to `isInt` and move it to the pkg/dynamic package?

pkg/strategy/output.go
```diff
@@ -0,0 +1,151 @@
...
+func DefaultWhiteList() []string {
+	return []string{"Window", "Interval", "Symbol"}
+}
+
+// @param s: strategy object
+// @param f: io.Writer used for writing the config dump
+// @param style: pretty print table style. Use DefaultStyle() to get default one.
+// @param withColor: whether to print with color
+// @param whiteLists: fields to be printed out from embedded struct (1st layer only)
+func PrintConfig(s interface{}, f io.Writer, style *table.Style, withColor bool, whiteLists ...string) {
```
c9s, line 54: move to pkg/dynamic?

pkg/strategy/drift/output.go
zenixls2: canInt maps to https://pkg.go.dev/reflect#Value.CanInt

pkg/bbgo/exit.go
zenixls2: using YAML format will occupy more spaces/lines than json.  Notice that the final output will be inside a column of a table.
also inside ExitMethod, we only define json tags, so YAML format won't work naively.

pkg/strategy/output.go
zenixls2: there'll be a import cycle if I put this in `pkg/dynamic`.

pkg/backtest/matching.go
```diff
@@ -147,6 +147,11 @@ func (m *SimplePriceMatching) PlaceOrder(o types.SubmitOrder) (*types.Order, *ty
 
 	case types.OrderTypeLimit, types.OrderTypeStopLimit, types.OrderTypeLimitMaker:
 		o.Price = m.Market.TruncatePrice(o.Price)
+		// if o.Price makes the limit order a taker, use the latest price as the order price
+		// otherwise we'll endup locking more assets than needed
+		if isLimitTakerOrder(o, m.LastPrice) {
+			o.Price = m.LastPrice
```
c9s, line 153: maybe you can assign to AveragePrice? the price of the limit order can not be changed.

pkg/backtest/matching_test.go
```diff
@@ -518,7 +518,8 @@ func TestSimplePriceMatching_LimitTakerOrder(t *testing.T) {
 	closedOrder, trade, err = engine.PlaceOrder(newLimitOrder("BTCUSDT", types.SideTypeSell, 19000.0, 1.0))
 	assert.NoError(t, err)
 	if assert.NotNil(t, closedOrder) {
-		assert.Equal(t, "19000", closedOrder.Price.String())
```
c9s, line 517: we need to ensure that the price of the limit order is not changed.

pkg/strategy/drift/output.go
c9s: got it

c9s (approved):

PR #890, opened 2022-08-23
weighted drift

zenixls2:
- add weighted drift
- add gma (geometric mean)
- export drfit filters to config
- truncate LastPrice when it's used as the order price
- clean unused graphs


pkg/indicator/gma.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"math"
+
+	"github.com/c9s/bbgo/pkg/types"
+)
+
+// Geometric Moving Average
+//go:generate callbackgen -type GMA
+type GMA struct {
```
c9s, line 11: you can add GMA to the standard indicator struct~

c9s (approved):

pkg/indicator/gma.go
zenixls2: I'll keep it here, since there's no standard implementation/definition of GMA in any other platform/lib.

PR #910, opened 2022-09-01
SerialMarketDataStore, elliottwave renewal

zenixls2:
the api design of SerialMarketDataStore still require further discussion


pkg/bbgo/order_executor_general.go
```diff
@@ -144,6 +144,20 @@ func (e *GeneralOrderExecutor) GracefulCancelActiveOrderBook(ctx context.Context
 	return nil
 }
 
+func (e *GeneralOrderExecutor) Cancel(ctx context.Context, order types.Order) error {
```
c9s, line 147: call it `CancelOrder`?

pkg/strategy/source.go
```diff
@@ -0,0 +1,50 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+type SourceFunc func(*types.KLine) fixedpoint.Value
+
+var Four fixedpoint.Value = fixedpoint.NewFromInt(4)
+var Three fixedpoint.Value = fixedpoint.NewFromInt(3)
+var Two fixedpoint.Value = fixedpoint.NewFromInt(2)
+
+type SourceSelector struct {
```
c9s, line 17: what's this for?

pkg/strategy/elliottwave/strategy.go
```diff
@@ -0,0 +1,495 @@
...
+}
+
+type SourceFunc func(*types.KLine) fixedpoint.Value
+
+type Strategy struct {
+	Symbol string `json:"symbol"`
+
+	bbgo.StrategyController
+	types.Market
+	strategy.SourceSelector
```
c9s, line 41: this does not need to be embedded in the struct? we can use a field for this?

zenixls2: it could be, just that when we call `getSource`
there's no syntax sugar like `s.getSource`. Need to write `s.Selector.getSource`.

pkg/strategy/source.go
zenixls2: switching between different price sources from config.
for example, use `close` in config provides functionality to fetch `kline.Close` automatically and feed into indicators.

pkg/bbgo/order_executor_general.go
zenixls2: call it `GracefulCancelOrder` instead.  
we already have `CancelOrders` which directly cancels the orders from the Exchange.

pkg/bbgo/session.go
```diff
@@ -499,6 +499,28 @@ func (session *ExchangeSession) MarketDataStore(symbol string) (s *MarketDataSto
...
+	if !ok {
+		log.Errorf("SerialMarketDataStore: cannot get 1m history")
+		return nil, false
+	}
+	for _, interval := range intervals {
+		store.Subscribe(interval)
+	}
+	for _, kline := range *klines {
+		store.AddKLine(kline)
+	}
```
c9s, line 519: pull 508 to 519 to a function, so that we can test it?

pkg/strategy/source.go
c9s: How about moving this to the types package?

pkg/bbgo/session.go
```diff
@@ -418,29 +420,35 @@ func (session *ExchangeSession) initSymbol(ctx context.Context, environ *Environ
...
 		endTime := environ.startTime
-		kLines, err := session.Exchange.QueryKLines(ctx, symbol, interval, types.KLineQueryOptions{
-			EndTime: &endTime,
-			Limit:   1000, // indicators need at least 100
-		})
-		if err != nil {
-			return err
-		}
+		var i int64
+		for i = 0; i < KLineLimit; i += 1000 {
```
c9s, line 424: KLineLimit is not used elsewhere?

pkg/bbgo/session.go
```diff
@@ -418,29 +420,35 @@ func (session *ExchangeSession) initSymbol(ctx context.Context, environ *Environ
...
+			var duration time.Duration = time.Duration(-i * int64(interval.Duration()))
+			e := endTime.Add(duration)
```
c9s, line 426: because it loads a closed=false klines? but I thought we have filtered them?

pkg/bbgo/session.go
```diff
@@ -418,29 +420,35 @@ func (session *ExchangeSession) initSymbol(ctx context.Context, environ *Environ
...
-		// update last prices by the given kline
-		lastKLine := kLines[len(kLines)-1]
-		if interval == types.Interval1m {
-			log.Infof("last kline %+v", lastKLine)
-			session.lastPrices[symbol] = lastKLine.Close
-		}
+			// update last prices by the given kline
+			lastKLine := kLines[len(kLines)-1]
+			if interval == types.Interval1m {
+				log.Infof("last kline %+v", lastKLine)
```
c9s, line 444: this log is just to verify the lastPrice, so we can remove it now

pkg/bbgo/session.go
c9s: updated to 516-527

pkg/bbgo/session.go
```diff
@@ -23,6 +23,8 @@ import (
 	"github.com/c9s/bbgo/pkg/util"
 )
 
+var KLineLimit int64 = 1000
```
c9s, line 26: Rename it to historyKLineLimit ? and provide a setter `SetHistoryKLineLimit` ?

pkg/bbgo/source.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/c9s/bbgo/pkg/fixedpoint"
+	"github.com/c9s/bbgo/pkg/types"
+	log "github.com/sirupsen/logrus"
+)
+
+type SourceFunc func(*types.KLine) fixedpoint.Value
+
+var Four fixedpoint.Value = fixedpoint.NewFromInt(4)
+var Three fixedpoint.Value = fixedpoint.NewFromInt(3)
+var Two fixedpoint.Value = fixedpoint.NewFromInt(2)
```
c9s, line 15: make these values private? or move to the fixedpoint package?

pkg/bbgo/source.go
```diff
@@ -0,0 +1,50 @@
...
+
+type SourceSelector struct {
+	Source    string `json:"source,omitempty"`
+	getSource SourceFunc
```
c9s, line 19: call it `sourceGetter`?

pkg/bbgo/source.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (s *SourceSelector) Init() {
```
c9s, line 22: You can implement a UnmarshalJSON method, and after you unmarshal the json, you assign the getSource function, this way the Init() method call can be removed.

c9s (approved):

pkg/bbgo/session.go
zenixls2: set to KLinePreloadLimit.  
I think no `Set` function should be fine.

pkg/bbgo/source.go
zenixls2: moved to fixedpoint.

pkg/bbgo/source.go
zenixls2: done

pkg/bbgo/session.go
zenixls2: done

pkg/bbgo/source.go
zenixls2: done

PR #915, opened 2022-09-04
feature: add G-H filter and Kalman filter

COLDTURNIP:
Observer filters are realtime estimation algorithms to measure the status changes in a noisy environment. These recursive filtering algorithms estimate the true target states by minimize the uncertainties from system noise and observation noise. Such technologies are useful in financial econometrics as estimators without any latency, and more accurate than moving average based indicators.

The G-H filter in this PR implement John Ehlers' work. The system uncertainty is defined as the change of middle prices of K-lines, and the observation uncertainty is the price range of each K-line. Instead of fixed alpha & beta of conventional G-H filter, these two factor are calculated using an exponential recursive method controlled by configured window length.

More generalized than G-H filter, Kalman filter replaces these update factors by an adaptive factor Kalman gain, calculates the gain from uncertainties of estimation and observation, and optimize the estimation base on the gain, the measurement, and the previous prediction. In our implementation, we add an additional factor `AdditionalSmoothWindow` in need of smoothness, and set to 0 (classic Kalman filter) by default.

## Estimation Accurateness

In this PR, tests are created to compare the estimation accurateness between the two filters and EMA by calculating the square errors.

## References

- [Detailed explain for observer filters](https://www.kalmanfilter.net)
- [Wikipedia G-H filter](https://en.wikipedia.org/wiki/Alpha_beta_filter)
- [Wikipedia Kalman filter](https://en.wikipedia.org/wiki/Kalman_filter)
- [Optimal Tracking Filters By John Ehlers](https://www.jamesgoulding.com/Research_II/Ehlers/Ehlers%20(Optimal%20Tracking%20Filters).doc)


pkg/indicator/ghfilter_test.go
```diff
@@ -0,0 +1,6151 @@
...
+		return openDiff*openDiff + highDiff*highDiff + lowDiff*lowDiff + closeDiff*closeDiff
+	}
+	closeSquareError := func(base float64, k types.KLine) float64 {
+		closeDiff := math.Abs(k.Close.Float64() - base)
+		return closeDiff * closeDiff
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			filter := &GHFilter{IntervalWindow: types.IntervalWindow{Window: tt.args.window}}
+			ewma := &EWMA{IntervalWindow: types.IntervalWindow{Window: tt.args.window}}
```
COLDTURNIP, line 6121: EWMA is widely used to calculate average price with less latency compare to SMA. Do we expect the comparison of ZLEMA?

pkg/indicator/kalmanfilter_test.go
```diff
@@ -0,0 +1,6191 @@
...
+		return openDiff*openDiff + highDiff*highDiff + lowDiff*lowDiff + closeDiff*closeDiff
+	}
+	closeSquareError := func(base float64, k types.KLine) float64 {
+		closeDiff := math.Abs(k.Close.Float64() - base)
+		return closeDiff * closeDiff
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			filter := &KalmanFilter{IntervalWindow: types.IntervalWindow{Window: tt.args.window}}
+			ewma := &EWMA{IntervalWindow: types.IntervalWindow{Window: tt.args.window}}
```
COLDTURNIP, line 6155: Same here, change to ZLEMA?

austin362667 (approved):

pkg/indicator/ghfilter.go
```diff
@@ -0,0 +1,96 @@
...
+	return inc.Values.Last()
+}
+
+var _ types.SeriesExtend = &GHFilter{}
+
+func (inc *GHFilter) PushK(k types.KLine) {
+	inc.update(k.Close.Float64(), k.High.Float64()-k.Low.Float64())
+}
+
+func (inc *GHFilter) CalculateAndUpdate(allKLines []types.KLine) {
```
c9s, line 74: this is a legacy method, you can remove it

pkg/indicator/ghfilter.go
```diff
@@ -0,0 +1,96 @@
...
+		inc.EmitUpdate(inc.Last())
+		return
+	}
+	for _, k := range allKLines {
+		inc.PushK(k)
+		inc.EmitUpdate(inc.Last())
+	}
+}
+
+func (inc *GHFilter) handleKLineWindowUpdate(interval types.Interval, window types.KLineWindow) {
```
c9s, line 87: legacy method too

pkg/indicator/kalmanfilter.go
```diff
@@ -0,0 +1,108 @@
...
+	return inc.Values.Last()
+}
+
+var _ types.SeriesExtend = &KalmanFilter{}
+
+func (inc *KalmanFilter) PushK(k types.KLine) {
+	inc.update(k.Close.Float64(), (k.High.Float64()-k.Low.Float64())/2)
+}
+
+func (inc *KalmanFilter) CalculateAndUpdate(allKLines []types.KLine) {
```
c9s, line 86: legacy method~ you can remove it

pkg/indicator/kalmanfilter.go
```diff
@@ -0,0 +1,108 @@
...
+		inc.EmitUpdate(inc.Last())
+		return
+	}
+	for _, k := range allKLines {
+		inc.PushK(k)
+		inc.EmitUpdate(inc.Last())
+	}
+}
+
+func (inc *KalmanFilter) handleKLineWindowUpdate(interval types.Interval, window types.KLineWindow) {
```
c9s, line 99: legacy method too~ you can remove it

c9s:
You can add these two indicators to the standardIndicator struct

PR #919, opened 2022-09-07
feature: add fixedpoint.Reducer, Counter and add update stats method on TradeStats

c9s:
This counts profits by orderID

### DIFF
--- a/pkg/fixedpoint/slice.go
+++ b/pkg/fixedpoint/slice.go
@@ -1,0 +1,7 @@
+package fixedpoint
+
+type Slice []Value
+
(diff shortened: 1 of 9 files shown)


narumiruna (approved):

pkg/fixedpoint/slice.go
```diff
@@ -0,0 +1,24 @@
+package fixedpoint
+
+type Slice []Value
```
narumiruna, line 3: move to `pkg/types/value_slice.go`?
[pkg/types/value_map.go](https://github.com/c9s/bbgo/blob/7225a597f205b151ad5dc6837bd55ff8a7d58435/pkg/types/value_map.go)

PR #927, opened 2022-09-09
refactor: simplify submit order

c9s:
all: simplify underlying exchange submitOrder method

- Replace SubmitOrders with SubmitOrder
- Accept only one submit order and return one created order
- Add bbgo.BatchPlaceOrders helper method and bbgo.BatchRetryPlaceOrders method

### DIFF
--- a/examples/exchange-api/binance-margin/main.go
+++ b/examples/exchange-api/binance-margin/main.go
@@ -94,7 +94,7 @@ var rootCmd = &cobra.Command{
 
 		time.Sleep(time.Second)
 
-		createdOrders, err := exchange.SubmitOrders(ctx, types.SubmitOrder{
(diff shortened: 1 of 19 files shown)


examples/exchange-api/binance-margin/main.go
```diff
@@ -94,7 +94,7 @@ var rootCmd = &cobra.Command{
 
 		time.Sleep(time.Second)
 
-		createdOrders, err := exchange.SubmitOrders(ctx, types.SubmitOrder{
+		createdOrders, err := exchange.SubmitOrder(ctx, types.SubmitOrder{
```
zenixls2, line 97: createdOrder

c9s: updated

PR #929, opened 2022-09-10
feature: strategy/supertrend: draw pnl on

frin1:
- Add pnl and cumpnl drawing for backtesting
- Pnl and cumpnl commands for tg

### DIFF
--- a/pkg/strategy/supertrend/draw.go
+++ b/pkg/strategy/supertrend/draw.go
@@ -11,9 +11,9 @@ import (
 	"github.com/wcharczuk/go-chart/v2"
 )
 
-func (s *Strategy) InitDrawCommands(store *bbgo.SerialMarketDataStore, profit, cumProfit types.Series) {
+func (s *Strategy) InitDrawCommands(profit, cumProfit types.Series) {
 	bbgo.RegisterCommand("/pnl", "Draw PNL(%) per trade", func(reply interact.Reply) {
-		canvas := s.DrawPNL(profit)
+		canvas := DrawPNL(s.InstanceID(), profit)
 		var buffer bytes.Buffer
 		if err := canvas.Render(chart.PNG, &buffer); err != nil {
 			log.WithError(err).Errorf("cannot render pnl in drift")
@@ -23,7 +23,7 @@ func (s *Strategy) InitDrawCommands(store *bbgo.SerialMarketDataStore, profit, c
 		bbgo.SendPhoto(&buffer)
 	})
 	bbgo.RegisterCommand("/cumpnl", "Draw Cummulative PNL(Quote)", func(reply interact.Reply) {
-		canvas := s.DrawCumPNL(cumProfit)
+		canvas := DrawCumPNL(s.InstanceID(), cumProfit)
 		var buffer bytes.Buffer
 		if err := canvas.Render(chart.PNG, &buffer); err != nil {
 			log.WithError(err).Errorf("cannot render cumpnl in drift")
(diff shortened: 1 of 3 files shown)


c9s (dismissed):

andycheng123 (approved):

pkg/strategy/supertrend/strategy.go
```diff
@@ -413,6 +424,11 @@ func (s *Strategy) calculateQuantity(ctx context.Context, currentPrice fixedpoin
 	}
 }
 
+func (s *Strategy) CalcAssetValue(price fixedpoint.Value) fixedpoint.Value {
```
c9s, line 427: You can use this method to calculate net asset `func (m BalanceMap) Assets(prices map[string]fixedpoint.Value, priceTime time.Time) AssetMap {`

pkg/strategy/supertrend/strategy.go
```diff
@@ -478,14 +494,59 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
 
 			s.AccumulatedProfitReport.RecordProfit(profit.Profit)
 		})
-		s.orderExecutor.TradeCollector().OnTrade(func(trade types.Trade, profit fixedpoint.Value, netProfit fixedpoint.Value) {
-			s.AccumulatedProfitReport.RecordTrade(trade.Fee)
-		})
+		// s.orderExecutor.TradeCollector().OnTrade(func(trade types.Trade, profit fixedpoint.Value, netProfit fixedpoint.Value) {
+		// 	s.AccumulatedProfitReport.RecordTrade(trade.Fee)
+		// })
```
c9s, line 499: wonder the reason to remove it? maybe just remove the block if it's not used anymore?

pkg/strategy/supertrend/strategy.go
```diff
@@ -478,14 +494,59 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
...
+			s.buyPrice = 0
+			s.highestPrice = 0
+			s.lowestPrice = s.sellPrice
+		}
+	})
+
+	// event trigger order: s.Interval => Interval1m
+	store, ok := session.SerialMarketDataStore(s.Symbol, []types.Interval{s.Interval, types.Interval1m})
+	if !ok {
+		panic("cannot get 1m history")
```
c9s, line 546: since it's the run method, you can simply return error, by `errors.New` or `fmt.Errorf`

pkg/strategy/supertrend/draw.go
```diff
@@ -0,0 +1,91 @@
...
+			if vf, isFloat := v.(float64); isFloat {
+				return fmt.Sprintf("%.4f", vf)
+			}
+			return ""
+		},
+	}
+	return canvas
+}
+
+func (s *Strategy) Draw(store *bbgo.SerialMarketDataStore, profit, cumProfit types.Series) {
```
c9s, line 68: Ths `store` varaible is not used? can we remove it?

pkg/strategy/supertrend/draw.go
```diff
@@ -0,0 +1,91 @@
...
+
+	canvas := s.DrawPNL(profit)
+	f, err := os.Create(s.GraphPNLPath)
+	if err != nil {
+		log.WithError(err).Errorf("cannot create on path " + s.GraphPNLPath)
+		return
```
c9s, line 74: it's better to return error and log the error from the caller.

pkg/strategy/supertrend/draw.go
```diff
@@ -0,0 +1,91 @@
...
+			return
+		}
+		bbgo.SendPhoto(&buffer)
+	})
+}
+
+func (s *Strategy) DrawPNL(profit types.Series) *types.Canvas {
+	canvas := types.NewCanvas(s.InstanceID())
+	length := profit.Length()
+	log.Errorf("pnl Highest: %f, Lowest: %f", types.Highest(profit, length), types.Lowest(profit, length))
```
c9s, line 40: use info log instead ?

pkg/strategy/supertrend/draw.go
```diff
@@ -0,0 +1,91 @@
...
+				return fmt.Sprintf("%.4f", vf)
+			}
+			return ""
+		},
+	}
+	canvas.PlotRaw("1", types.NumberSeries(1), length)
+	return canvas
+}
+
+func (s *Strategy) DrawCumPNL(cumProfit types.Series) *types.Canvas {
```
c9s, line 54: pass instanceID from the func argument so this can be a simple function?

pkg/strategy/supertrend/draw.go
```diff
@@ -0,0 +1,91 @@
...
+		if err := canvas.Render(chart.PNG, &buffer); err != nil {
+			log.WithError(err).Errorf("cannot render cumpnl in drift")
+			reply.Message(fmt.Sprintf("[error] canot render cumpnl in drift: %v", err))
+			return
+		}
+		bbgo.SendPhoto(&buffer)
+	})
+}
+
+func (s *Strategy) DrawPNL(profit types.Series) *types.Canvas {
```
c9s, line 37: pass instanceID from the func argument so this can be a simple function?

pkg/strategy/supertrend/strategy.go
frin1: Removed (OnTrade is still in use, but not only for backtesting)

pkg/strategy/supertrend/strategy.go
frin1: Done

pkg/strategy/supertrend/draw.go
frin1: Done

pkg/strategy/supertrend/draw.go
frin1: Done

pkg/strategy/supertrend/draw.go
frin1: Done

pkg/strategy/supertrend/draw.go
frin1: Done

pkg/strategy/supertrend/draw.go
frin1: Done

c9s (approved):
Thanks!

andycheng123:
@c9s I think this should be a common package and called by interact. It has a bug. when i have several supertrend strategy running at the same time, the /pnl command only responded by one of them

c9s:
I agree, we need more refactoring

PR #930, opened 2022-09-11
Fix: Pivotshort 

andycheng123:
Fix trend EMA and make strategy controller work

### DIFF
--- a/pkg/bbgo/trend_ema_test.go
+++ b/pkg/bbgo/trend_ema_test.go
@@ -1,0 +1,21 @@
+package bbgo
+
+import (
+	"testing"
+
+	"github.com/c9s/bbgo/pkg/types"
+)
+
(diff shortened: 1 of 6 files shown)


pkg/bbgo/trend_ema_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+func Test_TrendEMA(t *testing.T) {
+	t.Run("Test Trend EMA", func(t *testing.T) {
+		trendEMA_test := TrendEMA{
+			IntervalWindow: types.IntervalWindow{Window: 1},
+		}
+		trendEMA_test.last = 1000.0
+		trendEMA_test.current = 1200.0
+
+		if trendEMA_test.Gradient() != 1.2 {
```
c9s, line 17: you can use assert.Equal

c9s (approved):

PR #942, opened 2022-09-13
feature: add modify tg command. fix wdrift ma length

zenixls2:
add `/modify` command to modify strategy fields that is attributed with both `json` and `modifiable:"true"`.  
also will show all the modifiable keys as button in tg.  
FYI.
to pass strings to the `interact`, users need to type `\"string\"` in tg.

### DIFF
--- a/pkg/bbgo/config.go
+++ b/pkg/bbgo/config.go
@@ -325,8 +325,6 @@ type Config struct {
 	CrossExchangeStrategies []CrossExchangeStrategy `json:"-" yaml:"-"`
 
 	PnLReporters []PnLReporterConfig `json:"reportPnL,omitempty" yaml:"reportPnL,omitempty"`
-
-	ConfigFile string
 }
 
 func (c *Config) Map() (map[string]interface{}, error) {
@@ -479,10 +477,6 @@ func Load(configFile string, loadStrategies bool) (*Config, error) {
 
 	if err := yaml.Unmarshal(content, &config); err != nil {
 		return nil, err
-	}
-	config.ConfigFile = configFile
-	for _, session := range config.Sessions {
-		session.ConfigPath = config.ConfigFile
 	}
 
 	// for backward compatible
--- a/pkg/bbgo/config.go
+++ b/pkg/bbgo/config.go
@@ -325,6 +325,8 @@ type Config struct {
 	CrossExchangeStrategies []CrossExchangeStrategy `json:"-" yaml:"-"`
 
 	PnLReporters []PnLReporterConfig `json:"reportPnL,omitempty" yaml:"reportPnL,omitempty"`
+
+	ConfigFile string
 }
 
 func (c *Config) Map() (map[string]interface{}, error) {
@@ -477,6 +479,10 @@ func Load(configFile string, loadStrategies bool) (*Config, error) {
 
 	if err := yaml.Unmarshal(content, &config); err != nil {
 		return nil, err
+	}
+	config.ConfigFile = configFile
+	for _, session := range config.Sessions {
+		session.ConfigPath = config.ConfigFile
 	}
 
 	// for backward compatible
--- a/pkg/bbgo/interact_modify.go
+++ b/pkg/bbgo/interact_modify.go
@@ -4,8 +4,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
-	"strings"
 
+	"github.com/c9s/bbgo/pkg/dynamic"
 	"github.com/c9s/bbgo/pkg/interact"
 	"github.com/c9s/bbgo/pkg/util"
 	log "github.com/sirupsen/logrus"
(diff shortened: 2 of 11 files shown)


zenixls2:
`/save` command is still under design. only field values in the config file are modifiable.
comments should be kept as before.

pkg/bbgo/config.go
```diff
@@ -478,6 +480,10 @@ func Load(configFile string, loadStrategies bool) (*Config, error) {
 	if err := yaml.Unmarshal(content, &config); err != nil {
 		return nil, err
 	}
+	config.ConfigFile = configFile
+	for _, session := range config.Sessions {
+		session.ConfigPath = config.ConfigFile
+	}
```
c9s, line 486: Can you pull out the config things to another PR? if you would like this PR to get merged first.

I think the config saving stuff needs to be designed carefully.

pkg/bbgo/modify.go
```diff
@@ -0,0 +1,74 @@
...
+		val = val.Elem()
+	}
+	var targetName string
+	var currVal interface{}
+	var mapping map[string]string
+	// currently we only allow users to modify the first layer of fields
+	RegisterCommand("/modify", "Modify config", func(reply interact.Reply) {
+		reply.Message("Please choose the field name in config to modify:")
+		mapping = make(map[string]string)
+		for i := 0; i < val.Type().NumField(); i++ {
```
c9s, line 26: can you pull out this to the dynamic.GetModifiableFields or dynamic.GetChangableFields function?

pkg/bbgo/modify.go
```diff
@@ -0,0 +1,74 @@
+package bbgo
```
c9s, line 1: will be better to put this in the file `interact_modify`

pkg/bbgo/modify.go
```diff
@@ -0,0 +1,74 @@
...
+func RegisterModifier(s interface{}) {
+	val := reflect.ValueOf(s)
+	if val.Type().Kind() == util.Pointer {
+		val = val.Elem()
+	}
+	var targetName string
+	var currVal interface{}
+	var mapping map[string]string
+	// currently we only allow users to modify the first layer of fields
+	RegisterCommand("/modify", "Modify config", func(reply interact.Reply) {
```
c9s, line 23: wondering if we should use `change` instead of `modify`?

pkg/bbgo/session.go
```diff
@@ -40,6 +40,7 @@ type ExchangeSession struct {
 	Secret       string             `json:"secret,omitempty" yaml:"secret,omitempty"`
 	Passphrase   string             `json:"passphrase,omitempty" yaml:"passphrase,omitempty"`
 	SubAccount   string             `json:"subAccount,omitempty" yaml:"subAccount,omitempty"`
+	ConfigPath   string
```
c9s, line 43: ditto, Can you pull out the config things to another PR? if you would like this PR to get merged first.

I think the config saving stuff needs to be designed carefully.

pkg/strategy/drift/driftma.go
```diff
@@ -14,7 +14,13 @@ type DriftMA struct {
 
 func (s *DriftMA) Update(value, weight float64) {
 	s.ma1.Update(value)
+	if s.ma1.Length() == 0 {
+		return
+	}
 	s.drift.Update(s.ma1.Last(), weight)
+	if s.drift.Length() == 0 {
+		return
+	}
```
c9s, line 23: would be great if this can be a single PR, this looks like a fix?

pkg/bbgo/modify.go
zenixls2: done

pkg/bbgo/config.go
zenixls2: done

pkg/bbgo/session.go
zenixls2: done

pkg/strategy/drift/driftma.go
zenixls2: I would like to merge this fix together

pkg/bbgo/modify.go
zenixls2: maybe still modify?
notice that we also have the cmd 'modifyposition'

pkg/bbgo/modify.go
zenixls2: done

pkg/bbgo/interact_modify.go
```diff
@@ -0,0 +1,70 @@
...
+			reply.Message(fmt.Sprintf("target %s is not modifiable", targetName))
+			return
+		}
+		x := reflect.New(field.Type())
+		xi := x.Interface()
+		if err := json.Unmarshal([]byte(value), &xi); err != nil {
+			reply.Message(fmt.Sprintf("fail to unmarshal the value: %s, err: %v", value, err))
+			return
+		}
+		field.Set(x.Elem())
```
c9s, line 59: Need to take care about the Pointer field and Value field, here it looks like only for pointer field? can we add a function for this set operation and add a test for it?

pkg/bbgo/interact_modify.go
```diff
@@ -0,0 +1,70 @@
...
+		newVal := field.Interface()
+		if e, err := json.Marshal(value); err == nil {
+			newVal = string(e)
+		}
+		reply.Message(fmt.Sprintf("update to %v successfully", newVal))
+	})
+
+	RegisterCommand("/save", "Save config", func(reply interact.Reply) {
```
c9s, line 67: let's move this save config thing to the next PR

pkg/dynamic/field.go
```diff
@@ -24,3 +27,39 @@ func LookupSymbolField(rs reflect.Value) (string, bool) {
...
+		callback(name, t.Name)
+	}
+}
+
+var ZeroValue reflect.Value = reflect.Zero(reflect.TypeOf(0))
+
+func GetModifiableField(val reflect.Value, name string) (reflect.Value, bool) {
+	field, ok := val.Type().FieldByName(name)
+	if !ok {
+		return ZeroValue, ok
```
c9s, line 55: it's better to create the zero reflect value from it's field type instead of `0`

pkg/strategy/drift/strategy.go
```diff
@@ -572,35 +560,39 @@ func (s *Strategy) klineHandler1m(ctx context.Context, kline types.KLine) {
...
 	numPending := 0
+	var err error
 	if numPending, err = s.smartCancel(ctx, pricef, atr); err != nil {
 		log.WithError(err).Errorf("cannot cancel orders")
+		s.positionLock.Unlock()
 		return
 	}
 	if numPending > 0 {
-		log.Infof("pending orders: %d, exit", numPending)
+		s.positionLock.Unlock()
```
c9s, line 587: types.Position has it's lock, maybe you can reuse it?

pkg/bbgo/interact_modify.go
```diff
@@ -0,0 +1,70 @@
...
+func RegisterModifier(s interface{}) {
+	val := reflect.ValueOf(s)
+	if val.Type().Kind() == util.Pointer {
+		val = val.Elem()
+	}
+	var targetName string
+	var currVal interface{}
+	var mapping map[string]string
+	// currently we only allow users to modify the first layer of fields
+	RegisterCommand("/modify", "Modify config", func(reply interact.Reply) {
```
c9s, line 23: You need to let the user to choose the strategy first

pkg/strategy/drift/strategy.go
```diff
@@ -980,6 +977,8 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
 		reply.Message(buffer.String())
 	})
 
+	bbgo.RegisterModifier(s)
```
c9s, line 980: multiple strategy instances will override the previous command registration.

pkg/bbgo/interact_modify.go
zenixls2: since x is created from reflect.New, as https://pkg.go.dev/reflect#New mentioned, the created `reflect.Value` must be a Pointer. That's why we only care about the pointer type here.

zenixls2: also added test cases

pkg/dynamic/field.go
zenixls2: non-existing field will not have type. any error will be presented in ok (the second return value)

pkg/strategy/drift/strategy.go
zenixls2: right now only two options:
1. tg vs strategy = 1 to 1 mapping
2. add this to coreInteract

since we currently have only drift applying this configuration, before any useful mechanism been discussed, I prefer to  keep it simple here and recommend users to try 1.

pkg/strategy/drift/strategy.go
zenixls2: I believe it's for different purpose.  
this position lock is mainly for price updates (buyPrice, sellPrice, lowestPrice, highestPrice), so should be fine.

pkg/bbgo/interact_modify.go
zenixls2: removed

c9s (approved):

PR #945, opened 2022-09-13
FEATURE: marketcap: get marketcap values from coinmarketcap

narumiruna:


pkg/strategy/marketcap/strategy.go
```diff
@@ -23,7 +23,7 @@ func init() {
 
 type Strategy struct {
 	Notifiability *bbgo.Notifiability
```
c9s, line 25: Notifiability can be removed

pkg/strategy/marketcap/strategy.go
```diff
@@ -23,7 +23,7 @@ func init() {
...
-	glassnode     *glassnode.DataSource
+	Datasource    *coinmarketcap.DataSource
```
c9s, line 26: this is not externally used, can be a private field?

pkg/strategy/marketcap/strategy.go
narumiruna: ok

c9s (approved):

PR #952, opened 2022-09-15
fix: fix profit stats calculation and since time field initialisation

c9s:


pkg/types/profit.go
```diff
@@ -194,12 +194,12 @@ func (s *ProfitStats) AddProfit(profit Profit) {
 	s.TodayPnL = s.TodayPnL.Add(profit.Profit)
 	s.TodayNetProfit = s.TodayNetProfit.Add(profit.NetProfit)
 
-	if profit.Profit.Sign() < 0 {
+	if profit.Profit.Sign() > 0 {
```
zenixls2, line 197: any reason why we make the change on this file? the result seems the same.

austin362667 (approved):

pkg/types/profit.go
c9s: To make it more intuitive to ready, process the positive profit first.

zenixls2 (approved):

PR #962, opened 2022-09-20
fix: add rate limit on telegram api and split messages by unicode

zenixls2:
- add rate limit in telegram.
- split messages by size limit and unicode header.

not strictly following the size limit (4096 bytes) and rate limit (100req/s) written in telegram api doc.  
use 3000, 10 instead.

### DIFF
--- a/pkg/interact/telegram.go
+++ b/pkg/interact/telegram.go
@@ -5,8 +5,8 @@ import (
 	"fmt"
 	"strings"
 	"time"
-	"unicode/utf8"
 
+	"github.com/c9s/bbgo/pkg/util"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/time/rate"
 	"gopkg.in/tucnak/telebot.v2"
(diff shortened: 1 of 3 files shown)


austin362667 (approved):

pkg/interact/telegram.go
```diff
@@ -62,7 +68,14 @@ type TelegramReply struct {
 }
 
 func (r *TelegramReply) Send(message string) {
-	checkSendErr(r.bot.Send(r.session.Chat, message))
+	var left, right int
+	for left, right = 0, maxMessageSize; right < len(message); left, right = right, right+maxMessageSize {
```
c9s, line 72: you only need to split the message when its length > maxMessageSize?

pkg/interact/telegram.go
```diff
@@ -158,7 +171,24 @@ func (tm *Telegram) Start(context.Context) {
 		if reply.set {
 			reply.build()
 			if len(reply.message) > 0 || reply.menu != nil {
-				checkSendErr(tm.Bot.Send(m.Chat, reply.message, reply.menu))
+				var left, right int
+				for left, right = 0, maxMessageSize; right < len(reply.message); left, right = right, right+maxMessageSize {
```
c9s, line 175: for the split thing, can you pull out it to a function?

pkg/interact/telegram.go
zenixls2: yes

pkg/interact/telegram.go
zenixls2: done

c9s (approved):

PR #963, opened 2022-09-20
Feature: limit how many metrics is shown by optimizer

frin1:


pkg/cmd/optimize.go
```diff
@@ -124,8 +130,12 @@ var optimizeCmd = &cobra.Command{
 					continue
 				}
 
+				if len(values) < resultLimit || resultLimit == 0 {
+					resultLimit = len(values)
+				}
```
c9s, line 135: select the slice only when the length exceeds the resultLimit?

if len(values) > resultLimit {
    values = values[:resultLimit]
}

frin1: done

c9s (approved):

PR #966, opened 2022-09-23
fix: binance futures position side and reduce only

oumeta:
binance websock add paperTrade


pkg/exchange/binance/exchange.go
```diff
@@ -1041,13 +1041,27 @@ func (e *Exchange) submitFuturesOrder(ctx context.Context, order types.SubmitOrd
 	if err != nil {
 		return nil, err
 	}
+	side := futures.PositionSideTypeBoth
+	if order.Side == types.SideTypeBuy {
```
c9s, line 1045: could you use switch case for the side buy and side sell here?

pkg/exchange/binance/exchange.go
```diff
@@ -1041,13 +1041,27 @@ func (e *Exchange) submitFuturesOrder(ctx context.Context, order types.SubmitOrd
 	if err != nil {
 		return nil, err
 	}
+	side := futures.PositionSideTypeBoth
```
c9s, line 1044: call it  positionSide to separate the order.Side?

pkg/exchange/binance/stream.go
```diff
@@ -300,6 +300,18 @@ func (s *Stream) getEndpointUrl(listenKey string) string {
...
 	}
+	if paperTrade() {
+		if s.IsFutures {
+			url = FuturesWebSocketTestURL + "/ws"
+		} else if isBinanceUs() {
+			url = WebSocketTestURL + "/ws"
+		} else {
+			url = WebSocketTestURL + "/ws"
+		}
+		fmt.Println("cao ni ma ")
```
c9s, line 311: remove this debug code?

PR #969, opened 2022-09-28
fix: reduce Quantity precheck, drift condition, ewo refactor

zenixls2:


zenixls2:
cumpnl
possible to have more smooth cumpnl.

PR #975, opened 2022-10-03
interval: add 1s support

austin362667:


pkg/types/interval.go
```diff
@@ -89,6 +98,36 @@ func ParseInterval(input Interval) int {
...
+		t *= 60 * 24 * 7
+	case "mo":
+		t *= 60 * 24 * 30
+	default:
+		panic("unknown input: " + input)
+	}
+	return t
+}
+
 var SupportedIntervals = map[Interval]int{
```
c9s, line 131: you need to fix this as well?

pkg/types/interval.go
```diff
@@ -17,8 +17,16 @@ func (i Interval) Minutes() int {
 	return m
 }
```
c9s, line 18: how about this `Minutes` function?

pkg/cmd/backtest.go
```diff
@@ -597,6 +589,27 @@ var BacktestCmd = &cobra.Command{
...
+func collectSubscriptionIntervals(backTestIntervals []types.Interval, requiredInterval types.Interval, environ *bbgo.Environment) (map[types.Interval]struct{}, types.Interval) {
+	allKLineIntervals := make(map[types.Interval]struct{})
+	for _, interval := range backTestIntervals {
+		allKLineIntervals[interval] = struct{}{}
+	}
+
+	for _, session := range environ.Sessions() {
+		for _, sub := range session.Subscriptions {
+			if sub.Channel == types.KLineChannel {
+				if sub.Options.Interval == types.Interval1s {
```
zenixls2, line 601: I would recommend using sub.Options.Interval.Seconds() % 60 > 0 as the condition.

pkg/types/interval.go
```diff
@@ -17,8 +17,16 @@ func (i Interval) Minutes() int {
...
+	m, ok := SupportedIntervals[i]
+	if !ok {
+		return ParseIntervalSeconds(i)
+	}
+	return m
+}
+
 func (i Interval) Duration() time.Duration {
-	return time.Duration(i.Minutes()) * time.Minute
+	return time.Duration(i.Seconds()) * time.Minute
```
zenixls2, line 29: * time.Second

pkg/types/interval.go
```diff
@@ -89,6 +98,36 @@ func ParseInterval(input Interval) int {
 	return t
 }
 
+func ParseIntervalSeconds(input Interval) int {
```
zenixls2, line 101: Should provide ParseInterval instead of ParseIntervalSeconds.
All result from now on should be based on Seconds.

pkg/cmd/backtest.go
austin362667: what if it's 1h?

pkg/cmd/backtest.go
```diff
@@ -597,6 +585,32 @@ var BacktestCmd = &cobra.Command{
...
+		allKLineIntervals[interval] = struct{}{}
+	}
+	// default interval is 1m for all exchanges
+	requiredInterval = types.Interval1m
+	for _, session := range environ.Sessions() {
+		for _, sub := range session.Subscriptions {
+			if sub.Channel == types.KLineChannel {
+				if sub.Options.Interval.Seconds()%60 > 0 {
+					// if any subscription is 1s, then we will use 1s for back-testing
+					requiredInterval = sub.Options.Interval
```
zenixls2, line 604: use `types.Interval1s` here

zenixls2: notice that `Options.Interval = "3s"` is also valid since it's parsable.

pkg/types/interval.go
```diff
@@ -10,6 +10,14 @@ import (
 type Interval string
 
 func (i Interval) Minutes() int {
+	m, ok := SupportedIntervals[i]
+	if !ok {
+		return int(ParseInterval(i) / 60.)
+	}
+	return m
```
zenixls2, line 17: m / 60

pkg/types/interval.go
```diff
@@ -10,6 +10,14 @@ import (
 type Interval string
 
 func (i Interval) Minutes() int {
+	m, ok := SupportedIntervals[i]
+	if !ok {
+		return int(ParseInterval(i) / 60.)
```
zenixls2, line 15: return ParseInterval(i) / 60

pkg/backtest/exchange.go
```diff
@@ -369,14 +369,14 @@ func (e *Exchange) ConsumeKLine(k types.KLine) {
 		matching.klineCache = make(map[types.Interval]types.KLine)
 	}
 
-	kline1m, ok := matching.klineCache[k.Interval]
+	requiredKline, ok := matching.klineCache[k.Interval]
 	if ok { // pop out all the old
-		if kline1m.Interval != types.Interval1m {
-			panic("expect 1m kline, got " + kline1m.Interval.String())
+		if requiredKline.Interval.Seconds() < requiredInterval.Seconds() {
```
zenixls2, line 374: use `!=` here.
The order of kline is already restricted by SQL, so the first popped out kline should be the same as the requiredInterval

austin362667: Aalready tried. Not working

pkg/cmd/backtest.go
austin362667: https://github.com/c9s/bbgo/pull/975#discussion_r986441789

pkg/backtest/exchange.go
austin362667: should re-sync databse

c9s (approved):

PR #980, opened 2022-10-04
fix: support 1s interval, add binance request rate limiter

zenixls2:


pkg/bbgo/session.go
```diff
@@ -497,12 +501,13 @@ func (session *ExchangeSession) Positions() map[string]*types.Position {
...
 	s, ok = session.marketDataStores[symbol]
-	if !ok {
+	// FIXME: the returned MarketDataStore when !ok will be empty
+	/*if !ok {
 		s = NewMarketDataStore(symbol)
 		s.BindStream(session.MarketDataStream)
 		session.marketDataStores[symbol] = s
 		return s, true
-	}
+	}*/
```
c9s, line 510: why not initialize a new market data store when it's not found?

pkg/bbgo/activeorderbook.go
```diff
@@ -92,60 +92,50 @@ func (b *ActiveOrderBook) waitAllClear(ctx context.Context, waitTime, timeout ti
...
 // GracefulCancel cancels the active orders gracefully
-func (b *ActiveOrderBook) GracefulCancel(ctx context.Context, ex types.Exchange) error {
+func (b *ActiveOrderBook) GracefulCancel(ctx context.Context, ex types.Exchange, orders ...types.Order) error {
+	// if no orders are given, set to cancelAll
+	if len(orders) == 0 {
+		orders = b.Orders()
+	} else {
+		// simple check on given input
+		for _, o := range orders {
+			if o.Symbol != b.Symbol {
```
c9s, line 132: make sure `b.Symbol != ""`

pkg/bbgo/activeorderbook.go
```diff
@@ -168,6 +156,10 @@ func (b *ActiveOrderBook) GracefulCancel(ctx context.Context, ex types.Exchange)
 		}
 
 		log.Debugf("[ActiveOrderBook] waiting %s for %s orders to be cancelled...", waitTime, b.Symbol)
+		for _, o := range b.Orders() {
+			b.Remove(o)
+		}
+		break
```
c9s, line 162: why break here?

pkg/bbgo/activeorderbook.go
```diff
@@ -92,60 +92,50 @@ func (b *ActiveOrderBook) waitAllClear(ctx context.Context, waitTime, timeout ti
 	}
 }
 
-// Cancel cancels the given order from activeOrderBook gracefully
-func (b *ActiveOrderBook) Cancel(ctx context.Context, ex types.Exchange, order types.Order) error {
-	if !b.Exists(order) {
-		return fmt.Errorf("cannot find %v in orderbook", order)
+func (b *ActiveOrderBook) CancelNoWait(ctx context.Context, ex types.Exchange, orders ...types.Order) error {
```
c9s, line 95: can you convert the NoWait  to an option defined in ActiveOrderBook? like activeOrders.SetNoWait(true)

zenixls2: we might need both operations on the same activeOrderbook in a single strategy.  
for example, in emergency exit, we could like all the orders to be successfully cancelled, so we call GracefulCancel, while in high frequency trading, we still need CancelNoWait.

pkg/bbgo/activeorderbook.go
zenixls2: testing code not yet been cleaned up

pkg/exchange/binance/parse.go
```diff
@@ -350,6 +350,11 @@ func parseWebSocketEvent(message []byte) (interface{}, error) {
 		err = json.Unmarshal([]byte(message), &event)
 		return &event, err
 
+	case "aggTrade":
```
c9s, line 353: can we have aggTrade in one single PR?

c9s:
conflict~

PR #986, opened 2022-10-07
fix: general order executor: ClosePosition() works on futures position

andycheng123:


pkg/bbgo/order_executor_general.go
```diff
@@ -407,22 +407,32 @@ func (e *GeneralOrderExecutor) ClosePosition(ctx context.Context, percentage fix
...
-			submitOrder.Quantity = fixedpoint.Min(submitOrder.Quantity, baseBalance.Available)
-		}
-		if submitOrder.Quantity.IsZero() {
-			return fmt.Errorf("insufficient base balance, can not sell: %+v", submitOrder)
+	if e.session.Futures { // Futures: Use base qty in e.position
+		submitOrder.Quantity = e.position.GetBase().Abs()
+		submitOrder.ReduceOnly = true
+		if e.position.IsLong() {
+			submitOrder.Side = types.SideTypeSell
+		} else {
```
c9s, line 415: check e.position.IsShort() and avoid dust quantity?

c9s (approved):

pkg/bbgo/order_executor_general.go
andycheng123: For futures, the dust position still can be closed if the ordeer is reduced only

PR #1048, opened 2023-01-10
feature: grid2: integrate prometheus metrics

c9s:


pkg/strategy/grid2/strategy.go
```diff
@@ -854,9 +873,30 @@ func (s *Strategy) openGrid(ctx context.Context, session *bbgo.ExchangeSession)
 
 	s.logger.Infof("ALL GRID ORDERS SUBMITTED")
 	s.EmitGridReady()
+
+	s.updateOpenOrderPricesMetrics(createdOrders)
 	return nil
 }
 
+func (s *Strategy) updateOpenOrderPricesMetrics(orders []types.Order) {
```
gx578007, line 881: Need to delete existing values of the same metric and add new values from scratch.

gx578007 (approved):

pkg/strategy/grid2/strategy.go
c9s: fixed, added Reset call

PR #1050, opened 2023-01-14
Feature: New indicators

frin1:
Added indicators from tradingview and added function to existing supertrend indicator.
New indicators:
- UT Bot Alerts
- Pivot Point Supertrend


austin362667:
@andycheng123  please check supertrend indicator

austin362667:
@andycheng123  please check supertrend indicator

austin362667:
ref: [Quantnomad UT Bot in Python](https://quantnomad.com/implementing-ut-bot-strategy-in-python-with-vectorbt/)

c9s:
could you please rebase the commits? thanks!


git fetch upstream
git rebase upstream/main
git push -f origin HEAD

frin1:
> 

Rebase done

pkg/indicator/supertrend.go
```diff
@@ -128,6 +134,21 @@ func (inc *Supertrend) GetSignal() types.Direction {
 	return inc.tradeSignal
 }
 
+// GetDirection returns current trend
+func (inc *Supertrend) GetDirection() types.Direction {
```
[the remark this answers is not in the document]
frin1: Fixed. I've unintentionally removed zenixls2 and andycheng123 as reviewers. Could you please re-add them?

pkg/indicator/supertrendPivot.go
```diff
@@ -0,0 +1,217 @@
...
+type PivotSupertrend struct {
+	types.SeriesBase
+	types.IntervalWindow
+	ATRMultiplier float64 `json:"atrMultiplier"`
+	PivotWindow   int     `json:"pivotWindow"`
+
+	AverageTrueRange *ATR // Value must be set when initialized in strategy
+
+	PivotLow  *PivotLow  // Value must be set when initialized in strategy
+	PivotHigh *PivotHigh // Value must be set when initialized in strategy
```
andycheng123, line 27: I don't see the reason why the pivots should be here with supertrend instead of standalone indicator

frin1: The Update function is not the same for supertrend and supertrendPivot. There is not a big difference, but I've tried to keep it similar to the tradingview script (Pivot Point Supertrend by LonesomeTheBlue).

andycheng123: what i mean is that in this supertrendPivot, supertrend and pivot high/low seems irrelevant. Maybe pivot high/low can be a standalone indicator? for the supertrend part,  if it's better than the original supertrend indicator, you can improve the supertrend indicator. Or is there any reason to put both pivot and supertrend into an indicator?

frin1: supertrendPivot calculates the support and resistance lines based on the pivot high/low (not closePrice). SupertrendPivot is not necessarily better than the original, but the result is a bit different as the original supertrend is calculated based on close price. When using pivot high/low instead of close price, it typically gives a bit fewer signals. Also, supertrendPivot is weighting the src. Maybe I'm overlooking something, but I can't see how a standalone pivot high/low indicator and supertrend could generate the same output?

andycheng123: oh, i see it now. Sorry that i missed the inc.src part. It uses last pivot and last src to calculate the new src, right?

andycheng123 (approved):

c9s (approved):

pkg/indicator/supertrend.go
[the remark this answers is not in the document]
c9s: Ok

PR #1060, opened 2023-02-17
fix: process pending order update for active order book

c9s:


pkg/bbgo/order_execution.go
```diff
@@ -62,6 +62,33 @@ func BatchRetryPlaceOrder(ctx context.Context, exchange types.Exchange, errIdx [
 	return createdOrders, err
 }
 
+// BatchPlaceOrderChan post orders with a channel, the created order will be sent to this channel immediately, so that
+// the caller can add the created order to the active order book or the order store to collect trades.
+// this method is used when you have large amount of orders to be sent and most of the orders might be filled as taker order.
+// channel orderC will be closed when all the submit orders are submitted.
+func BatchPlaceOrderChan(ctx context.Context, exchange types.Exchange, orderC chan types.Order, submitOrders ...types.SubmitOrder) (types.OrderSlice, []int, error) {
```
c9s, line 69: this is unrelated, but we might need this in the future

pkg/bbgo/activeorderbook.go
```diff
@@ -277,25 +285,48 @@ func (b *ActiveOrderBook) Print() {
...
+		}
+
+		b.add(order)
+	}
+}
+
+// add the order to the active order book and check the pending order
+func (b *ActiveOrderBook) add(order types.Order) {
+	if pendingOrder, ok := b.pendingOrderUpdates.Get(order.OrderID); ok {
+		if pendingOrder.UpdateTime.Time().After(order.UpdateTime.Time()) {
```
gx578007, line 310: There's a rare condition that a pending order and a new coming order are updated at the same timestamp (in millisecond or even in second). 
Another way is to compare the filled amount or the state change.

gx578007 (approved):
Others LGTM

PR #1070, opened 2023-03-01
fix: add context, exponential backoff and max retry limit

c9s:


pkg/bbgo/order_execution.go
```diff
@@ -332,24 +334,36 @@ func BatchRetryPlaceOrder(ctx context.Context, exchange types.Exchange, errIdx [
...
 		// iterate the error index and re-submit the order
-		for _, idx := range errIdx {
+		log.Warnf("starting retry round #%d...", retryRound+1)
+		for _, idx := range errIdxNext {
 			submitOrder := submitOrders[idx]
 
 			op := func() error {
 				// can allocate permanent error backoff.Permanent(err) to stop backoff
-				createdOrder, err2 := exchange.SubmitOrder(ctx, submitOrder)
+				createdOrder, err2 := exchange.SubmitOrder(timeoutCtx, submitOrder)
```
ycdesu, line 364: The imeout of `timeoutCtx` is 30 minutes. Assuming that the first retrying action takes over 30 minutes, and all the remaining actions will since the timeoutCtx is already timed out.

We could create another timeout ctx in this retrying for loop.

c9s: yeah I would like to make the whole SubmitOrders including retry is controled under 30m

gx578007 (approved):

PR #1074, opened 2023-03-02
strategy: rebalance: add order type parameter

narumiruna:


config/rebalance.yaml
```diff
@@ -12,9 +12,9 @@ backtest:
   startTime: "2022-01-01"
   endTime: "2022-10-01"
   symbols:
-  - BTCUSDT
-  - ETHUSDT
-  - MAXUSDT
+    - BTCUSDT
+    - ETHUSDT
+    - MAXUSDT
```
c9s, line 17: would suggest the indentation you used on the left side.

pkg/strategy/rebalance/strategy.go
```diff
@@ -24,9 +25,9 @@ type Strategy struct {
 	QuoteCurrency string           `json:"quoteCurrency"`
 	TargetWeights types.ValueMap   `json:"targetWeights"`
 	Threshold     fixedpoint.Value `json:"threshold"`
+	MaxAmount     fixedpoint.Value `json:"maxAmount"` // max amount to buy or sell per order
+	OrderType     string           `json:"orderType"`
```
c9s, line 29: you can use types.OrderType type directly, it could be unmarshalled

narumiruna: ok

c9s (approved):

PR #1076, opened 2023-03-03
fix: grid2: quantity fee reduction for quote currency

c9s:


pkg/strategy/grid2/strategy.go
```diff
@@ -405,6 +427,11 @@ func (s *Strategy) processFilledOrder(o types.Order) {
 
 		// use the profit to buy more inventory in the grid
 		if s.Compound || s.EarnBase {
+			// if it's not using the platform fee currency, reduce the quote quantity for the buy order
+			if feeCurrency == s.Market.QuoteCurrency {
+				orderExecutedQuoteAmount = orderExecutedQuoteAmount.Sub(feeQuantityReduction)
```
gx578007, line 432: What if the fee currency is base currency?

pkg/strategy/grid2/strategy.go
```diff
@@ -414,19 +441,7 @@ func (s *Strategy) processFilledOrder(o types.Order) {
...
-		s.logger.Infof("GRID BUY ORDER BASE FEE: %s %s", baseSellQuantityReduction.String(), s.Market.BaseCurrency)
-
-		baseSellQuantityReduction = roundUpMarketQuantity(s.Market, baseSellQuantityReduction)
-		s.logger.Infof("GRID BUY ORDER BASE FEE (Rounding with precision %d): %s %s",
-			s.Market.VolumePrecision,
-			baseSellQuantityReduction.String(),
-			s.Market.BaseCurrency)
-
-		newQuantity = newQuantity.Sub(baseSellQuantityReduction)
+		newQuantity = newQuantity.Sub(feeQuantityReduction)
```
gx578007, line 446: Need to check fee is in quote or base currency here?

c9s: You're right

pkg/strategy/grid2/strategy.go
c9s: it's sell triggers buy order, so we only care about the quote currency here?

pkg/strategy/grid2/strategy.go
c9s: Fixed

gx578007 (approved):

PR #1077, opened 2023-03-03
FEATURE: save expiring data to redis

gx578007:


pkg/service/persistence.go
```diff
@@ -10,6 +12,10 @@ type Store interface {
 	Reset() error
 }
 
+type ExpiringData interface {
```
c9s, line 15: Expirable ?
https://en.wiktionary.org/wiki/expirable#:~:text=Adjective,being%20brought%20to%20an%20end.

pkg/service/persistence.go
```diff
@@ -10,6 +12,10 @@ type Store interface {
...
+	Expiration() time.Duration
```
c9s, line 16: use TimeToLive() or TTL()?

Expiration() looks like it will return a specific time point?

c9s (approved):

pkg/service/persistence.go
gx578007: I just followed the naming of redis client. It uses "expiration" as `time.Duration`.

gx578007:
polygon:0x8A19A182Af3f0090a678204e4E4D048C2C2715De

PR #1078, opened 2023-03-03
fix: grid2: improve the onStart callback and the cancel loop

c9s:


pkg/strategy/grid2/strategy.go
```diff
@@ -875,48 +875,48 @@ func (s *Strategy) cancelAll(ctx context.Context) error {
...
-			s.logger.Infof("found OrderGroupID (%d), using group ID for canceling orders...", s.OrderGroupID)
+		for {
+			s.logger.Infof("checking %s open orders...", s.Symbol)
 
-			op := func() error {
-				_, cancelErr := service.CancelOrdersByGroupID(ctx, s.OrderGroupID)
-				return cancelErr
-			}
-			err := backoff.Retry(op, backoff.WithMaxRetries(backoff.NewExponentialBackOff(), 101))
+			openOrders, err := session.Exchange.QueryOpenOrders(ctx, s.Symbol)
```
gx578007, line 881: Is this `ctx` constrained with a deadline?

c9s: not confirmed yet. maybe different when calling CleanUp()

pkg/strategy/grid2/strategy.go
```diff
@@ -875,48 +875,48 @@ func (s *Strategy) cancelAll(ctx context.Context) error {
...
 			if err != nil {
-				werr = multierr.Append(werr, err)
-			}
-		} else {
-			s.logger.Infof("canceling all orders...")
-			op := func() error {
-				_, cancelErr := service.CancelAllOrders(ctx)
-				return cancelErr
+				return err
```
gx578007, line 883: Could it fallback to cancel all orders anyway if having any error? Querying open orders is only for skipping when there is no open orders.

c9s: it's exp backoff, so if we have error, it will always retry

gx578007 (approved):

PR #1080, opened 2023-03-03
strategy: marketcap: add orderType parameter

narumiruna:


c9s (approved):

pkg/strategy/marketcap/strategy.go
```diff
@@ -77,6 +78,10 @@ func (s *Strategy) Validate() error {
 		return fmt.Errorf("maxAmount shoud not less than 0")
 	}
 
+	if s.OrderType == "" {
+		s.OrderType = types.OrderTypeLimitMaker
+	}
```
c9s, line 83: You can move this to `Defaults() error` method (minor)

narumiruna: 好喔

narumiruna: 好喔

PR #1083, opened 2023-03-06
FIX: add group id on submit order API

kbearXD:


pkg/exchange/max/exchange.go
```diff
@@ -480,7 +480,9 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
 		Side(toLocalSideType(o.Side)).
 		Volume(quantityString).
 		OrderType(orderType).
-		ClientOrderID(clientOrderID)
+		ClientOrderID(clientOrderID).
+		// TODO: make sure MAX API support uint32 group id
+		GroupID(fmt.Sprintf("%d", o.GroupID))
```
c9s, line 485: check if o.GroupID > 0

c9s (approved):

PR #1097, opened 2023-03-10
FEATURE: get filled orders when bbgo down

kbearXD:


pkg/strategy/grid2/strategy.go
```diff
@@ -1385,9 +1385,36 @@ func (s *Strategy) recoverGridWithOpenOrdersByScanningTrades(ctx context.Context
...
 
 	// 1. build pin-order map
-	// 2. fill the pin-order map by querying trades
+	pinOrderMapWithOpenOrders, err := s.buildPinOrderMap(grid, openOrdersOnGrid)
+	if err != nil {
+		return errors.Wrapf(err, "failed to build pin order map with open orders")
+	}
+
+	// 2. build the filled pin-order map by querying trades
+	pinOrderMapWithFilledOrders, err := s.buildFilledPinOrderMapFromTrades(ctx, historyService, pinOrderMapWithOpenOrders)
```
c9s, line 1394: create a type for `map[string]types.Order` ? and then you can move some methods to that type?

c9s, line 1394: you can simplify the var name, like `pinOrders`

pkg/strategy/grid2/strategy.go
```diff
@@ -1385,9 +1385,36 @@ func (s *Strategy) recoverGridWithOpenOrdersByScanningTrades(ctx context.Context
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to build filled pin order map")
+	}
+
 	// 3. get the filled orders from pin-order map
+	filledOrders := getOrdersFromPinOrderMapInAscOrder(pinOrderMapWithFilledOrders)
```
c9s, line 1400: so here,  can be simplified to `pinOrders.OrdersAscending()`

c9s: or `pinOrders.AscendingOrders()`

pkg/strategy/grid2/strategy.go
```diff
@@ -1385,9 +1385,36 @@ func (s *Strategy) recoverGridWithOpenOrdersByScanningTrades(ctx context.Context
...
+		i++
+	} else {
+		return fmt.Errorf("not reasonable num of filled orders")
+	}
+
 	// 4. emit the filled orders
+	activeOrderBook := s.orderExecutor.ActiveMakerOrders()
+	for ; i < numsFilledOrders; i++ {
+		filledOrder := filledOrders[i]
+		s.logger.Infof("[DEBUG] emit filled order: %s (%s)", filledOrder.String(), filledOrder.UpdateTime)
```
c9s, line 1415: you don't need this `[DEBUG]` flag?

pkg/strategy/grid2/strategy.go
```diff
@@ -1385,9 +1385,36 @@ func (s *Strategy) recoverGridWithOpenOrdersByScanningTrades(ctx context.Context
...
+		activeOrderBook.EmitFilled(filledOrder)
```
c9s, line 1416: might not be easy to write tests for this?

pkg/strategy/grid2/strategy.go
```diff
@@ -1428,6 +1455,96 @@ func (s *Strategy) buildPinOrderMap(grid *Grid, openOrders []types.Order) (map[s
...
+// buildFilledPinOrderMapFromTrades will query the trades from last 24 hour and use them to build a pin order map
+// It will skip the orders on pins at which open orders are already
+func (s *Strategy) buildFilledPinOrderMapFromTrades(ctx context.Context, historyService types.ExchangeTradeHistoryService, openPinOrderMap map[string]types.Order) (map[string]types.Order, error) {
+	filledPinOrderMap := make(map[string]types.Order)
+	var limit int64 = 1000
+	// get the filled orders when bbgo is down in order from trades
+	// [NOTE] only retrieve from last 24 hours !!!
+	var fromTradeID uint64 = 0
+	for {
+		trades, err := historyService.QueryTrades(ctx, s.Symbol, &types.TradeQueryOptions{
```
c9s, line 1467: maybe you can use BatchQueryTrades

pkg/strategy/grid2/strategy.go
kbearXD: Use this flag to find my log easily XD

pkg/strategy/grid2/strategy.go
kbearXD: Use this flag to find my log easily XD

pkg/strategy/grid2/strategy.go
kbearXD: I will keep it first to know it's only for filled orders.

pkg/strategy/grid2/strategy.go
kbearXD: I think it's not hard, we can OnFilled another assertion function.

pkg/strategy/grid2/strategy.go
```diff
@@ -53,6 +53,28 @@ func (pp PrettyPins) String() string {
...
 }
 
+// PinOrderMap store the pin-order's relation, we will change key from string to fixedpoint.Value when FormatString fixed
+type PinOrderMap map[string]types.Order
+
+// AscendingOrders get the orders from pin order map and sort it in asc order
+func (m PinOrderMap) AscendingOrders() []types.Order {
+	var orders []types.Order
+	for _, order := range m {
+		// skip empty order
```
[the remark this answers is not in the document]
gx578007: Oh, I got it. If there is no order at a pin, that instance will be zero.

pkg/strategy/grid2/strategy.go
```diff
@@ -1438,6 +1487,78 @@ func (s *Strategy) buildPinOrderMap(grid *Grid, openOrders []types.Order) (map[s
...
+				return nil, fmt.Errorf("the price of the order with the same GroupID is not in pins")
+			}
+
+			// skip open orders on grid
+			if v.OrderID != 0 {
+				continue
+			}
+
+			// check the order's creation time
+			if pinOrder, exist := filledPinOrderMap[priceStr]; exist && pinOrder.CreationTime.Time().After(order.CreationTime.Time()) {
```
gx578007, line 1544: How about comparing orders by order id ?

pkg/strategy/grid2/strategy.go
```diff
@@ -1438,6 +1487,78 @@ func (s *Strategy) buildPinOrderMap(grid *Grid, openOrders []types.Order) (map[s
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to query trades to recover the grid with open orders")
+		}
+
+		s.logger.Infof("[DEBUG] len of trades: %d", len(trades))
+
+		for _, trade := range trades {
+			order, err := s.orderQueryService.QueryOrder(ctx, types.OrderQuery{
+				OrderID: strconv.FormatUint(trade.OrderID, 10),
+			})
```
gx578007, line 1514: You can create a cache to store queried orders because multiple trades can be mapped to the same order id. The current implementation can cause duplicate queries.

gx578007: Also, to further speed up, if an order id is found in the open orders, you don't have to query it again and can skip current iteration as well.

pkg/strategy/grid2/strategy.go
```diff
@@ -1438,6 +1487,78 @@ func (s *Strategy) buildPinOrderMap(grid *Grid, openOrders []types.Order) (map[s
...
+
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to query order by trade")
```
gx578007, line 1517: What if the error is about rate limit?

gx578007: Maybe, it's better to sleep between queries?

pkg/strategy/grid2/strategy.go
gx578007: Some exchanges might impose order on order id so I think let's use creation time.

pkg/strategy/grid2/strategy.go
```diff
@@ -53,6 +53,28 @@ func (pp PrettyPins) String() string {
...
+		if order.OrderID == 0 {
+			continue
+		}
+
+		orders = append(orders, order)
+	}
+
+	sort.Slice(orders, func(i, j int) bool {
+		return orders[i].UpdateTime.Before(orders[j].UpdateTime.Time())
+	})
```
c9s, line 73: you can use types.SortOrdersUpdateTimeAscending

pkg/strategy/grid2/strategy.go
```diff
@@ -53,6 +53,28 @@ func (pp PrettyPins) String() string {
 	return fmt.Sprintf("%v", ss)
 }
 
+// PinOrderMap store the pin-order's relation, we will change key from string to fixedpoint.Value when FormatString fixed
+type PinOrderMap map[string]types.Order
```
c9s, line 57: move this to a single file?

pkg/strategy/grid2/strategy.go
c9s: then you can use `pinOrdersFilled` and `pinOrdersOpen`

pkg/strategy/grid2/strategy.go
c9s: Okay, remember to remove it when this PR is ready

pkg/strategy/grid2/strategy.go
c9s: yes, this might cause duplicated order queries

pkg/strategy/grid2/strategy.go
```diff
@@ -1413,10 +1462,10 @@ func (s *Strategy) recoverGridWithOpenOrdersByScanningTrades(ctx context.Context
...
 // The keys of this map contains all required pins of this grid.
 // If the Order of the pin is empty types.Order (OrderID == 0), it means there is no open orders at this pin.
-func (s *Strategy) buildPinOrderMap(grid *Grid, openOrders []types.Order) (map[string]types.Order, error) {
-	pinOrderMap := make(map[string]types.Order)
+func (s *Strategy) buildPinOrderMap(pins []Pin, openOrders []types.Order) (PinOrderMap, error) {
+	pinOrderMap := make(PinOrderMap)
 
-	for _, pin := range grid.Pins {
+	for _, pin := range pins {
 		priceStr := s.FormatPrice(fixedpoint.Value(pin))
```
c9s, line 1469: can you move FormatPrice func to types.Market? so that you don't need to depend on the strategy object.

pkg/strategy/grid2/pin_order_map.go
```diff
@@ -0,0 +1,34 @@
...
+
+		orders = append(orders, order)
+	}
+
+	types.SortOrdersUpdateTimeAscending(orders)
+
+	return orders
+}
+
+func (m PinOrderMap) buildSyncOrderMap() *types.SyncOrderMap {
```
c9s, line 27: You can use `SyncOrderMap` as the method name?

gx578007 (approved):

PR #1098, opened 2023-03-10
FIX: fix format string float point issue

kbearXD:


c9s (dismissed):

c9s:
left for bohan to review this XD

pkg/fixedpoint/convert.go
```diff
@@ -115,9 +115,50 @@ func (v Value) FormatString(prec int) string {
...
-	pow := math.Pow10(prec)
-	return strconv.FormatFloat(
-		math.Trunc(float64(v)/DefaultPow*pow)/pow, 'f', prec, 64)
+
+	u := int64(v)
+
+	// trunc precision
+	precDiff := DefaultPrecision - prec
+	if precDiff > 0 {
+		powDiff := int64(math.Pow10(precDiff))
```
gx578007, line 124: Better to use `math.Round` to convert float64 to int64.

pkg/fixedpoint/convert.go
```diff
@@ -115,9 +115,50 @@ func (v Value) FormatString(prec int) string {
...
 	}
-	pow := math.Pow10(prec)
-	return strconv.FormatFloat(
-		math.Trunc(float64(v)/DefaultPow*pow)/pow, 'f', prec, 64)
+
+	u := int64(v)
+
+	// trunc precision
+	precDiff := DefaultPrecision - prec
+	if precDiff > 0 {
```
c9s, line 123: what about negative precDiff?

kbearXD: precDiff > 0 means the prec is larger than 8 and our precision is only 8 for fixedpoint.Value
As a result, we don't need to trunc it.

kbearXD: if prec = 5, precDiff = 8 - 5 = 3.
For 123456789 (1.23456789) => 123456000 (1.23456)

pkg/fixedpoint/convert_test.go
```diff
@@ -0,0 +1,59 @@
...
+		assert.Equal("-0.57000", s)
+	})
+
+	t.Run("Value(-123456789) with prec = 9, expected 1.23456789", func(t *testing.T) {
+		v := MustNewFromString("-1.23456789")
+		s := v.FormatString(9)
+		assert.Equal("-1.234567890", s)
+	})
+
+	t.Run("Value(1234567890) with prec = -1, expected 10", func(t *testing.T) {
```
gx578007, line 48: Could add another test with prec=-3 which expects "0"?

kbearXD: OK!

c9s:
check failed

PR #1099, opened 2023-03-10
FIX: [grid2] avoid handling one orderID twice

gx578007:


pkg/strategy/grid2/strategy.go
```diff
@@ -1766,6 +1775,8 @@ func (s *Strategy) getWriteContext(fallbackCtxList ...context.Context) context.C
 func (s *Strategy) Run(ctx context.Context, _ bbgo.OrderExecutor, session *bbgo.ExchangeSession) error {
 	instanceID := s.InstanceID()
 
+	s.filledOrderIDSet = map[uint64]struct{}{}
```
c9s, line 1778: you can move this initialization to the Initialize() method

c9s, line 1778: maybe use types.OrderMap? or types.SyncOrderMap for thread safefy?

gx578007: Okay, I'll take a look.

c9s (approved):

PR #1104, opened 2023-03-13
fix: rebalance: adjust max amount by balance

narumiruna:


pkg/strategy/rebalance/strategy.go
```diff
@@ -169,12 +170,16 @@ func (s *Strategy) prices(ctx context.Context) types.ValueMap {
 	return m
 }
 
+func (s *Strategy) balance(currency string) types.Balance {
+	balance, _ := s.session.GetAccount().Balance(currency)
+	return balance
```
c9s, line 175: it's possible to get a Zeroed types.Balance here, are you okay with that?

c9s (approved):

PR #1108, opened 2023-03-14
exchange:binance: add QueryFuturesDepth

austin362667:
exchange:binance: add QueryFuturesDepth

### DIFF
--- a/pkg/exchange/binance/exchange.go
+++ b/pkg/exchange/binance/exchange.go
@@ -1622,6 +1622,60 @@ func (e *Exchange) QueryDepth(ctx context.Context, symbol string) (snapshot type
 	return snapshot, finalUpdateID, nil
 }
 
+// QueryDepth query the order book depth of a symbol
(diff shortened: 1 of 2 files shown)


pkg/exchange/binance/exchange.go
```diff
@@ -1622,6 +1622,60 @@ func (e *Exchange) QueryDepth(ctx context.Context, symbol string) (snapshot type
 	return snapshot, finalUpdateID, nil
 }
 
+// QueryDepth query the order book depth of a symbol
+func (e *Exchange) QueryFuturesDepth(ctx context.Context, symbol string) (snapshot types.SliceOrderBook, finalUpdateID int64, err error) {
+	var response *futures.DepthResponse
+	if e.IsFutures {
```
c9s, line 1628: since it's QueryFuturesDepth, so you don't have to check the switch?

austin362667:
After carefully check, depth response is the same between futures & spot.

PR #1109, opened 2023-03-14
fix: rebalance: fix positions and profit stats map

narumiruna:


c9s (approved):

pkg/strategy/rebalance/position_map.go
```diff
@@ -6,15 +6,17 @@ import (
 
 type PositionMap map[string]*types.Position
 
-func NewPositionMap(markets []types.Market) PositionMap {
-	m := make(PositionMap)
-
+func (m PositionMap) createPositions(markets []types.Market) PositionMap {
```
c9s, line 9: why not NewPositionMap ?

narumiruna: If we change the config from 
```yaml
quoteCurrency: USDT
targetWeights:
        BTC: 50%
        USDT: 50%
```
to
```yaml
quoteCurrency: USDT
targetWeights:
        BTC: 50%
        ETH: 25%
        USDT: 25%
```
, then the previous `PositionMap` (loaded from Redis) doesn't contain ETHUSDT.

PR #1122, opened 2023-03-16
strategy: fixedmaker: clamp skew

narumiruna:


pkg/fixedpoint/convert.go
```diff
@@ -590,3 +590,23 @@ func Abs(a Value) Value {
...
+	if x < min {
+		return min
+	}
+	if x > max {
+		return max
+	}
+	return x
+}
+
+func (x Value) Clamp(min, max Value) Value {
```
c9s, line 604: Please run with -tag dnum mode to verify this.

we have two implementations

narumiruna: ok

c9s (approved):

PR #1147, opened 2023-04-14
FEATURE: max get-order v3 api support client order id parameter

kbearXD:


pkg/exchange/max/exchange.go
```diff
@@ -210,16 +210,30 @@ func (e *Exchange) QueryOrderTrades(ctx context.Context, q types.OrderQuery) ([]
...
+	if len(q.OrderID) != 0 {
+		orderID, err := strconv.ParseInt(q.OrderID, 10, 64)
+		if err != nil {
+			return nil, err
+		}
+
+		request.Id(uint64(orderID))
+	}
+
+	if len(q.ClientOrderID) != 0 {
```
c9s, line 232: you need `else if` here?

kbearXD: No, because we already exclude the situation 'q.OrderID == 0 && q.ClientOrderID == 0'.

PR #1163, opened 2023-04-27
FEATURE: use TwinOrder to recover

kbearXD:


pkg/strategy/grid2/twin_order.go
```diff
@@ -0,0 +1,113 @@
...
+
+		return fixedpoint.Value(pin), nil
+	}
+
+	return fixedpoint.Zero, fmt.Errorf("unsupported side: %s of this order (%d)", order.Side, order.OrderID)
+}
+
+func (m TwinOrderMap) AscendingOrders() []types.Order {
+	var orders []types.Order
+	for _, order := range m {
```
gx578007, line 71: How about renaming `order` to `twinOrder`?

pkg/strategy/grid2/recover.go
```diff
@@ -319,3 +310,173 @@ func addOrdersIntoPinOrderMap(pinOrders PinOrderMap, orders []types.Order) error
...
+	s.debugOrders("verifying filled grid - filled orders", filledOrders)
+
+	if err := s.addOrdersIntoTwinOrderMap(twinOrders, filledOrders); err != nil {
+		return errors.Wrapf(err, "verifying filled grid error when add orders into twin order map")
+	}
+
+	s.debugLog("verifying filled grid - filled twin orders:\n%+v", twinOrders.String())
+
+	for i, pin := range pins {
+		if i == 0 {
```
gx578007, line 326: Could you leave a comment here describing it's because sell price is used as the key of twin order map?

pkg/strategy/grid2/recover.go
```diff
@@ -319,3 +310,173 @@ func addOrdersIntoPinOrderMap(pinOrders PinOrderMap, orders []types.Order) error
...
+			continue
+		}
+
+		twin, exist := twinOrders[fixedpoint.Value(pin)]
+		if !exist {
+			return fmt.Errorf("there is no order at price (%+v)", pin)
+		}
+
```
gx578007, line 334: Is it required to call `twin.IsValid()` here?

gx578007 (approved):

pkg/strategy/grid2/recover.go
kbearXD: We can add more checks here.

c9s (changes requested):
let's hold on this for a while, thanks!

c9s (approved):

PR #1165, opened 2023-04-28
FIX: supertrend config type error

RicoToothless:
`interval` only accepts string

The error from backtest

```
[0000] FATAL cannot execute command error=json: cannot unmarshal number into Go struct field HigherHighLowerLowStop.exits.higherHighLowerLowStopLoss.interval of type string
```


RicoToothless:
polygon:0xD95A6459ed4603Ca235d7830D765Dd571dDE6e63

c9s (approved):

PR #1176, opened 2023-05-23
FIX: [grid2] fix base + quote calculation and add baseGridNumber config field

c9s:


pkg/strategy/grid2/strategy.go
```diff
@@ -761,25 +764,29 @@ func (s *Strategy) calculateBaseQuoteInvestmentQuantity(quoteInvestment, baseInv
...
+			price := fixedpoint.Value(pin)
+			sellPrice := price
+			if s.ProfitSpread.Sign() > 0 {
+				sellPrice = sellPrice.Add(s.ProfitSpread)
+			}
 
-		if sellPrice.Compare(lastPrice) < 0 {
-			break
-		}
+			if sellPrice.Compare(lastPrice) < 0 {
```
gx578007, line 779: Might be worth adding testcase for `sellPrice == lastPrice`.

c9s: I can add it in the next PR

gx578007 (approved):

PR #1177, opened 2023-05-24
fix: some types and operations in SeriesExtended are not supported

zenixls2:
Add int32/int64/int support on:
- Dot
- Add
- Mul


c9s:
Want to use generic? :-P

c9s (approved):

PR #1233, opened 2023-07-18
FIX: [grid2] fix upper pin

gx578007:


c9s (approved):

pkg/strategy/grid2/grid.go
```diff
@@ -52,7 +52,12 @@ func calculateArithmeticPins(lower, upper, spread, tickSize fixedpoint.Value) []
 	}
 
 	// this makes sure there is no error at the upper price
-	pins = append(pins, Pin(upper))
+	upp := math.Round(upper.Float64()*pow10*10.0) / 10.0
```
c9s, line 55: can be `pp`? because `pp` is the scope varaible in the for loop?

kbearXD (approved):

c9s (approved):

PR #1238, opened 2023-07-21
TEST: add unit test for okex exchange

MengShue:
add some unit test for OKEX exchange

### DIFF
--- a/pkg/exchange/okex/convert.go
+++ b/pkg/exchange/okex/convert.go
@@ -165,14 +165,10 @@ func toGlobalTrades(orderDetails []okexapi.OrderDetails) ([]types.Trade, error) 
 func toGlobalOrders(orderDetails []okexapi.OrderDetails) ([]types.Order, error) {
 	var orders []types.Order
 	for _, orderDetail := range orderDetails {
-		isMargin := false
-		if orderDetail.InstrumentType == string(okexapi.InstrumentTypeMARGIN) {
-			isMargin = true
-		}
 
-		o, err := toGlobalOrder(&orderDetail, isMargin)
+		o, err := toGlobalOrder(&orderDetail)
 		if err != nil {
-			log.WithError(err).Error("order convert error")
+			return nil, err
 		} else {
 			orders = append(orders, *o)
 		}
@@ -218,10 +214,10 @@ func toGlobalOrderType(orderType okexapi.OrderType) (types.OrderType, error) {
 	case okexapi.OrderTypeMarket:
 		return types.OrderTypeMarket, nil
 
-	case okexapi.OrderTypeLimit, okexapi.OrderTypeMarketMakerProtection:
+	case okexapi.OrderTypeLimit, okexapi.OrderTypeFOK, okexapi.OrderTypeIOC, okexapi.OrderTypeMarketMakerProtection:
 		return types.OrderTypeLimit, nil
 
-	case okexapi.OrderTypePostOnly, okexapi.OrderTypeIOC, okexapi.OrderTypeFOK, okexapi.OrderTypeMarektMakerProtectionPostOnly:
+	case okexapi.OrderTypePostOnly, okexapi.OrderTypeMarketMakerProtectionPostOnly:
 		return types.OrderTypeLimitMaker, nil
 
 	}
@@ -236,18 +232,28 @@ func toLocalInterval(src string) string {
 	})
 }
 
-func toGlobalOrder(okexOrder *okexapi.OrderDetails, isMargin bool) (*types.Order, error) {
+func toGlobalSide(side okexapi.SideType) (s types.SideType) {
+	switch string(side) {
+	case "sell":
+		s = types.SideTypeSell
+	case "buy":
+		s = types.SideTypeBuy
+	}
+	return s
+}
+
+func toGlobalOrder(okexOrder *okexapi.OrderDetails) (*types.Order, error) {
 
 	orderID, err := strconv.ParseInt(okexOrder.OrderID, 10, 64)
 	if err != nil {
-		return &types.Order{}, err
+		return nil, err
 	}
 
-	side := types.SideType(strings.ToUpper(string(okexOrder.Side)))
+	side := toGlobalSide(okexOrder.Side)
 
 	orderType, err := toGlobalOrderType(okexOrder.OrderType)
 	if err != nil {
-		return &types.Order{}, err
+		return nil, err
 	}
 
 	timeInForce := types.TimeInForceGTC
@@ -260,7 +266,7 @@ func toGlobalOrder(okexOrder *okexapi.OrderDetails, isMargin bool) (*types.Order
 
 	orderStatus, err := toGlobalOrderStatus(okexOrder.State)
 	if err != nil {
-		return &types.Order{}, err
+		return nil, err
 	}
 
 	isWorking := false
@@ -268,6 +274,11 @@ func toGlobalOrder(okexOrder *okexapi.OrderDetails, isMargin bool) (*types.Order
 	case types.OrderStatusNew, types.OrderStatusPartiallyFilled:
 		isWorking = true
 
+	}
+
+	isMargin := false
+	if okexOrder.InstrumentType == string(okexapi.InstrumentTypeMARGIN) {
+		isMargin = true
 	}
 
 	return &types.Order{
--- a/pkg/exchange/okex/convert.go
+++ b/pkg/exchange/okex/convert.go
@@ -168,10 +168,9 @@ func toGlobalOrders(orderDetails []okexapi.OrderDetails) ([]types.Order, error) 
 
 		o, err := toGlobalOrder(&orderDetail)
 		if err != nil {
-			return nil, err
-		} else {
-			orders = append(orders, *o)
+			log.WithError(err).Error("order convert error")
 		}
+		orders = append(orders, *o)
 	}
 
 	return orders, nil
@@ -210,6 +209,8 @@ func toLocalOrderType(orderType types.OrderType) (okexapi.OrderType, error) {
 }
 
 func toGlobalOrderType(orderType okexapi.OrderType) (types.OrderType, error) {
+	// Okex IOC and FOK only implement limit order
+	// reference: https://www.okx.com/cn/help-center/360025135731
 	switch orderType {
 	case okexapi.OrderTypeMarket:
 		return types.OrderTypeMarket, nil
--- a/pkg/exchange/okex/convert.go
+++ b/pkg/exchange/okex/convert.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	"go.uber.org/multierr"
 
 	"github.com/c9s/bbgo/pkg/exchange/okex/okexapi"
 	"github.com/c9s/bbgo/pkg/fixedpoint"
@@ -164,16 +165,17 @@ func toGlobalTrades(orderDetails []okexapi.OrderDetails) ([]types.Trade, error) 
 
 func toGlobalOrders(orderDetails []okexapi.OrderDetails) ([]types.Order, error) {
 	var orders []types.Order
+	var err error
 	for _, orderDetail := range orderDetails {
 
-		o, err := toGlobalOrder(&orderDetail)
-		if err != nil {
-			log.WithError(err).Error("order convert error")
+		o, err2 := toGlobalOrder(&orderDetail)
+		if err2 != nil {
+			err = multierr.Append(err, err2)
 		}
 		orders = append(orders, *o)
 	}
 
-	return orders, nil
+	return orders, err
 }
 
 func toGlobalOrderStatus(state okexapi.OrderState) (types.OrderStatus, error) {
@@ -209,8 +211,7 @@ func toLocalOrderType(orderType types.OrderType) (okexapi.OrderType, error) {
 }
 
 func toGlobalOrderType(orderType okexapi.OrderType) (types.OrderType, error) {
-	// Okex IOC and FOK only implement limit order
-	// reference: https://www.okx.com/cn/help-center/360025135731
+	// IOC, FOK are only allowed with limit order type, so we assume the order type is always limit order for FOK, IOC orders
 	switch orderType {
 	case okexapi.OrderTypeMarket:
 		return types.OrderTypeMarket, nil
--- a/pkg/exchange/okex/convert.go
+++ b/pkg/exchange/okex/convert.go
@@ -2,7 +2,6 @@ package okex
 
 import (
 	"fmt"
-	"hash/fnv"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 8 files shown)


pkg/exchange/okex/okexapi/client_test.go
```diff
@@ -0,0 +1,76 @@
...
+	client := NewClient()
+	client.Auth(key, secret, passphrase)
+	return client
+}
+
+func TestClient_GetInstrumentsRequest(t *testing.T) {
+	client := NewClient()
+	ctx := context.Background()
+
+	ser := PublicDataService{client: client}
```
c9s, line 34: ser -> srv (service abbr)

pkg/exchange/okex/okexapi/client_test.go
```diff
@@ -0,0 +1,76 @@
...
+		Do(ctx)
+	assert.NoError(t, err)
+	assert.NotEmpty(t, instruments)
+	t.Logf("instruments: %+v", instruments)
+}
+
+func TestClient_GetFundingRateRequest(t *testing.T) {
+	client := NewClient()
+	ctx := context.Background()
+	ser := PublicDataService{client: client}
```
c9s, line 48: would be better to use &PublicDataService{} to create it as a pointer object

pkg/exchange/okex/convert.go
```diff
@@ -255,20 +251,27 @@ func toLocalOrderType(orderType types.OrderType) (okexapi.OrderType, error) {
...
+		return types.OrderTypeLimit
+
 	case okexapi.OrderTypePostOnly:
-		return types.OrderTypeLimitMaker, nil
+		return types.OrderTypeLimitMaker
 
-	case okexapi.OrderTypeFOK:
-	case okexapi.OrderTypeIOC:
+	case okexapi.OrderTypeIOC, okexapi.OrderTypeFOK:
+		return types.OrderTypeMarket
```
gx578007, line 266: It's a bit weird to make ioc/fok orders as market orders. @c9s, what do you think to make these two types `types.OrderTypeLimitMaker`?

pkg/exchange/okex/convert.go
```diff
@@ -223,20 +219,20 @@ func toGlobalOrders(orderDetails []okexapi.OrderDetails) ([]types.Order, error)
 	return orders, nil
 }
 
-func toGlobalOrderStatus(state okexapi.OrderState) (types.OrderStatus, error) {
+func toGlobalOrderStatus(state okexapi.OrderState) types.OrderStatus {
```
c9s, line 222: is there a reason to remove the error?

pkg/exchange/okex/convert.go
c9s: yeah, IOC or FOK doesn't mean Market order type, you might need to check some other property.

pkg/exchange/okex/convert.go
```diff
@@ -277,3 +280,34 @@ func toLocalInterval(src string) string {
...
+func toGlobalOrder(okexOrder *okexapi.OrderDetails, isMargin bool) (*types.Order, error) {
+	timeInForce := types.TimeInForceFOK
+	if okexOrder.OrderType == okexapi.OrderTypeIOC {
+		timeInForce = types.TimeInForceIOC
+	}
+	return &types.Order{
+		SubmitOrder: types.SubmitOrder{
+			ClientOrderID: okexOrder.ClientOrderID,
+			Symbol:        okexOrder.InstrumentID,
+			Side:          types.SideType(okexOrder.Side),
```
c9s, line 299: okexOrder.Side vlaues are BUY, SELL in upper case?

pkg/exchange/okex/convert.go
```diff
@@ -277,3 +280,34 @@ func toLocalInterval(src string) string {
...
+			Type:          toGlobalOrderType(okexOrder.OrderType),
+			Quantity:      okexOrder.Quantity,
+			Price:         okexOrder.Price,
+			TimeInForce:   types.TimeInForce(timeInForce),
+		},
+		Exchange:         types.ExchangeOKEx,
+		OrderID:          hashStringID(okexOrder.OrderID),
```
c9s, line 306: how do we convert the order ID in other cases?

pkg/exchange/okex/convert.go
MengShue: No, No need to remove the error. Sorry, I used other exchange as reference, right now I use toGlobalOrders() as reference.

pkg/exchange/okex/convert.go
MengShue: ok, IOC/FOK I change to `types.OrderTypeLimitMaker`

pkg/exchange/okex/convert.go
```diff
@@ -255,20 +251,27 @@ func toLocalOrderType(orderType types.OrderType) (okexapi.OrderType, error) {
 	return "", fmt.Errorf("unknown or unsupported okex order type: %s", orderType)
 }
 
-func toGlobalOrderType(orderType okexapi.OrderType) (types.OrderType, error) {
+func toGlobalOrderType(orderType okexapi.OrderType) types.OrderType {
 	switch orderType {
-	case okexapi.OrderTypeMarket:
-		return types.OrderTypeMarket, nil
+	case okexapi.OrderTypeMarket, okexapi.OrderTypeMarketMakerProtection:
```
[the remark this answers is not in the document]
MengShue: `okexapi.OrderTypeMarketMakerProtection` ->  `types.OrderTypeLimit`
`okexapi.OrderTypeMarektMakerProtectionPostOnly` -> `types.OrderTypeLimitMaker`
Is that right ?
But I still feel some confuse, so Market Markets usually don't use market orders, just use limit orders?

pkg/exchange/okex/convert.go
MengShue: lower case, sorry.  I used other exchange as reference, right now I use toGlobalOrders() as reference.

pkg/exchange/okex/convert.go
MengShue: I used kucoin exchange as reference, <- wrong, 
 Right now I use toGlobalOrders() as reference.

pkg/exchange/okex/convert.go
```diff
@@ -164,60 +165,17 @@ func toGlobalTrades(orderDetails []okexapi.OrderDetails) ([]types.Trade, error)
...
+		isMargin := false
+		if orderDetail.InstrumentType == string(okexapi.InstrumentTypeMARGIN) {
+			isMargin = true
 		}
 
-		orderStatus, err := toGlobalOrderStatus(orderDetail.State)
+		o, err := toGlobalOrder(&orderDetail, isMargin)
 		if err != nil {
-			return orders, err
+			log.WithError(err).Error("order convert error")
```
c9s, line 175: when convert error, you need to "continue"?

pkg/exchange/okex/convert.go
```diff
@@ -259,15 +217,15 @@ func toGlobalOrderType(orderType okexapi.OrderType) (types.OrderType, error) {
...
+
+	case okexapi.OrderTypeLimit, okexapi.OrderTypeMarketMakerProtection:
 		return types.OrderTypeLimit, nil
-	case okexapi.OrderTypePostOnly:
-		return types.OrderTypeLimitMaker, nil
 
-	case okexapi.OrderTypeFOK:
-	case okexapi.OrderTypeIOC:
+	case okexapi.OrderTypePostOnly, okexapi.OrderTypeIOC, okexapi.OrderTypeFOK, okexapi.OrderTypeMarektMakerProtectionPostOnly:
+		return types.OrderTypeLimitMaker, nil
```
c9s, line 225: IOC or FOK is not an order type, it's a "TimeInForce" property, that means:

IOC and FOK could be used with Limit maker, limit taker, market taker order together

c9s: You need to convert FOK and IOC into "TimeInForce"

pkg/exchange/okex/convert.go
```diff
@@ -277,3 +235,60 @@ func toLocalInterval(src string) string {
...
 }
+
+func toGlobalOrder(okexOrder *okexapi.OrderDetails, isMargin bool) (*types.Order, error) {
+
+	orderID, err := strconv.ParseInt(okexOrder.OrderID, 10, 64)
+	if err != nil {
+		return &types.Order{}, err
+	}
+
+	side := types.SideType(strings.ToUpper(string(okexOrder.Side)))
```
c9s, line 246: can we use toGlobalSide here?

pkg/exchange/okex/convert.go
c9s: does okex's doc explain the IOC, FOK?

pkg/exchange/okex/convert.go
```diff
@@ -277,3 +235,60 @@ func toLocalInterval(src string) string {
 		return strings.ToUpper(w)
 	})
 }
+
+func toGlobalOrder(okexOrder *okexapi.OrderDetails, isMargin bool) (*types.Order, error) {
+
+	orderID, err := strconv.ParseInt(okexOrder.OrderID, 10, 64)
+	if err != nil {
+		return &types.Order{}, err
```
c9s, line 243: should be `return nil, err`

pkg/exchange/okex/convert.go
```diff
@@ -277,3 +235,60 @@ func toLocalInterval(src string) string {
...
+	}
+
+	side := types.SideType(strings.ToUpper(string(okexOrder.Side)))
+
+	orderType, err := toGlobalOrderType(okexOrder.OrderType)
+	if err != nil {
+		return &types.Order{}, err
```
c9s, line 250: should be `return nil, err`

pkg/exchange/okex/convert.go
```diff
@@ -277,3 +235,60 @@ func toLocalInterval(src string) string {
...
+	switch okexOrder.OrderType {
+	case okexapi.OrderTypeFOK:
+		timeInForce = types.TimeInForceFOK
+	case okexapi.OrderTypeIOC:
+		timeInForce = types.TimeInForceIOC
+	}
+
+	orderStatus, err := toGlobalOrderStatus(okexOrder.State)
+	if err != nil {
+		return &types.Order{}, err
```
c9s, line 263: should be `return nil, err`

pkg/exchange/okex/convert.go
```diff
@@ -277,3 +235,60 @@ func toLocalInterval(src string) string {
...
+	}
+
+	isWorking := false
+	switch orderStatus {
+	case types.OrderStatusNew, types.OrderStatusPartiallyFilled:
+		isWorking = true
```
c9s, line 269: it's good, I think this makes sense

pkg/exchange/okex/exchange.go
```diff
@@ -339,3 +339,30 @@ func (e *Exchange) QueryKLines(ctx context.Context, symbol string, interval type
 	return klines, nil
 
 }
+
+func (e *Exchange) QueryOrder(ctx context.Context, q types.OrderQuery) (*types.Order, error) {
+	if len(q.Symbol) == 0 {
+		return nil, errors.New("okex.QueryOrder: InstrumentID is required parameter")
```
c9s, line 345: fixed error object could be pulled out to:

```
var ErrXXX = errors.New("...")
```

pkg/exchange/okex/exchange.go
```diff
@@ -339,3 +339,30 @@ func (e *Exchange) QueryKLines(ctx context.Context, symbol string, interval type
 	return klines, nil
 
 }
+
+func (e *Exchange) QueryOrder(ctx context.Context, q types.OrderQuery) (*types.Order, error) {
+	if len(q.Symbol) == 0 {
```
ycdesu, line 344: The purpose of `Exchange` is encapsulating the underlying details of okex. User will not know what `InstrumentID` is. User only knows what a `Symbol` is, so it would be better not to use `InstrumentID` in the error message.

pkg/exchange/okex/exchange.go
```diff
@@ -339,3 +339,30 @@ func (e *Exchange) QueryKLines(ctx context.Context, symbol string, interval type
...
+		return nil, errors.New("okex.QueryOrder: InstrumentID is required parameter")
+	}
+	if len(q.OrderID) == 0 && len(q.ClientOrderID) == 0 {
+		return nil, errors.New("okex.QueryOrder: ordId or clOrdId is required parameter")
```
ycdesu, line 348: The error message is for human, and `clOrdId` is not cleared enough. Using `ClientOrderID` directly is clearer.

pkg/exchange/okex/exchange.go
```diff
@@ -339,3 +339,30 @@ func (e *Exchange) QueryKLines(ctx context.Context, symbol string, interval type
...
+	order, err := req.Do(ctx)
+
+	if err != nil {
+		return nil, err
+	}
+
+	isMargin := false
+	if order.InstrumentType == string(okexapi.InstrumentTypeMARGIN) {
+		isMargin = true
+	}
```
ycdesu, line 365: The conversion does not depend on any variables within `Exchange` struct. Is there any reason not to check the `order.InstrumentType` in the `toGlobalOrder` function?

pkg/exchange/okex/okexapi/client_test.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+func getTestClientOrSkip(t *testing.T) *RestClient {
+	if b, _ := strconv.ParseBool(os.Getenv("CI")); b {
+		t.Skip("skip test for CI")
+	}
+
+	key, secret, passphrase, ok := testutil.IntegrationTestWithPassphraseConfigured(t, "OKEX")
+	if !ok {
+		t.SkipNow()
```
ycdesu, line 21: When this case really happens, how do we know from the CI log? I mean, is there any way to remind us to configure the credential?

pkg/exchange/okex/exchange.go
```diff
@@ -339,3 +339,30 @@ func (e *Exchange) QueryKLines(ctx context.Context, symbol string, interval type
...
+
+	return toGlobalOrder(order, isMargin)
```
c9s, line 367: since the order has the InstrumentType field, you don't have to pass it as an argument?

MengShue: Updated. isMargin is no longer an argument

pkg/exchange/okex/okexapi/client_test.go
MengShue: Should print log to indicate credentials need to be added

pkg/exchange/okex/exchange.go
MengShue: Updated. I check `order.InstrumentType` in the `toGlobalOrder` function now

pkg/exchange/okex/exchange.go
MengShue: Updated.

pkg/exchange/okex/exchange.go
MengShue: Updated.

pkg/exchange/okex/exchange.go
MengShue: Updated, I declare a error type variable in exchange.go

pkg/exchange/okex/convert.go
MengShue: Updated

pkg/exchange/okex/convert.go
MengShue: Updated

pkg/exchange/okex/convert.go
MengShue: Updated

pkg/exchange/okex/convert.go
MengShue: Updated

pkg/exchange/okex/convert.go
MengShue: The same as MAX, only limit order allowed in IOC and FOK, so I update IOC and FOK to `types.OrderTypeLimit` 
https://www.okx.com/cn/help-center/360025135731

pkg/exchange/okex/convert.go
MengShue: I return `nil ,err` now. But in MAX convert.go, it still continue while encounter error. So I should fix it too ?

pkg/exchange/okex/convert.go
```diff
@@ -164,60 +165,13 @@ func toGlobalTrades(orderDetails []okexapi.OrderDetails) ([]types.Trade, error)
...
-			timeInForce = types.TimeInForceIOC
 
-		}
-
-		orderStatus, err := toGlobalOrderStatus(orderDetail.State)
+		o, err := toGlobalOrder(&orderDetail)
 		if err != nil {
-			return orders, err
+			return nil, err
+		} else {
```
c9s, line 172: this "else" could be removed

c9s: keep:

```
orders = append(orders, *o)
```

pkg/exchange/okex/convert.go
c9s: Got it, thanks. Could you add the doc explaination into the comment?

pkg/exchange/okex/exchange.go
```diff
@@ -26,6 +26,8 @@ var log = logrus.WithFields(logrus.Fields{
 	"exchange": ID,
 })
 
+var ErrSymbolRequired = errors.New("Symbol is required parameter")
```
c9s, line 29: use `symbol is a required parameter`, error message starts with lower case

gx578007 (approved):

pkg/exchange/okex/convert.go
c9s: you can:

- wrap error with the multierr package and continue to the next order
- or, return the error directly

but we should avoid logging error in the under layer method

pkg/exchange/okex/convert.go
```diff
@@ -256,18 +209,20 @@ func toLocalOrderType(orderType types.OrderType) (okexapi.OrderType, error) {
 }
 
 func toGlobalOrderType(orderType okexapi.OrderType) (types.OrderType, error) {
+	// Okex IOC and FOK only implement limit order
```
c9s, line 212: "IOC, FOK are only allowed with limit order type, so we assume the order type is always limit order for FOK, IOC orders"

MengShue:
polygon:0x75b149B14A5a7CE223198dD35f17835DD4e07fbC

MengShue:
`polygon:0x75b149B14A5a7CE223198dD35f17835DD4e07fbC`

PR #1239, opened 2023-07-22
FIX: [grid2] add remove duplicated pins and pull out filter price prec func

c9s:


pkg/strategy/grid2/grid.go
```diff
@@ -35,28 +35,48 @@ type Grid struct {
...
+
+func removeDuplicatedPins(pins []Pin) []Pin {
+	var buckets = map[string]struct{}{}
+	var out []Pin
+
+	for _, pin := range pins {
+		p := fixedpoint.Value(pin)
+
+		if _, exists := buckets[p.String()]; exists {
+			continue
```
gx578007, line 57: It might be worth error logging if any duplicated pin is found.

gx578007 (approved):

PR #1243, opened 2023-07-24
FEATURE: pkg/exchange: add query market to bybit exchange

bailantaotao:


pkg/exchange/bybit/convert.go
```diff
@@ -0,0 +1,53 @@
...
+	//    MinOrderQty: 0.000048
+	//    MaxOrderQty: 71.73956243
+	//    MinOrderAmt: 1
+	//    MaxOrderAmt: 2000000
+	//}
+	//
+	//PriceFilter:
+	//{
+	//    TickSize: 0.01
+	//}
```
gx578007, line 32: Could you add a unit test for this converting function to make sure parsing is performed as expected?

c9s (approved):

pkg/exchange/bybit/convert.go
```diff
@@ -0,0 +1,53 @@
...
+	return types.Market{
+		Symbol:          m.Symbol,
+		LocalSymbol:     m.Symbol,
+		PricePrecision:  int(math.Log10(m.LotSizeFilter.QuotePrecision.Float64())),
+		VolumePrecision: int(math.Log10(m.LotSizeFilter.BasePrecision.Float64())),
+		QuoteCurrency:   m.QuoteCoin,
+		BaseCurrency:    m.BaseCoin,
+		MinNotional:     m.LotSizeFilter.MinOrderAmt,
+		MinAmount:       m.LotSizeFilter.MinOrderAmt,
```
ycdesu, line 41: These two fields have the same value. Does that expect?

ycdesu: Please also add unit tests for this convertor.

pkg/exchange/bybit/convert.go
bailantaotao: sure!

pkg/exchange/bybit/convert.go
bailantaotao: This has been confirmed with @c9s . Some exchanges have a `MIN_NOTIONAL` and declare it in the API response. But in BYBIT there is only a minimum amount limit, so we fill him. This value is used to confirm the minimum order amount when trading

c9s: Yes

pkg/exchange/bybit/convert.go
c9s: Yeah that would be better

gx578007 (approved):

PR #1248, opened 2023-07-25
pkg/exchange: add QueryOpenOrders API for bybit

bailantaotao:


pkg/exchange/bybit/bybitapi/client_test.go
```diff
@@ -57,4 +57,24 @@ func TestClient(t *testing.T) {
...
+			}
+			openOrders, err := req.Do(ctx)
+			assert.NoError(t, err)
+
+			for _, o := range openOrders.List {
+				t.Logf("openOrders: %+v", o)
+			}
+			if len(openOrders.NextPageCursor) == 0 {
+				break
+			}
```
ycdesu, line 76: Looks like we only check the err is nil or not. Could we add other cases, such as verify the `len(openOrders.List)`?

pkg/exchange/bybit/convert.go
```diff
@@ -43,3 +45,123 @@ func toGlobalTicker(stats bybitapi.Ticker, time time.Time) types.Ticker {
...
+
+func hashStringID(s string) uint64 {
+	h := fnv.New64a()
+	h.Write([]byte(s))
+	return h.Sum64()
+}
+
+func isWorking(status bybitapi.OrderStatus) bool {
+	s, _ := toGlobalOrderStatus(status)
+	return s == types.OrderStatusNew || s == types.OrderStatusPartiallyFilled
```
ycdesu, line 166: If you swallow the error, we will never know we receive an unexpected error. You can add a warning log here.

ycdesu (approved):
I have left some minor comments.

pkg/exchange/bybit/convert.go
bailantaotao: Originally, I thought there would only be true or false, so I didn't check for errors. However, I agree with what you said, and I will add checks.

pkg/exchange/bybit/bybitapi/client_test.go
c9s: it's an integration test, so the length of the openOrders.List may vary

pkg/exchange/bybit/bybitapi/get_open_order_request.go
```diff
@@ -0,0 +1,159 @@
...
+import (
+	"github.com/c9s/bbgo/pkg/fixedpoint"
+	"github.com/c9s/bbgo/pkg/types"
+	"github.com/c9s/requestgen"
+)
+
+//go:generate -command GetRequest requestgen -method GET -responseType .APIResponse -responseDataField Result
+//go:generate -command PostRequest requestgen -method POST -responseType .APIResponse -responseDataField Result
+
+type OpenOrders struct {
```
c9s, line 12: Could we rename to OpenOrdersResponse to avoid confusion?

pkg/exchange/bybit/bybitapi/get_open_order_request.go
```diff
@@ -0,0 +1,159 @@
...
+const (
+	OpenOnlyOrder OpenOnly = 0
+)
+
+type Side string
+
+const (
+	SideBuy  Side = "Buy"
+	SideSell Side = "Sell"
+)
```
c9s, line 89: move these to types.go?

pkg/exchange/bybit/bybitapi/get_open_order_request.go
```diff
@@ -0,0 +1,159 @@
...
+	TpslMode           string                     `json:"tpslMode"`
+	TpLimitPrice       string                     `json:"tpLimitPrice"`
+	SlLimitPrice       string                     `json:"slLimitPrice"`
+	PlaceType          string                     `json:"placeType"`
+	CreatedTime        types.MillisecondTimestamp `json:"createdTime"`
+	UpdatedTime        types.MillisecondTimestamp `json:"updatedTime"`
+}
+
+//go:generate GetRequest -url "/v5/order/realtime" -type GetOpenOrderRequest -responseDataType .OpenOrders
+type GetOpenOrderRequest struct {
```
c9s, line 63: since it returns more than one open order, call it `GetOpenOrdersRequest`?

pkg/exchange/bybit/bybitapi/get_open_order_request.go
```diff
@@ -0,0 +1,159 @@
...
+		OrderStatusPartiallyFilled,
+		OrderStatusPartiallyFilledCanceled,
+		OrderStatusFilled,
+		OrderStatusCancelled,
+		OrderStatusUntriggered,
+		OrderStatusTriggered,
+		OrderStatusDeactivated,
+		OrderStatusActive,
+	}
+)
```
c9s, line 137: move these to types.go?

pkg/exchange/bybit/bybitapi/get_open_order_request.go
bailantaotao: ok

gx578007 (approved):

PR #1249, opened 2023-07-26
FEATURE: [bybit] support place order

bailantaotao:


c9s (approved):

pkg/exchange/bybit/exchange.go
```diff
@@ -169,3 +175,82 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
 
 	return orders, nil
 }
+
+func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (*types.Order, error) {
+	if len(order.Market.Symbol) == 0 {
+		log.Warnf("empty symbol, orderClientId: %s", order.ClientOrderID)
+		return nil, fmt.Errorf("empty symbol")
```
c9s, line 182: maybe just:

```
fmt.Errorf("order.Market is required: %+v", order)
```

and remove the warning, because SubmitOrder is a critical operation, we should stop the when error happens

pkg/exchange/bybit/exchange.go
```diff
@@ -169,3 +175,82 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
...
+		return nil, err
+	}
+	res, err := req.Do(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(res.OrderId) == 0 || res.OrderLinkId != order.ClientOrderID {
+		log.Warnf("unexpected response, orderId: %s, orderLinkId: %s, clientOrderId: %s", res.OrderId, res.OrderLinkId, order.ClientOrderID)
+		return nil, fmt.Errorf("unexpected response")
```
c9s, line 239: you can remove the warning and move the warning message to the error message, we can provide more details in the Errorf() call

pkg/exchange/bybit/exchange.go
```diff
@@ -169,3 +175,82 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
...
+	}
+
+	timeNow := time.Now()
+	return &types.Order{
```
c9s, line 243: If bybit api does not return the details of the order, maybe we need to query the order again by its order id?

gx578007 (approved):

PR #1255, opened 2023-07-28
FEATURE: [bybit] add query trade api

bailantaotao:


pkg/exchange/bybit/bybitapi/v3/types.go
```diff
@@ -0,0 +1,15 @@
+package v3
+
+type Side string
+
+const (
+	SideBuy  Side = "0"
+	SideSell Side = "1"
```
c9s, line 7: wondering if we can add a new type for parsing bool values like "0", "1", and 0, 1, "true", "false", true, false?

c9s (approved):

pkg/exchange/bybit/bybitapi/v3/types.go
bailantaotao: Considering that the returned parameter is named `IsBuy`, where 0 represents Buyer instead of 1 😵‍💫, I feel it might cause confusion, so I didn't add it.

`isMaker` are the same. 0 = Maker, 1 is Taker

c9s: Got it

c9s (approved):

pkg/exchange/bybit/exchange.go
```diff
@@ -315,3 +319,70 @@ func (e *Exchange) QueryClosedOrders(ctx context.Context, symbol string, since,
...
+QueryTrades queries trades by time range or trade id range.
+If options.StartTime is not specified, you can only query for records in the last 7 days.
+If you want to query for records older than 7 days, options.StartTime is required.
+It supports to query records up to 180 days.
+
+If the orderId is null, fromTradeId is passed, and toTradeId is null, then the result is sorted by
+ticketId in ascend. Otherwise, the result is sorted by ticketId in descend.
+
+** Here includes MakerRebate. If needed, let's discuss how to modify it to return in trade. **
+*/
```
gx578007, line 333: Could you also comment that startTime and endTime are inclusive or exclusive?  That will help a lot!

gx578007 (approved):

PR #1267, opened 2023-08-02
FEATURE: add custom heart beat func to StandardStream

bailantaotao:


pkg/types/stream.go
```diff
@@ -129,10 +134,12 @@ type StandardStreamEmitter interface {
 }
 
 func NewStandardStream() StandardStream {
-	return StandardStream{
+	s := StandardStream{
 		ReconnectC: make(chan struct{}, 1),
 		CloseC:     make(chan struct{}),
 	}
+	s.heartBeat = s.ping(pingInterval)
```
c9s, line 141: I think we still need the under layer ping/pong worker, you should not replace it

PR #1268, opened 2023-08-03
FEATURE: [bybit] implement order book streaming

bailantaotao:
- [x] add test


ycdesu (approved):

pkg/exchange/bybit/types.go
```diff
@@ -21,19 +31,108 @@ type WebSocketEvent struct {
...
+			Volume: ask[1],
+		})
+	}
+
+	return snapshot
+}
+
+const topicSeparator = "."
+
+func genTopicName(in ...interface{}) string {
```
c9s, line 124: can be just `genTopic` ?

pkg/exchange/bybit/types.go
```diff
@@ -21,19 +31,108 @@ type WebSocketEvent struct {
...
+	out := make([]string, len(in))
+	for k, v := range in {
+		out[k] = fmt.Sprintf("%v", v)
+	}
+	return strings.Join(out, topicSeparator)
+}
+
+func getTopicName(topic string) TopicType {
```
c9s, line 132: since the return type is TopicType, how about `getTopicType` ?

pkg/exchange/bybit/types.go
```diff
@@ -21,19 +31,108 @@ type WebSocketEvent struct {
...
+
+// PriceVolumeSlice represents a slice of price and value.
+//
+// index 0 is Bid/Ask price.
+// index 1 is Bid/Ask size. The *delta data* has size=0, which means that all quotations for this price have been filled or cancelled
+type PriceVolumeSlice [2]fixedpoint.Value
+
+type BookEvent struct {
+	// Symbol name
+	S string `json:"s"`
```
c9s, line 85: Field name doesn't have to be abbr, you can use full field name, like:

```
    Symbol string `json:"s"`
```

so that the field name could also explain the purpose of the field.

pkg/exchange/bybit/types.go
```diff
@@ -21,19 +31,108 @@ type WebSocketEvent struct {
...
+	// Bids. For snapshot stream, the element is sorted by price in descending order
+	B []PriceVolumeSlice `json:"b"`
```
c9s, line 87: Can be `Bids`

pkg/exchange/bybit/types.go
```diff
@@ -21,19 +31,108 @@ type WebSocketEvent struct {
...
+	// Asks. For snapshot stream, the element is sorted by price in ascending order
+	A []PriceVolumeSlice `json:"a"`
```
c9s, line 89: and `Asks`

pkg/exchange/bybit/types.go
```diff
@@ -21,19 +31,108 @@ type WebSocketEvent struct {
...
+	// The timestamp (ms) that the system generates the data
+	Ts   types.MillisecondTimestamp `json:"ts"`
+	Data interface{}                `json:"data"`
+}
+
+// PriceVolumeSlice represents a slice of price and value.
+//
+// index 0 is Bid/Ask price.
+// index 1 is Bid/Ask size. The *delta data* has size=0, which means that all quotations for this price have been filled or cancelled
+type PriceVolumeSlice [2]fixedpoint.Value
```
c9s, line 81: use `PriceVolume` to avoid confusion, because we already have some types for this kind of data:

image

c9s (approved):

PR #1274, opened 2023-08-07
FEATURE: [bybit] add balance snapshot event

bailantaotao:


ycdesu (approved):
Nice

pkg/exchange/bybit/stream.go
```diff
@@ -32,7 +32,8 @@ type Stream struct {
 	key, secret string
 	types.StandardStream
 
-	bookEventCallbacks []func(e BookEvent)
+	bookEventCallbacks   []func(e BookEvent)
+	walletEventCallbacks []func(e []*WalletEvent)
```
c9s, line 36: use slice of value?

c9s: oh it's a very large struct, ok let's use pointer LOL

c9s (approved):

pkg/exchange/bybit/stream.go
bailantaotao: Yeah~ that's why I chose him in the end

PR #1299, opened 2023-09-01
pkg/exchange: add time to SliceOrderBook

bailantaotao:
Provide server time if have.

Only Binance does not provide timestamp in initial order book, since the book is fetched from the REST API, which does not include this information. So i left it as time.Time{}.


pkg/cmd/orderbook.go
```diff
@@ -71,9 +71,10 @@ var orderbookCmd = &cobra.Command{
 			}
 
 			if bid, ask, ok := orderBook.BestBidAndAsk(); ok {
-				log.Infof("ASK | %f x %f / %f x %f | BID",
+				log.Infof("ASK | %f x %f / %f x %f | BID | %v",
```
c9s, line 74: %v has performance issue, you can use %s and call time.String() to stringify it

kbearXD (approved):

gx578007 (approved):

PR #1305, opened 2023-09-06
FEATURE: refactor okex to future use

MengShue:
1. To add `QueryTrades() , QueryClosedOrders(), QueryOrderTrades()` Need to refactor okex structure first, otherwise I can't use requestgen easily.
2. The following 1 - 2 PR will add  `QueryTrades() , QueryClosedOrders(), QueryOrderTrades()`

### DIFF
--- a/pkg/exchange/okex/exchange.go
+++ b/pkg/exchange/okex/exchange.go
@@ -35,10 +35,7 @@ type Exchange struct {
 }
 
 func New(key, secret, passphrase string) (*Exchange, error) {
-	client, err := okexapi.NewClient()
-	if err != nil {
-		return nil, err
-	}
+	client := okexapi.NewClient()
 
 	if len(key) > 0 && len(secret) > 0 {
 		client.Auth(key, secret, passphrase)
--- a/pkg/exchange/okex/exchange.go
+++ b/pkg/exchange/okex/exchange.go
@@ -34,7 +34,7 @@ type Exchange struct {
 	client *okexapi.RestClient
 }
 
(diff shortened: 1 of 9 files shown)


gx578007 (approved):

pkg/exchange/okex/okexapi/client.go
```diff
@@ -60,34 +61,26 @@ const (
...
 	Key, Secret, Passphrase string
-
-	TradeService      *TradeService
-	PublicDataService *PublicDataService
-	MarketDataService *MarketDataService
 }
 
-func NewClient() *RestClient {
+func NewClient() (*RestClient, error) {
 	u, err := url.Parse(RestBaseURL)
```
c9s, line 70: if we really want to check this error, could we move this parse into the init func.

constructor design should avoid return error, or the error should be checked outside of the constructor

```
var parsedBaseURL *url.URL
func init() {
      parsedBaseURL, err := url.Parse(RestBaseURL)
      if err != nil {
              panic(err)
      }
}
```

pkg/exchange/okex/exchange.go
```diff
@@ -46,15 +46,15 @@ func New(key, secret, passphrase string) *Exchange {
 		secret:     secret,
 		passphrase: passphrase,
 		client:     client,
-	}
+	}, nil
```
c9s, line 49: the error here can be removed

MengShue: Yes, I forgot it.

c9s (approved):

PR #1306, opened 2023-09-07
Improve: Update Binance futures account api to v2

andycheng123:


pkg/exchange/binance/futures.go
```diff
@@ -63,7 +63,9 @@ func (e *Exchange) TransferFuturesAccountAsset(ctx context.Context, asset string
 // Balance.Available = Wallet Balance(in Binance UI) - Used Margin
 // Balance.Locked = Used Margin
 func (e *Exchange) QueryFuturesAccount(ctx context.Context) (*types.Account, error) {
-	account, err := e.futuresClient.NewGetAccountService().Do(ctx)
+	//account, err := e.futuresClient.NewGetAccountService().Do(ctx)
+	req_account := e.futuresClient2.NewFuturesGetAccountRequest()
```
c9s, line 67: can be just req ?

c9s (approved):

pkg/exchange/binance/futures.go
andycheng123: there's another "req" down below

kbearXD (approved):

PR #1317, opened 2023-09-19
FIX: Wait for all routines to close while streaming is reconnecting

bailantaotao:
### Problem
During a websocket reconnect, the event status has unexpected ordering, such as going from 'connected' to 'disconnected'.

### Expected 
The correct order should be 'disconnected' -> 'connected'.

### Root Cause
EmitDisconnected is being called from within a goroutine, not the main thread.

### FIX 
image


pkg/types/waitgroup.go
```diff
@@ -0,0 +1,49 @@
...
+func (w *SyncGroup) Add(fn syncGroupFunc) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	w.wg.Add(1)
+	w.sgFuncs = append(w.sgFuncs, fn)
+}
+
+func (w *SyncGroup) Run() {
+	for _, goFn := range w.sgFuncs {
```
c9s, line 43: the mutex is used for w.sgFuncs? so you can do:

lock()
fs := w.sgFuncs
unlock()
for _ , f := fs { ... }

like this?

pkg/types/waitgroup.go
```diff
@@ -0,0 +1,49 @@
...
+	mu      sync.Mutex
+	sgFuncs []syncGroupFunc
+}
+
+func NewSyncGroup() SyncGroup {
+	return SyncGroup{}
+}
+
+func (w *SyncGroup) WaitAndClear() {
+	w.mu.Lock()
```
c9s, line 27: move the lock to where you want clear the sgFuncs?

c9s (approved):

pkg/types/waitgroup.go
bailantaotao: fixed

kbearXD (approved):

PR #1322, opened 2023-09-26
FEATURE: add get_all_fee_rate and supported_interval

MengShue:
1. add GetAllFeeRates for future use, ex. using in `stream.SetBeforeConnect()`
2. add supported interval, so I updated `toLocalInterval()`
3. add unit test for them respectively


pkg/exchange/okex/exchange.go
```diff
@@ -414,3 +417,25 @@ func (e *Exchange) QueryOrderTrades(ctx context.Context, q types.OrderQuery) ([]
...
+func (e *Exchange) SupportedInterval() map[types.Interval]int {
+	return okexapi.SupportedIntervals
+}
+
+func (e *Exchange) IsSupportedInterval(interval types.Interval) bool {
+	_, ok := okexapi.SupportedIntervals[interval]
+	return ok
+}
+
+func (e *Exchange) GetAllFeeRates(ctx context.Context) (okexapi.FeeRateList, error) {
```
bailantaotao, line 430: Why you need GetAllFeeRates?

pkg/exchange/okex/convert.go
```diff
@@ -206,11 +205,19 @@ func toGlobalOrderType(orderType okexapi.OrderType) (types.OrderType, error) {
 	return "", fmt.Errorf("unknown or unsupported okex order type: %s", orderType)
 }
 
-func toLocalInterval(src string) string {
-	var re = regexp.MustCompile(`\d+[hdw]`)
-	return re.ReplaceAllStringFunc(src, func(w string) string {
-		return strings.ToUpper(w)
-	})
+func toLocalInterval(interval types.Interval) (string, error) {
```
bailantaotao, line 208: Can you add unit test?

bailantaotao (approved):

pkg/exchange/okex/exchange.go
bailantaotao: Bybit uses it because the stream event does not return the fee and the fee currency.

MengShue: ok, I will remove it, I found out 'order' websocket would return fee, so I don't need GetAllFeeRates
https://www.okx.com/docs-v5/en/?shell#order-book-trading-trade-ws-order-channel

pkg/exchange/okex/convert.go
MengShue: in `Test_QueryKlines`, will add more unit test

pkg/exchange/okex/convert.go
```diff
@@ -206,11 +205,19 @@ func toGlobalOrderType(orderType okexapi.OrderType) (types.OrderType, error) {
...
+	if _, ok := okexapi.SupportedIntervals[interval]; !ok {
+		return "", fmt.Errorf("interval not supported: %s", interval)
```
c9s, line 210: "interval %s is not supported"

pkg/exchange/okex/convert.go
```diff
@@ -206,11 +205,19 @@ func toGlobalOrderType(orderType okexapi.OrderType) (types.OrderType, error) {
...
+	}
+
+	switch i := interval.String(); {
+	case strings.HasSuffix(i, "h"), strings.HasSuffix(i, "d"), strings.HasSuffix(i, "w"):
+		return strings.ToUpper(i), nil
+	case strings.HasSuffix(i, "mo"):
```
c9s, line 216: this will result 4 calls to HasSuffix, is there a simpler way to convert this?

pkg/exchange/okex/okexapi/types.go
```diff
@@ -0,0 +1,39 @@
+package okexapi
+
+import "github.com/c9s/bbgo/pkg/types"
+
+var (
```
c9s, line 5: should move this definition to the parent level package

pkg/exchange/okex/convert.go
MengShue: updated

pkg/exchange/okex/convert.go
MengShue: updated

pkg/exchange/okex/convert.go
```diff
@@ -206,11 +206,23 @@ func toGlobalOrderType(orderType okexapi.OrderType) (types.OrderType, error) {
...
+		return "", fmt.Errorf("interval %s is not supported", interval)
+	}
+
+	switch i := interval.String(); {
+	case strings.HasSuffix(i, "m"):
+		return i, nil
+	case strings.HasSuffix(i, "mo"):
+		return "1M", nil
+	default:
+		hdwRegex := regexp.MustCompile("h$|d$|w$")
```
c9s, line 220: should be written as:

```
\d+[hdw]$
```

bailantaotao:
Long time no response, so closed this PR.

PR #1325, opened 2023-09-26
fix: listenKeyExpired event sends string timestamp

zenixls2:
After listening for more than one hour, the event: ListenKeyExpired will be sent from binance server.
```text
ERROR websocket event parse error, message: {"e": "listenKeyExpired","E": "1695716647064","listenKey": "2rtkU3NxjtBFIw2zxhLMvTJ0jGsp87IZUDc4gY0F47aoF9pVvdxZUbozMog74fGU"} error=json: cannot unmarshal string into Go struct field ListenKeyExpired.E of type int64
```
To fix this elegantly, use json.Number to store timestamp.

### DIFF
--- a/pkg/exchange/binance/parse.go
+++ b/pkg/exchange/binance/parse.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"strconv"
 	"time"
 
 	"github.com/adshao/go-binance/v2/futures"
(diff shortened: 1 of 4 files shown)


pkg/exchange/binance/parse.go
```diff
@@ -17,8 +18,8 @@ import (
 )
 
 type EventBase struct {
-	Event string `json:"e"` // event name
-	Time  int64  `json:"E"` // event time
+	Event string      `json:"e"` // event name
+	Time  json.Number `json:"E"` // event time
```
c9s, line 22: you can use type MillisecondTimestamp here instead of this json.Number, it can unmarshal the timestamp

zenixls2: done

c9s (approved):

PR #1326, opened 2023-09-26
FIX: [bybit] fix bybit query trades

bailantaotao:
1. The previous implementation aimed to ensure that any time range is under control, but it seems that it did not achieve that. So i removed it.
```
	if options.StartTime != nil && options.EndTime != nil && options.EndTime.Sub(*options.StartTime) > halfYearDuration {
		return nil, fmt.Errorf("StartTime and EndTime cannot exceed 180 days, startTime: %v, endTime: %v, diff: %v",
			options.StartTime.String(),
			options.EndTime.String(),
			options.EndTime.Sub(*options.StartTime)/24)
	}
```

2. Modify the lastTradeId to 1 if it is currently 0. This adjustment will enable us to obtain additional data using the FromTradeId option. By specifying the FromTradeId, we can ensure that the query result is in ascending order.

### DIFF
--- a/pkg/exchange/bybit/exchange.go
+++ b/pkg/exchange/bybit/exchange.go
@@ -437,13 +437,6 @@ ticketId in ascend. Otherwise, the result is sorted by ticketId in descend.
 ** StartTime and EndTime cannot exceed 180 days. **
 */
 func (e *Exchange) QueryTrades(ctx context.Context, symbol string, options *types.TradeQueryOptions) (trades []types.Trade, err error) {
-	if options.StartTime != nil && options.EndTime != nil && options.EndTime.Sub(*options.StartTime) > halfYearDuration {
-		return nil, fmt.Errorf("StartTime and EndTime cannot exceed 180 days, startTime: %v, endTime: %v, diff: %v",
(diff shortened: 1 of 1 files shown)


narumiruna (approved):

pkg/types/exchange.go
```diff
@@ -153,8 +153,9 @@ type ExchangeRewardService interface {
...
 type TradeQueryOptions struct {
-	StartTime   *time.Time
-	EndTime     *time.Time
-	Limit       int64
-	LastTradeID uint64
+	StartTime      *time.Time
+	EndTime        *time.Time
+	Limit          int64
+	LastTradeID    uint64
+	MinimumTradeID uint64
```
c9s, line 160: you can add a comment for this MinimuTradeID? thanks

c9s:
Is this minimalTradeID backward compatible?

kbearXD (approved):

pkg/types/exchange.go
bailantaotao: ok

c9s (approved):

PR #1331, opened 2023-09-28
WIP: FEATURE: [fixedmaker] Use Binance's price data to avoid losses

narumiruna:


c9s:
Can we make it compatible with single exchange strategy? meaning, you can pass the session object for reference price and trading separately.

Or you can simplify this by copying into a new strategy called xfixedmaker (x means cross)

narumiruna:
I'd like to add a new strategy called xfixedmaker

c9s (approved):

PR #1335, opened 2023-10-04
FEATURE: add custom private channel support to max

c9s:


gx578007 (approved):

kbearXD (approved):

pkg/exchange/max/stream.go
```diff
@@ -119,14 +130,16 @@ func (s *Stream) handleConnect() {
 			}
 		}
 
+		log.Debugf("user data websocket filters: %v", filters)
```
bailantaotao, line 133: Do you need a sanity check for filters? to avoid zero subscription

c9s: It's ok, because empty filters = default filters

PR #1336, opened 2023-10-04
FEATURE: add randtrader strategy

narumiruna:
The `randtrader` strategy is very simple. It randomly places a buy or sell market order when the kline is closed.


config/randtrader.yaml
```diff
@@ -0,0 +1,9 @@
+---
+exchangeStrategies:
+  - on: max
+    randtrader:
```
c9s, line 4: how about just `rand` or `random`?

narumiruna: いいよ

c9s (approved):

PR #1344, opened 2023-10-18
FEATURE: [bitget] support book stream on bitget

bailantaotao:
1. The symbol is upper case, e.q. ETHUSDT
2. The `books` channel might return less than 200 bids/asks as per symbol's orderbook various from each other; The number of bids/asks is not a fixed value and may vary in the future.
3. The supported channels are books, books5, and books15.

### Reference
https://bitgetlimited.github.io/apidoc/en/spot/#depth-channel


pkg/exchange/bitget/stream.go
```diff
@@ -0,0 +1,167 @@
...
+func (s *Stream) handlerConnect() {
+	if s.PublicOnly {
+		// errors are handled in the syncSubscriptions, so they are skipped here.
+		_ = s.syncSubscriptions(WsEventSubscribe)
+	} else {
+		log.Error("*** PRIVATE API NOT IMPLEMENTED ***")
+	}
+}
+
+func (s *Stream) convertSubscription(sub types.Subscription) (WebsocketArg, error) {
```
c9s, line 132: can be a simple function if it does not use *Stream properties?

pkg/exchange/bitget/stream.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+		return w, nil
+	}
+
+	return w, fmt.Errorf("unsupported stream channel: %s", sub.Channel)
+}
+
+func (s *Stream) handleBooksEvent(o BookEvent) {
+	for _, book := range o.OrderBooks() {
+		switch {
```
c9s, line 159: switch o.Type ?

pkg/exchange/bitget/stream.go
```diff
@@ -0,0 +1,167 @@
...
+
+	switch sub.Channel {
+	case types.BookChannel:
+		w.Channel = ChannelOrderBook5
+
+		switch sub.Options.Depth {
+		case types.DepthLevel15:
+			w.Channel = ChannelOrderBook15
+		case types.DepthLevel200:
+			log.Info("*** The subscription events for the order book may return fewer than 200 bids/asks at a depth of 200. ***")
```
c9s, line 148: can be `Warn` ?

pkg/exchange/bitget/stream.go
```diff
@@ -0,0 +1,167 @@
...
+	if s.PublicOnly {
+		// errors are handled in the syncSubscriptions, so they are skipped here.
+		_ = s.syncSubscriptions(WsEventSubscribe)
+	} else {
+		log.Error("*** PRIVATE API NOT IMPLEMENTED ***")
+	}
+}
+
+func (s *Stream) convertSubscription(sub types.Subscription) (WebsocketArg, error) {
+	w := WebsocketArg{
```
c9s, line 133: name it `arg` or something?

pkg/exchange/bitget/types.go
```diff
@@ -0,0 +1,149 @@
...
+	Args []WebsocketArg `json:"args"`
+}
+
+// WsEvent is the lowest level of event type. We use this struct to convert the received data, so that we will know
+// whether the event belongs to WsOpEvent or WsChannelEvent.
+type WsEvent struct {
+	Arg WebsocketArg `json:"arg"`
+	// "op" and "channel" are exclusive.
+	*WsOpEvent
+	*WsChannelEvent
```
c9s, line 54: embed multiple struct like this may be dangerous?

pkg/exchange/bitget/types.go
```diff
@@ -0,0 +1,149 @@
...
+		Asks types.PriceVolumeSlice `json:"asks"`
+		// Order book on buy side, descending order
+		Bids     types.PriceVolumeSlice     `json:"bids"`
+		Ts       types.MillisecondTimestamp `json:"ts"`
+		Checksum int                        `json:"checksum"`
+	}
+
+	// internal use
+	Type   ActionType
+	instId string
```
c9s, line 134: Make this field public?

pkg/exchange/bitget/stream.go
gx578007, line 148: It's better to add the exchange name "bitget" in this log.

pkg/exchange/bitget/stream.go
```diff
@@ -0,0 +1,167 @@
...
+	case e.IsOp():
+		return e.WsOpEvent, nil
+
+	case e.IsChannel():
+		switch e.Arg.Channel {
+		case ChannelOrderBook, ChannelOrderBook5, ChannelOrderBook15:
+			var books BookEvent
+			err = json.Unmarshal(e.WsChannelEvent.Data, &books.Events)
+			if err != nil {
+				return nil, fmt.Errorf("failed to unmarshal data into BookEvent: %+v, err: %w", string(e.WsChannelEvent.Data), err)
```
ycdesu, line 111: Could we log `e.Arg.Channel` as well? This would allow us to immediately identify which channel is problematic upon encountering the error.

pkg/exchange/bitget/stream.go
bailantaotao: The log already with the filed 

https://github.com/c9s/bbgo/blob/5f8bda7d72933772b200bd746e5923361ec0e5eb/pkg/exchange/bitget/exchange.go#L14

pkg/exchange/bitget/types.go
bailantaotao: I think that is fine because we have a sanity check before calling the pointer struct. 
```
func (w *WsEvent) IsOp() bool {
	return w.WsOpEvent != nil && w.WsChannelEvent == nil
}

func (w *WsEvent) IsChannel() bool {
	return w.WsOpEvent == nil && w.WsChannelEvent != nil
}
```

This event is only used in parseWebSocketEvent. If it does not belong to any type, it will return an error.
```
func (s *Stream) parseWebSocketEvent(in []byte) (interface{}, error) {
	var e WsEvent

	err := json.Unmarshal(in, &e)
	if err != nil {
		return nil, err
	}

	switch {
	case e.IsOp():
		return e.WsOpEvent, nil

	case e.IsChannel():
		switch e.Arg.Channel {
		case ChannelOrderBook, ChannelOrderBook5, ChannelOrderBook15:
			...
			return &books, nil
		}
	}

	return nil, fmt.Errorf("unhandled websocket event: %+v", string(in))
}
```

pkg/exchange/bitget/stream.go
bailantaotao: The reason for the `info` is just a reminder, but a warning is acceptable. I have updated it.

pkg/exchange/bitget/stream.go
```diff
@@ -0,0 +1,167 @@
...
+	if opType != WsEventUnsubscribe && opType != WsEventSubscribe {
+		return fmt.Errorf("unexpected subscription type: %v", opType)
+	}
+
+	logger := log.WithField("opType", opType)
+	args := []WebsocketArg{}
+	for _, subscription := range s.Subscriptions {
+		arg, err := convertSubscription(subscription)
+		if err != nil {
+			logger.WithError(err).Errorf("convert error, subscription: %+v", subscription)
```
c9s, line 43: can simply return the error here I think

pkg/exchange/bitget/stream.go
```diff
@@ -0,0 +1,167 @@
...
+	stream.OnBooksEvent(stream.handleBooksEvent)
+	return stream
+}
+
+func (s *Stream) syncSubscriptions(opType WsEventType) error {
+	if opType != WsEventUnsubscribe && opType != WsEventSubscribe {
+		return fmt.Errorf("unexpected subscription type: %v", opType)
+	}
+
+	logger := log.WithField("opType", opType)
```
c9s, line 38: opType log tag might not be so helpful?

pkg/exchange/bitget/stream.go
bailantaotao: oh, it cannot be. The reason is that the `syncSubscriptions` function is invoked by both `handleConnect` and `Unsubscribe`. However, there is no error returned to the caller to inform them that an error has occurred.

handleConnect
```
stream.OnConnect(stream.handlerConnect)

func (s *Stream) handlerConnect() {
	if s.PublicOnly {
		// errors are handled in the syncSubscriptions, so they are skipped here.
		_ = s.syncSubscriptions(WsEventSubscribe)
	} else {
		log.Error("*** PRIVATE API NOT IMPLEMENTED ***")
	}
}
```


Unsubscribe
```
func (s *Stream) Unsubscribe() {
	// errors are handled in the syncSubscriptions, so they are skipped here.
	_ = s.syncSubscriptions(WsEventUnsubscribe)
	s.Resubscribe(func(old []types.Subscription) (new []types.Subscription, err error) {
		// clear the subscriptions
		return []types.Subscription{}, nil
	})
}
```

pkg/exchange/bitget/stream.go
bailantaotao: The op type represents a subscription or unsubscription. If an error occurs, we can identify the type that is causing the problem and address it.

pkg/exchange/bitget/stream.go
c9s: 👍 got it, thanks

c9s (approved):

gx578007 (approved):

PR #1359, opened 2023-10-24
FEATURE: add balance check strategy

narumiruna:
This strategy can remind you if the currency in your account matches your expectations. Otherwise, it will keep sending error messages to you.


c9s:
would be better to add this on the existing xalign or xnav strategy

PR #1363, opened 2023-10-25
FEATURE: [bitget] support ping/pong

bailantaotao:


gx578007 (approved):

pkg/exchange/bitget/stream.go
```diff
@@ -116,6 +137,36 @@ func (s *Stream) handleBookEvent(o BookEvent) {
...
+		select {
+
+		case <-ctx.Done():
+			return
+
+		case <-s.CloseC:
+			return
+
+		case <-pingTicker.C:
+			err := conn.WriteMessage(websocket.TextMessage, pingBytes)
```
c9s, line 160: when you need to write message, you need to lock it, or the built-in ping pong handler could cause panic

pkg/exchange/bitget/stream.go
```diff
@@ -1,12 +1,26 @@
...
+
+const (
+	// Client should keep ping the server in every 30 seconds. Server will close the connections which has no ping over
+	// 120 seconds(even when the client is still receiving data from the server)
+	pingInterval = 30 * time.Second
+)
+
+var (
+	pingBytes = []byte("ping")
+	pongBytes = []byte("pong")
```
c9s, line 23: bybit uses text message and not in json format?

bailantaotao: yes, it's not json 😂

c9s (approved):

PR #1372, opened 2023-10-27
FIX: [bybit][kucoin] fix negative volume, price precision

bailantaotao:
The precision should be positive. Bybit is according to Kucoin, so it is also incorrect..

Other exchanges are correct.

bybit document: https://bybit-exchange.github.io/docs/v5/market/instrument
```
              "basePrecision": "0.000001",
                "quotePrecision": "0.00000001",
```

kucoin document: https://www.kucoin.com/docs/rest/spot-trading/market-data/get-symbols-list
```
    "quoteIncrement": "0.000001",
    "priceIncrement": "0.000001",
```

### DIFF
--- a/pkg/exchange/bybit/convert.go
+++ b/pkg/exchange/bybit/convert.go
@@ -16,8 +16,8 @@ func toGlobalMarket(m bybitapi.Instrument) types.Market {
 	return types.Market{
 		Symbol:          m.Symbol,
 		LocalSymbol:     m.Symbol,
-		PricePrecision:  int(math.Log10(m.LotSizeFilter.QuotePrecision.Float64())),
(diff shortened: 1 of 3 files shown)


gx578007 (approved):

c9s (approved):

bailantaotao:
Since the test with -dnum flag caused a precision problem, for example:
```
a := -int(math.Log10(fixedpoint.NewFromFloat(0.000001).Float64()))
b := -int(math.Log10(fixedpoint.NewFromFloat(0.00000001).Float64()))
t.Log(a, b)
// 6, 7
```

But b must be 8.

After discussing with @c9s , we can solve it by using  NumFractionalDigits.

PR #1380, opened 2023-10-31
FEATURE: [bitget] support kline subscription on stream

bailantaotao:
https://bitgetlimited.github.io/apidoc/en/spot/#candlesticks-channel

The closed flag of Kline is not supported by Bitget. Therefore, we store the last Kline to compare the time duration and determine whether the Kline is closed.

### DIFF
--- a/pkg/exchange/bitget/stream.go
+++ b/pkg/exchange/bitget/stream.go
@@ -5,17 +5,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"time"
+	"github.com/gorilla/websocket"
+	"strings"
 
 	"github.com/c9s/bbgo/pkg/exchange/bitget/bitgetapi"
 	"github.com/c9s/bbgo/pkg/types"
-	"github.com/gorilla/websocket"
-)
-
-const (
-	// Client should keep ping the server in every 30 seconds. Server will close the connections which has no ping over
-	// 120 seconds(even when the client is still receiving data from the server)
-	pingInterval = 30 * time.Second
 )
 
 var (
@@ -29,11 +23,15 @@ type Stream struct {
 
 	bookEventCallbacks        []func(o BookEvent)
 	marketTradeEventCallbacks []func(o MarketTradeEvent)
+	KLineEventCallbacks       []func(o KLineEvent)
+
+	lastCandle map[string]types.KLine
 }
 
 func NewStream() *Stream {
 	stream := &Stream{
 		StandardStream: types.NewStandardStream(),
+		lastCandle:     map[string]types.KLine{},
 	}
 
 	stream.SetEndpointCreator(stream.createEndpoint)
@@ -44,6 +42,7 @@ func NewStream() *Stream {
 
 	stream.OnBookEvent(stream.handleBookEvent)
 	stream.OnMarketTradeEvent(stream.handleMaretTradeEvent)
+	stream.OnKLineEvent(stream.handleKLineEvent)
 	return stream
 }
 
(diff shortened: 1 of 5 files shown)


pkg/exchange/bitget/stream.go
```diff
@@ -212,7 +225,19 @@ func parseEvent(in []byte) (interface{}, error) {
...
 
-	case ChannelTrade:
+	case strings.HasPrefix(string(ch), "candle"):
+		var kline KLineEvent
+		err = json.Unmarshal(event.Data, &kline.Events)
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal data into KLineEvent, Arg: %+v Data: %s, err: %w", event.Arg, string(event.Data), err)
+		}
+
+		kline.actionType = event.Action
```
c9s, line 235: i think you can make these field public?

pkg/exchange/bitget/stream_test.go
```diff
@@ -106,6 +106,22 @@ func TestStream(t *testing.T) {
...
+		assert.NoError(t, err)
+
+		s.OnKLine(func(kline types.KLine) {
+			t.Log("got update", kline)
+		})
+		s.OnKLineClosed(func(kline types.KLine) {
+			t.Log("got closed update", kline)
+		})
+		c := make(chan struct{})
+		<-c
```
c9s, line 122: this test will be blocked?

pkg/exchange/bitget/types.go
```diff
@@ -260,3 +261,134 @@ type MarketTradeEvent struct {
...
+		Symbol:                   symbol,
+		StartTime:                types.Time(startTime),
+		EndTime:                  types.Time(startTime.Add(interval.Duration() - time.Millisecond)),
+		Interval:                 interval,
+		Open:                     k.OpenPrice,
+		Close:                    k.ClosePrice,
+		High:                     k.HighestPrice,
+		Low:                      k.LowestPrice,
+		Volume:                   k.Volume,
+		QuoteVolume:              0, // not supported
```
c9s, line 314: lint error, you should use fixedpoint.Zero here

pkg/exchange/bitget/stream.go
```diff
@@ -212,7 +225,19 @@ func parseEvent(in []byte) (interface{}, error) {
 		book.instId = event.Arg.InstId
 		return &book, nil
 
-	case ChannelTrade:
+	case strings.HasPrefix(string(ch), "candle"):
```
c9s, line 228: maybe you can pull out this case to a if statement that runs before this switch, so that you switch cases could be simplified ?

pkg/exchange/bitget/stream.go
bailantaotao: The action type is used in the current package, so i made it private.

pkg/exchange/bitget/stream_test.go
bailantaotao: yeah~but it will be skipped in CI tests. if we want to see something, we can run this test to view the results.

pkg/exchange/bitget/stream.go
bailantaotao: Or i pull it out to default? like: 
```

default: 
    if strings.HasPrefix(string(ch), "candle") {
        var kline KLineEvent
		err = json.Unmarshal(event.Data, &kline.Events)
		if err != nil {
			return nil, fmt.Errorf("failed to unmarshal data into KLineEvent, Arg: %+v Data: %s, err: %w", event.Arg, string(event.Data), err)
		}

		kline.actionType = event.Action
                kline.channel = ch
		kline.instId = event.Arg.InstId
		return &kline, nil
    }

    // other case are return err.

```
I hope that every event is handled in switch-case

c9s (approved):

PR #1392, opened 2023-11-05
FEATURE: [bitget] add query open orders

bailantaotao:
https://www.bitget.com/zh-CN/api-doc/spot/trade/Get-Unfilled-Orders


pkg/exchange/bitget/exchange.go
```diff
@@ -29,12 +35,15 @@ var (
...
 	queryTickersRateLimiter = rate.NewLimiter(rate.Every(time.Second/10), 5)
+	// queryOpenOrdersRateLimiter has its own rate limit. https://www.bitget.com/zh-CN/api-doc/spot/trade/Get-Unfilled-Orders
+	queryOpenOrdersRateLimiter = rate.NewLimiter(rate.Every(time.Second/10), 5)
 )
 
 type Exchange struct {
 	key, secret, passphrase string
 
-	client *bitgetapi.RestClient
+	client   *bitgetapi.RestClient
```
c9s, line 45: is this v3? call it v3client?

bailantaotao: No, i assume the client is v1, so i don't want to rename it

bailantaotao: I think it will be deprecated soon...

c9s (approved):

PR #1396, opened 2023-11-06
FEATURE: add ttl for position/grid2.profit_stats persistence

kbearXD:


pkg/types/position.go
```diff
@@ -65,6 +65,17 @@ type Position struct {
...
+
+	// ttl is the ttl to keep in persistence
+	ttl time.Duration
+}
+
+func (s *Position) SetTTL(ttl time.Duration) {
+	s.ttl = ttl
+}
+
+func (s *Position) Expiration() time.Duration {
```
c9s, line 77: do we check the returned Expiration()  in the caller? e.g., skip 0 expiration?

pkg/strategy/grid2/strategy.go
```diff
@@ -1835,13 +1836,17 @@ func (s *Strategy) Run(ctx context.Context, _ bbgo.OrderExecutor, session *bbgo.
 		s.ProfitSpread = s.Market.TruncatePrice(s.ProfitSpread)
 	}
 
+	s.logger.Infof("ttl: %s", s.PersistenceTTL.Duration())
```
c9s, line 1839: `using persistence ttl: %s`

pkg/strategy/grid2/strategy.go
```diff
@@ -1835,13 +1836,17 @@ func (s *Strategy) Run(ctx context.Context, _ bbgo.OrderExecutor, session *bbgo.
...
+
 	if s.GridProfitStats == nil {
 		s.GridProfitStats = newGridProfitStats(s.Market)
 	}
+	s.GridProfitStats.SetTTL(s.PersistenceTTL.Duration())
```
c9s, line 1844: should only call SetTTL when .PersistenceTTL is set and > 0

pkg/types/position.go
kbearXD: I think we don't need to do it. Because if the value is 0, it means it will not expire anymore.

pkg/strategy/grid2/active_order_recover.go
```diff
@@ -66,22 +66,26 @@ func (s *Strategy) recoverActiveOrdersPeriodically(ctx context.Context) {
...
 		select {
-
 		case <-ctx.Done():
 			return
 
 		case <-ticker.C:
-			if err := syncActiveOrders(ctx, opts); err != nil {
-				log.WithError(err).Errorf("unable to sync active orders")
+			s.recoverC <- struct{}{}
+			bbgo.Sync(ctx, s)
```
c9s, line 78: you can remove this?

pkg/strategy/grid2/active_order_recover.go
```diff
@@ -30,7 +30,7 @@ func (s *Strategy) initializeRecoverC() bool {
 
 	if s.recoverC == nil {
 		s.logger.Info("initializing recover channel")
-		s.recoverC = make(chan struct{}, 1)
+		s.recoverC = make(chan struct{}, 10)
```
c9s, line 33: why use 10?

c9s (approved):

PR #1397, opened 2023-11-06
FEATURE: [bybit] to periodically fetch the fee rate

bailantaotao:
To periodically fetch the fee rate.

The `feeRatePoller.getAllFeeRates` are copied from `stream.getAllFeeRates`

### DIFF
--- a/pkg/exchange/bybit/stream.go
+++ b/pkg/exchange/bybit/stream.go
@@ -47,8 +47,7 @@ type Stream struct {
 
 	key, secret        string
 	streamDataProvider StreamDataProvider
-	// TODO: update the fee rate at 7:00 am UTC; rotation required.
-	symbolFeeDetails map[string]*symbolFeeDetail
+	feeRateProvider    *feeRatePoller
 
 	bookEventCallbacks        []func(e BookEvent)
 	marketTradeEventCallbacks []func(e []MarketTradeEvent)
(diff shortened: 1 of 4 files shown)


gx578007 (approved):

ycdesu (approved):

pkg/exchange/bybit/stream.go
```diff
@@ -65,13 +64,17 @@ func NewStream(key, secret string, userDataProvider StreamDataProvider) *Stream
...
 		streamDataProvider: userDataProvider,
+		feeRateProvider:    newFeeRatePoller(userDataProvider),
 	}
 
 	stream.SetEndpointCreator(stream.createEndpoint)
 	stream.SetParser(stream.parseWebSocketEvent)
 	stream.SetDispatcher(stream.dispatchEvent)
 	stream.SetHeartBeat(stream.ping)
-	stream.SetBeforeConnect(stream.getAllFeeRates)
+	stream.SetBeforeConnect(func(ctx context.Context) error {
```
c9s, line 74: one question, is it BeforeConnect or BeforeStart? because we trigger OnStart only once, and OnConnect many times (disconnect and reconnect)

c9s (approved):

pkg/exchange/bybit/stream.go
bailantaotao: We call the BeforeConnect before DialAndConnect 
```
// Connect starts the stream and create the websocket connection
func (s *StandardStream) Connect(ctx context.Context) error {
	if s.beforeConnect != nil {
		if err := s.beforeConnect(ctx); err != nil {
			return err
		}
	}
	err := s.DialAndConnect(ctx)
	if err != nil {
		return err
	}

	// start one re-connector goroutine with the base context
	// reconnector goroutine does not exit when the connection is closed
	go s.reconnector(ctx)

	s.EmitStart()
	return nil
}
```

PR #1398, opened 2023-11-07
FIX: [bybit] rm retry and add fee recover

bailantaotao:
1. Remove the retry func and we can rely on the ticker instead.
2. Add fee rate recover.

REMARK:
The default rate for takers/markets on Bybit SPOT is 0.1%
https://www.bybit.com/en-US/help-center/article/Trading-Fee-Structure
image

### DIFF
--- a/pkg/exchange/bybit/stream.go
+++ b/pkg/exchange/bybit/stream.go
@@ -10,6 +10,7 @@ import (
 	"github.com/gorilla/websocket"
 
 	"github.com/c9s/bbgo/pkg/exchange/bybit/bybitapi"
+	"github.com/c9s/bbgo/pkg/fixedpoint"
 	"github.com/c9s/bbgo/pkg/types"
 	"github.com/c9s/bbgo/pkg/util"
 )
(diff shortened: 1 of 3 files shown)


pkg/exchange/bybit/stream.go
```diff
@@ -420,3 +436,28 @@ func (s *Stream) handleTradeEvent(events []TradeEvent) {
 		s.StandardStream.EmitTradeUpdate(*gTrade)
 	}
 }
+
+// symbolSplitter attempts to divide a trading pair into its base and quote currency. This is particularly useful for
+// calculating the fee currency when we are unable to retrieve the fee rate from the poller.
+//
+// Bybit represents currencies with uppercase in API responses.
+var symbolSplitter = regexp.MustCompile(`(\w+)(USDT|USDC|USD)$`)
```
c9s, line 444: not using markets?

bailantaotao: This scenario represents the worst-case situation where we're unable to fetch fee rates due to the Poller's failure to function properly. 🥲

c9s (approved):

gx578007 (approved):

PR #1399, opened 2023-11-07
FEATURE: [bitget] support submit order

bailantaotao:


pkg/exchange/bitget/exchange.go
```diff
@@ -182,9 +184,105 @@ func (e *Exchange) QueryAccountBalances(ctx context.Context) (types.BalanceMap,
...
+
+	// set side
+	side, err := toLocalSide(order.Side)
+	if err != nil {
+		return nil, err
+	}
+	req.Side(side)
+
+	// set quantity
+	orderQty := order.Quantity
```
c9s, line 216: since we're in the order method, this could be named as just "qty" or "q", "quantity", the prefix "order" is not necessary

pkg/exchange/bitget/exchange.go
```diff
@@ -182,9 +184,105 @@ func (e *Exchange) QueryAccountBalances(ctx context.Context) (types.BalanceMap,
...
+		return nil, fmt.Errorf("unexpected order id, resp: %#v, order: %#v", res, order)
+	}
+
+	orderId := res.OrderId
+	ordersResp, err := e.v2Client.NewGetUnfilledOrdersRequest().OrderId(orderId).Do(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to query open order by order id: %s, err: %w", orderId, err)
+	}
+
+	switch len(ordersResp) {
```
c9s, line 265: can be simply if len()... {} statement?

bailantaotao: Yes, it can. But i think switch is more readable than if-else. 

Because we need to check the order has already been filled by NewGetHistoryOrdersRequest

pkg/exchange/bitget/exchange.go
```diff
@@ -183,9 +185,105 @@ func (e *Exchange) QueryAccountBalances(ctx context.Context) (types.BalanceMap,
...
+		if err != nil {
+			return nil, err
+		}
+		orderQty = order.Quantity.Mul(ticker.Buy)
+	}
+	req.Size(order.Market.FormatQuantity(orderQty))
+
+	// set price
+	switch order.Type {
+	case types.OrderTypeLimit:
```
c9s, line 230: and LimitMaker?

bailantaotao: done

gx578007 (approved):

PR #1400, opened 2023-11-07
FEATURE: [bitget] support query trades

bailantaotao:
https://www.bitget.com/api-doc/spot/trade/Get-Fills


pkg/exchange/bitget/exchange.go
```diff
@@ -368,3 +370,59 @@ func (e *Exchange) CancelOrders(ctx context.Context, orders ...types.Order) erro
...
+	}
+	if options.EndTime != nil {
+		req.EndTime(options.EndTime.UnixMilli())
+		if options.StartTime == nil {
+			return nil, errors.New("start time is required for query trades if you take end time")
+		}
+	}
+
+	limit := options.Limit
+	if limit > queryOrdersLimit || limit <= 0 {
```
gx578007, line 399: Should be `queryTradesLimit`?

pkg/exchange/bitget/exchange.go
```diff
@@ -368,3 +370,59 @@ func (e *Exchange) CancelOrders(ctx context.Context, orders ...types.Order) erro
...
+		log.Debugf("limtit is exceeded or zero, update to %d, got: %d", queryOrdersLimit, options.Limit)
+		limit = queryOrdersLimit
+	}
+	req.Limit(strconv.FormatInt(limit, 10))
+
+	if err := queryTradeRateLimiter.Wait(ctx); err != nil {
+		return nil, fmt.Errorf("trade rate limiter wait error: %w", err)
+	}
```
gx578007, line 407: How about moving this rate limit check to the beginning of this function?

pkg/exchange/bitget/exchange.go
```diff
@@ -368,3 +370,59 @@ func (e *Exchange) CancelOrders(ctx context.Context, orders ...types.Order) erro
...
+// ** StartTime and EndTime cannot exceed 90 days. **
+func (e *Exchange) QueryTrades(ctx context.Context, symbol string, options *types.TradeQueryOptions) (trades []types.Trade, err error) {
+	if options.LastTradeID != 0 {
+		log.Warn("!!!BITGET EXCHANGE API NOTICE!!! The trade of response is in descending order, so the last trade id not supported.")
+	}
+
+	req := e.v2Client.NewGetTradeFillsRequest()
+	req.Symbol(symbol)
+
+	if options.StartTime != nil {
```
c9s, line 388: if both StartTime and EndTime are given, you will need to check the time range?

ycdesu: Please ensure the `end >= start`

pkg/exchange/bitget/exchange.go
bailantaotao: Oh, both the trade and order have a limit of 100, so I am using it. I will rename it to `queryLimit`.

pkg/exchange/bitget/exchange.go
bailantaotao: The thought is to call it before req.Do, so that we can stay near the rate limit.

pkg/exchange/bitget/exchange.go
bailantaotao: I was just thinking that the caller should use batch.TradeBatchQuery, so I missed this part.

c9s (approved):

PR #1402, opened 2023-11-08
FEATURE: inventory skew

narumiruna:


pkg/strategy/fixedmaker/inventory_skew.go
```diff
@@ -0,0 +1,70 @@
...
+	TargetBaseRatio          fixedpoint.Value `json:"targetBaseRatio"`
+}
+
+func (s *InventorySkew) Initialize(session *bbgo.ExchangeSession, market types.Market) error {
+	s.Market = market
+	s.Session = session
+	return nil
+}
+
+func (s *InventorySkew) CalculateBidAskRatios(ctx context.Context, quantity fixedpoint.Value) (*InventorySkewBidAskRatios, error) {
```
c9s, line 37: maybe we can make it not to depend on session? because it could make the test harder to write.

how about just pass base balance, quote balance and the current price of base?

c9s: or maybe you can keep this InventorySkew, and pull out the formula to a function?

narumiruna: I would like to make it not to depend on session.

c9s:
@narumiruna 這隻可以先進？我們之後再調整 API ~~

narumiruna:
> @narumiruna 這隻可以先進？我們之後再調整 API ~~

好

c9s (approved):

c9s:
@narumiruna 要進了嗎

narumiruna:
> @narumiruna 要進了嗎

go

PR #1404, opened 2023-11-09
FEATURE: [bitget] support cancel order

bailantaotao:
https://www.bitget.com/api-doc/spot/trade/Cancel-Order


pkg/exchange/bitget/exchange.go
```diff
@@ -383,9 +385,53 @@ func (e *Exchange) QueryClosedOrders(ctx context.Context, symbol string, since,
...
+func (e *Exchange) CancelOrders(ctx context.Context, orders ...types.Order) (errs error) {
+	if len(orders) == 0 {
+		return nil
+	}
+
+	for _, order := range orders {
+		req := e.client.NewCancelOrderRequest()
+
+		reqId := ""
+		switch {
```
c9s, line 397: switch case works, but why not use if else clause?

bailantaotao: just i don't like the nested if-else statements

c9s (approved):

gx578007 (approved):

PR #1405, opened 2023-11-09
FIX: [grid2] use rest quote to place the last order when opening grid

kbearXD:


pkg/strategy/grid2/strategy.go
```diff
@@ -1378,9 +1382,21 @@ func (s *Strategy) generateGridOrders(totalQuote, totalBase, lastPrice fixedpoin
...
 
-			if usedQuote.Add(quoteQuantity).Compare(totalQuote) > 0 {
-				s.logger.Warnf("used quote %f > total quote %f, this should not happen", usedQuote.Add(quoteQuantity).Float64(), totalQuote.Float64())
-				continue
+			// because the precision issue, we need to round up quote quantity and add it into used quote
+			// e.g. quote we calculate : 8888.85, but it may lock 8888.9 due to their precision.
+			roundUpQuoteQuantity := quoteQuantity.Round(s.Market.PricePrecision, fixedpoint.Up)
+			if usedQuote.Add(roundUpQuoteQuantity).Compare(totalQuote) > 0 {
+				if i > 0 {
+					s.logger.Errorf("used quote %f > total quote %f, this should not happen", usedQuote.Add(quoteQuantity).Float64(), totalQuote.Float64())
```
c9s, line 1390: do we need to log here? we have already return the error with the same message

pkg/strategy/grid2/strategy.go
```diff
@@ -1378,9 +1382,21 @@ func (s *Strategy) generateGridOrders(totalQuote, totalBase, lastPrice fixedpoin
...
+					return nil, fmt.Errorf("used quote %f > total quote %f, this should not happen", usedQuote.Add(quoteQuantity).Float64(), totalQuote.Float64())
+				} else {
+					restQuote := totalQuote.Sub(usedQuote)
+					quantity = restQuote.Div(price).Round(s.Market.VolumePrecision, fixedpoint.Down)
+					if s.Market.MinQuantity.Compare(quantity) > 0 {
+						s.logger.Errorf("the round down quantity (%s) is less than min quantity (%s), we cannot place this order", quantity, s.Market.MinQuantity)
```
c9s, line 1396: ditto

c9s (approved):

PR #1415, opened 2023-11-14
FEATURE: [bitget] use v2 tickers

bailantaotao:
https://www.bitget.com/api-doc/spot/market/Get-Tickers


pkg/exchange/bitget/exchange.go
```diff
@@ -113,14 +113,17 @@ func (e *Exchange) QueryTicker(ctx context.Context, symbol string) (*types.Ticke
 		return nil, fmt.Errorf("ticker rate limiter wait error: %w", err)
 	}
 
-	req := e.client.NewGetTickerRequest()
+	req := e.v2Client.NewGetTickersRequest()
```
c9s, line 116: how about v2client? all in lower case

c9s (approved):

pkg/exchange/bitget/exchange.go
bailantaotao: ok, update in next

bailantaotao: Ah, I updated this PR because the checks were not working after another PR was merged...

PR #1419, opened 2023-11-15
FIX: [bitget] use the now - 90 days instead of return err if since is 90 days earlier

bailantaotao:


pkg/types/strint.go
```diff
@@ -41,3 +41,10 @@ func (s *StrInt64) UnmarshalJSON(body []byte) error {
 
 	return nil
 }
+
+func (s *StrInt64) String() string {
```
c9s, line 45: you can use value receiver here?

pkg/exchange/bitget/exchange.go
```diff
@@ -289,7 +289,7 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
 	// 2. The query oepn/closed order does not including the `force` in SPOT.
```
c9s, line 289: "e"open

c9s, line 289: and "does not include"

pkg/exchange/bitget/exchange.go
```diff
@@ -289,7 +289,7 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
...
 	// If we support FOK/IOC, but you can't query them, that would be unreasonable.
 	// The other case to consider is 'PostOnly', which is a trade-off because we want to support 'xmaker'.
-	if order.TimeInForce != types.TimeInForceGTC {
+	if order.TimeInForce != types.TimeInForceGTC && len(order.TimeInForce) != 0 {
```
c9s, line 292: put the length check before the gtc check?

pkg/exchange/bitget/exchange.go
```diff
@@ -397,18 +397,22 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
...
 // ** Since is inclusive, Until is exclusive. If you use a time range to query, you must provide both a start time and an end time. **
 // ** Since and Until cannot exceed 90 days. **
-// ** Since from the last 90 days can be queried. **
+// ** Since from the last 90 days can be queried **
 func (e *Exchange) QueryClosedOrders(ctx context.Context, symbol string, since, until time.Time, lastOrderID uint64) (orders []types.Order, err error) {
-	if time.Since(since) > queryMaxDuration {
-		return nil, fmt.Errorf("start time from the last 90 days can be queried, got: %s", since)
+	newSince := since
+
+	if time.Since(newSince) > queryMaxDuration {
```
c9s, line 407: name it maxHistoricalDataQueryPeriod?

c9s: and you should also modify the "until"?

bailantaotao: Are you suggesting changing 'until' to 'now'? I did consider it, but later on, I noticed that everyone else was using 'time.Now,' so I didn't make the change

pkg/types/strint.go
bailantaotao: No, I can't because we have a JSON marshaler on a pointer receiver. The combination of a pointer and value is not recommended by Golang

pkg/types/exchange.go
```diff
@@ -26,13 +26,13 @@ func (n *ExchangeName) UnmarshalJSON(data []byte) error {
 	}
 
 	switch s {
-	case "max", "binance", "okex", "kucoin":
+	case "binance", "bitget", "bybit", "max", "okex", "kucoin":
```
gx578007, line 29: How about using a slice for this list and checking for existence?

pkg/types/exchange.go
```diff
@@ -26,13 +26,13 @@ func (n *ExchangeName) UnmarshalJSON(data []byte) error {
...
 		*n = ExchangeName(s)
 		return nil
 
 	}
 
-	return fmt.Errorf("unknown or unsupported exchange name: %s, valid names are: max, binance, okex, kucoin", s)
+	return fmt.Errorf("unknown or unsupported exchange name: %s, valid names are: binance, bitget, bybit, max, okex, kucoin", s)
```
gx578007, line 35: With a slice, you don't have to duplicate the list twice.

gx578007 (approved):

c9s (approved):

pkg/types/exchange.go
bailantaotao: that's done

PR #1421, opened 2023-11-17
FEATURE: use new max v3 api to query closed orders by timestamp

kbearXD:


pkg/exchange/max/exchange.go
```diff
@@ -311,6 +315,47 @@ func (e *Exchange) queryClosedOrdersByLastOrderID(ctx context.Context, symbol st
...
+	market := toLocalSymbol(symbol)
+	walletType := maxapi.WalletTypeSpot
+	if e.MarginSettings.IsMargin {
+		walletType = maxapi.WalletTypeMargin
+	}
+
+	req := e.v3client.NewGetWalletClosedOrdersRequest(walletType).Market(market).Limit(1000)
+
+	if !until.IsZero() {
+		req.Timestamp(until)
```
c9s, line 332: what will happen if we don't set Timestamp?

pkg/exchange/max/exchange.go
```diff
@@ -311,6 +315,47 @@ func (e *Exchange) queryClosedOrdersByLastOrderID(ctx context.Context, symbol st
...
+	}
+
+	maxOrders, err := req.Do(ctx)
```
c9s, line 335: and, what's the default ordering?

pkg/exchange/max/exchange.go
```diff
@@ -311,6 +315,47 @@ func (e *Exchange) queryClosedOrdersByLastOrderID(ctx context.Context, symbol st
...
+	if err != nil {
+		return orders, err
+	}
+
+	for _, maxOrder := range maxOrders {
+		if maxOrder.CreatedAt.Time().Before(since) {
+			break
```
c9s, line 342: break? maybe just "continue" and skip it?

pkg/exchange/max/maxapi/v3/get_wallet_closed_orders_request.go
```diff
@@ -0,0 +1,27 @@
...
+
+//go:generate GetRequest -url "/api/v3/wallet/:walletType/orders/closed" -type GetWalletClosedOrdersRequest -responseType []Order
+type GetWalletClosedOrdersRequest struct {
+	client requestgen.AuthenticatedAPIClient
+
+	walletType WalletType `param:"walletType,slug,required"`
+
+	market    string     `param:"market,required"`
+	timestamp *time.Time `param:"timestamp,milliseconds,omitempty"`
+	orderBy   *string    `param:"order_by,omitempty" validValues:"asc,desc,asc_updated_at,desc_updated_at"`
```
c9s, line 25: can you define these strings as constants ? "asc,desc,asc_updated_at,desc_updated_at"

c9s: just like WalletType

pkg/exchange/max/exchange.go
kbearXD: it will use now timestamp as default to query closed orders

pkg/exchange/max/exchange.go
kbearXD: desc. there are four order type - asc, desc, asc_updated_at, desc_updated_at

c9s: but we should use asc here

kbearXD: ok!

c9s (approved):

pkg/exchange/max/exchange.go
```diff
@@ -268,6 +268,10 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
 // lastOrderID is not supported on MAX
 func (e *Exchange) QueryClosedOrders(ctx context.Context, symbol string, since, until time.Time, lastOrderID uint64) ([]types.Order, error) {
 	log.Warn("!!!MAX EXCHANGE API NOTICE!!! the since/until conditions will not be effected on closed orders query, max exchange does not support time-range-based query")
+	if !since.IsZero() || !until.IsZero() {
```
c9s, line 271: should be until.IsZero()? since the api only support one timestamp parameter?

kbearXD: We can give one of it. If we give until, we will use 2018-01-01 as timestamp to query and filter it by until.

PR #1428, opened 2023-11-21
FEATURE: use max v3 new open orders api

kbearXD:


pkg/exchange/max/exchange.go
```diff
@@ -253,18 +253,36 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
...
+
+			orders = append(orders, *order)
+		}
+
+		if len(maxOrders) < int(limit) {
+			break
+		}
+
+		// open orders api will get the open orders which created_at >= since, so we need to add 1 ms to avoid duplicated
+		since = since.Add(1 * time.Millisecond)
```
gx578007, line 285: sleep between each polling?

kbearXD: because the since/until in MAX API is inclusive, so we need to add 1ms to avoid duplicated order

ycdesu (approved):

pkg/exchange/max/exchange.go
```diff
@@ -253,18 +253,36 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
...
 
-	maxOrders, err := e.v3client.NewGetWalletOpenOrdersRequest(walletType).Market(market).Do(ctx)
-	if err != nil {
-		return orders, err
-	}
-
-	for _, maxOrder := range maxOrders {
-		order, err := toGlobalOrder(maxOrder)
+	// timestamp can't be negative, so we need to use time which epochtime is > 0
+	var since time.Time = time.Date(2018, time.January, 1, 0, 0, 0, 0, time.Local)
```
c9s, line 257: use getLaunchDate?

kbearXD: OK

pkg/exchange/max/exchange.go
```diff
@@ -246,28 +246,59 @@ func (e *Exchange) QueryOrder(ctx context.Context, q types.OrderQuery) (*types.O
...
+	for {
+		req := e.v3client.NewGetWalletOpenOrdersRequest(walletType).Market(market).Timestamp(since).OrderBy(maxapi.OrderByAsc).Limit(limit)
+		maxOrders, err := req.Do(ctx)
 		if err != nil {
-			return orders, err
+			return nil, err
 		}
 
-		orders = append(orders, *order)
+		noDuplicatedCnt := 0
```
c9s, line 274: is there a better variable name for this?

c9s (approved):

PR #1434, opened 2023-11-25
CHORE: add rollbar token environment variable

narumiruna:


pkg/cmd/root.go
```diff
@@ -75,7 +75,12 @@ var RootCmd = &cobra.Command{
 			}
 		}
 
-		if token := viper.GetString("rollbar-token"); token != "" {
+		token := viper.GetString("rollbar-token")
```
c9s, line 78: this key "rollbar-token" will be translated into "ROLLBAR_TOKEN"

PR #1436, opened 2023-11-29
FEATURE: cancel maker orders and open take profit order

kbearXD:


pkg/strategy/dca2/maker.go
```diff
@@ -9,6 +9,10 @@ import (
 	"github.com/c9s/bbgo/pkg/types"
 )
 
+type cancelOrdersByGroupIDApi interface {
+	CancelOrdersByGroupID(ctx context.Context, groupID int64) ([]types.Order, error)
+}
```
c9s, line 14: we don't need this?

pkg/strategy/dca2/maker.go
```diff
@@ -105,3 +109,24 @@ func (s *Strategy) generateMakerOrder(short bool, budget, price, margin fixedpoi
...
+		cancelledOrders, err := e.CancelOrdersByGroupID(ctx, int64(s.OrderGroupID))
+		if err != nil {
+			return err
+		}
+
+		for _, cancelledOrder := range cancelledOrders {
+			s.logger.Info("CANCEL ", cancelledOrder.String())
+		}
+	} else {
+		if err := s.OrderExecutor.ActiveMakerOrders().GracefulCancel(ctx, s.Session.Exchange); err != nil {
```
c9s, line 126: can be just s.OrderExecutor.GracefulCancel

pkg/strategy/dca2/takeProfit.go
```diff
@@ -0,0 +1,42 @@
+package dca2
```
c9s, line 1: filenames should be snake case

pkg/strategy/dca2/takeProfit.go
```diff
@@ -0,0 +1,42 @@
...
+func (s *Strategy) openTakeProfitOrders(ctx context.Context) error {
+	s.logger.Info("[DCA] open take profit orders")
+	takeProfitOrder := s.generateTakeProfitOrder(s.Short, s.Position)
+	createdOrders, err := s.OrderExecutor.SubmitOrders(ctx, takeProfitOrder)
+	if err != nil {
+		return err
+	}
+
+	for _, createdOrder := range createdOrders {
+		s.logger.Info("SUBMIT TAKE PROFIT ORDER ", createdOrder.String())
```
c9s, line 19: can we reuse the order debug function defined in the grid2 package?

pkg/strategy/dca2/takeProfit.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	return nil
+}
+
+func (s *Strategy) generateTakeProfitOrder(short bool, position *types.Position) types.SubmitOrder {
```
c9s, line 25: what's the short here?

pkg/strategy/dca2/takeProfit.go
```diff
@@ -0,0 +1,42 @@
...
+	takeProfitRatio := s.TakeProfitRatio
```
c9s, line 26: you can move this to the func arg, so that you just pass s.takeProfitRatio from the caller

pkg/strategy/dca2/maker.go
kbearXD: if we use GracefulCancel, we still need to GracefulCancel(ctx, s.OrderExecutor.ActiveMakerOrders().Orders()...)
I think it's almost the same ?

pkg/strategy/dca2/maker.go
kbearXD: Yes, I think GracefulCancel is enough.

pkg/strategy/dca2/maker.go
c9s: yes

c9s (approved):

PR #1439, opened 2023-12-04
FIX: fix list closed orders api limit

kbearXD:


pkg/exchange/max/exchange.go
```diff
@@ -325,11 +341,21 @@ func (e *Exchange) queryClosedOrdersByLastOrderID(
...
 }
 
-func (e *Exchange) queryClosedOrdersByTime(ctx context.Context, symbol string, since, until time.Time) (orders []types.Order, err error) {
+func (e *Exchange) queryClosedOrdersByTime(ctx context.Context, symbol string, since, until time.Time, orderByType maxapi.OrderByType, limit uint) (orders []types.Order, err error) {
 	if err := e.closedOrderQueryLimiter.Wait(ctx); err != nil {
 		return orders, err
 	}
 
+	// there is since limit for closed orders API
+	sinceLimit := time.Date(2018, time.January, 1, 0, 0, 0, 0, time.Local)
```
ycdesu, line 350: This time can be moved to package level to be shared with other functions.

ycdesu (approved):

pkg/exchange/max/exchange.go
```diff
@@ -270,18 +270,34 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
...
-	log.Warn("!!!MAX EXCHANGE API NOTICE!!! the since/until conditions will not be effected on closed orders query, max exchange does not support time-range-based query")
 	if !since.IsZero() || !until.IsZero() {
-		return e.queryClosedOrdersByTime(ctx, symbol, since, until)
+		return e.queryClosedOrdersByTime(ctx, symbol, since, until, maxapi.OrderByAsc, 1000)
 	}
 
 	return e.queryClosedOrdersByLastOrderID(ctx, symbol, lastOrderID)
 }
 
+func (e *Exchange) QueryClosedOrdersDesc(
```
bailantaotao, line 283: 沒有人呼叫他？

bailantaotao (approved):

pkg/exchange/max/exchange.go
kbearXD: dca2 會用到

pkg/exchange/max/exchange.go
kbearXD: will use getLaunchDate

pkg/exchange/max/exchange.go
```diff
@@ -270,18 +270,34 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
...
+	ctx context.Context, symbol string, since, until time.Time, lastOrderID uint64,
```
c9s, line 284: lastOrderID should be optional, not all exchanges support this

pkg/exchange/max/exchange.go
```diff
@@ -270,18 +270,34 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
...
+) ([]types.Order, error) {
+	closedOrders, err := e.queryClosedOrdersByTime(ctx, symbol, since, until, maxapi.OrderByDesc, 1000)
+	if lastOrderID == 0 {
+		return closedOrders, err
+	}
+
+	var filterClosedOrders []types.Order
+	for _, closedOrder := range filterClosedOrders {
+		if closedOrder.OrderID > lastOrderID {
```
c9s, line 293: should be compared only when lastOrderID is not zero

pkg/exchange/max/exchange.go
```diff
@@ -270,18 +270,34 @@ func (e *Exchange) QueryOpenOrders(ctx context.Context, symbol string) (orders [
...
+func (e *Exchange) QueryClosedOrdersDesc(
+	ctx context.Context, symbol string, since, until time.Time, lastOrderID uint64,
+) ([]types.Order, error) {
+	closedOrders, err := e.queryClosedOrdersByTime(ctx, symbol, since, until, maxapi.OrderByDesc, 1000)
+	if lastOrderID == 0 {
+		return closedOrders, err
+	}
+
+	var filterClosedOrders []types.Order
+	for _, closedOrder := range filterClosedOrders {
```
c9s, line 292: you're iterating an empty array?

pkg/exchange/max/exchange.go
```diff
@@ -325,11 +341,24 @@ func (e *Exchange) queryClosedOrdersByLastOrderID(
 	return types.SortOrdersAscending(orders), nil
 }
 
-func (e *Exchange) queryClosedOrdersByTime(ctx context.Context, symbol string, since, until time.Time) (orders []types.Order, err error) {
+func (e *Exchange) queryClosedOrdersByTime(ctx context.Context, symbol string, since, until time.Time, orderByType maxapi.OrderByType, limit uint) (orders []types.Order, err error) {
```
c9s, line 344: we can simply use the default limit

c9s: by the maximum number

pkg/exchange/max/exchange.go
kbearXD: Yes, check lastOrderID at line 287.

pkg/exchange/max/exchange.go
kbearXD: You are right.....

pkg/exchange/max/exchange.go
kbearXD: I think we can think if lastOrderID is zero means we don't need to care it ?

pkg/exchange/max/exchange.go
kbearXD: ok!

c9s (approved):

PR #1460, opened 2023-12-18
FIX: fix and improve session UpdatePrice method

c9s:


pkg/bbgo/session.go
```diff
@@ -717,7 +737,17 @@ func (session *ExchangeSession) UpdatePrices(ctx context.Context, currencies []s
 	var lastTime time.Time
 	for k, v := range tickers {
 		// for {Crypto}/USDT markets
-		session.lastPrices[k] = v.Last
+		// map things like BTCUSDT = {price}
+		if market, ok := markets[k]; ok {
+			if types.IsFiatCurrency(market.BaseCurrency) {
+				session.lastPrices[k] = v.Last.Div(fixedpoint.One)
```
gx578007, line 743: What's the purpose of being divided by one?

c9s: reversed quoting, e.g. USDT/TWD

PR #1463, opened 2023-12-19
FEATURE: [bollmaker] add emaCross signal 

c9s:


pkg/strategy/bollmaker/strategy.go
```diff
@@ -435,14 +445,20 @@ func (s *Strategy) placeOrders(ctx context.Context, midPrice fixedpoint.Value, k
 		}
 	}
 
+	if !s.shouldBuy {
```
zenixls2, line 448: Not shouldBuy != Should not buy

Maybe renane this variable to shouldNotBuy?

c9s: shouldBy is set to true by default, would like to keep it simple to not to put a reversed logic in the variable ^^||

kbearXD (approved):

narumiruna (approved):

PR #1478, opened 2024-01-03
FEATURE: rename and use specific profit stats

kbearXD:


pkg/strategy/dca2/profit_stats.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/c9s/bbgo/pkg/fixedpoint"
+	"github.com/c9s/bbgo/pkg/types"
+)
+
+type ProfitStats struct {
+	Symbol string       `json:"symbol"`
+	Market types.Market `json:"market,omitempty"`
+
+	CreatedAt       time.Time        `json:"since,omitempty"`
```
bailantaotao, line 14: Why is the json tag 'since' instead of 'CreatedAt'?

bailantaotao (approved):

c9s (approved):

PR #1479, opened 2024-01-04
FIX: [xgap] fix order cancel error

narumiruna:
- avoid canceling the order that has already been fulfilled
- remove unused constant

### DIFF
--- a/pkg/strategy/xgap/strategy.go
+++ b/pkg/strategy/xgap/strategy.go
@@ -18,8 +18,6 @@ import (
 
 const ID = "xgap"
 
-const stateKey = "state-v1"
-
 var log = logrus.WithField("strategy", ID)
 
 var StepPercentageGap = fixedpoint.NewFromFloat(0.05)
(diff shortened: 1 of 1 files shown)


bailantaotao (approved):

pkg/strategy/xgap/strategy.go
```diff
@@ -355,9 +358,11 @@ func (s *Strategy) CrossRun(ctx context.Context, _ bbgo.OrderExecutionRouter, se
 					log.WithError(err).Error("order submit error")
 				}
 
+				s.activeOrderBook.Add(createdOrders...)
+
 				time.Sleep(time.Second)
 
-				if err := tradingSession.Exchange.CancelOrders(ctx, createdOrders...); err != nil {
+				if err := tradingSession.Exchange.CancelOrders(ctx, s.activeOrderBook.Orders()...); err != nil {
```
kbearXD, line 365: Can we use s.activeOrderBook.GracefulCancel(ctx) here ?

narumiruna: ok~

c9s (approved):

PR #1486, opened 2024-01-08
FEATURE: [okx] support market trade streaming

bailantaotao:


pkg/exchange/okex/parse.go
```diff
@@ -363,3 +373,59 @@ func parseOrder(v *fastjson.Value) ([]okexapi.OrderDetails, error) {
...
+		return types.SideTypeSell, nil
+
+	default:
+		return types.SideType(side), fmt.Errorf("unexpected side: %s", side)
+	}
+}
+
+type MarketTradeEvent struct {
+	InstId    string                     `json:"instId"`
+	TradeId   string                     `json:"tradeId"`
```
c9s, line 392: you can use StrInt here?

c9s: see pkg/types/strint.go

c9s: or maybe you can create a StrUint for this?

c9s (approved):

pkg/exchange/okex/parse.go
bailantaotao: I asked for official support, which confirmed that the trade ID will be numeric. Therefore, I made the necessary changes to fix it.

c9s (approved):

PR #1487, opened 2024-01-08
FIX: [bitget] ignore offline symbols

c9s:


pkg/exchange/bitget/convert.go
```diff
@@ -25,9 +25,10 @@ func toGlobalBalance(asset v2.AccountAsset) types.Balance {
 }
 
 func toGlobalMarket(s v2.Symbol) types.Market {
-	if s.Status != v2.SymbolOnline {
+	if s.Status != v2.SymbolStatusOnline {
 		log.Warnf("The symbol %s is not online", s.Symbol)
```
bailantaotao, line 29: Can you add status to log?

c9s: Sure

bailantaotao (approved):

PR #1489, opened 2024-01-09
REFACTOR: [okx] refactor account info

bailantaotao:


pkg/exchange/okex/stream.go
```diff
@@ -211,7 +203,12 @@ func (s *Stream) createEndpoint(ctx context.Context) (string, error) {
 func (s *Stream) dispatchEvent(e interface{}) {
 	switch et := e.(type) {
 	case *WebSocketEvent:
-		s.EmitEvent(*et)
+		if err := et.IsValid(); err != nil {
+			log.Errorf("invalid event: %v", err)
```
gx578007, line 207: Is early return missing here?

gx578007 (approved):

PR #1498, opened 2024-01-11
FEATURE: [okx] support query open orders

bailantaotao:


pkg/exchange/okex/okexapi/get_open_orders_request.go
```diff
@@ -0,0 +1,112 @@
...
+	// Leverage, from 0.01 to 125.
+	// Only applicable to MARGIN/FUTURES/SWAP
+	Lever string `json:"lever"`
+	// Profit and loss, Applicable to orders which have a trade and aim to close position. It always is 0 in other conditions
+	Pnl          string `json:"pnl"`
+	PositionSide string `json:"posSide"`
+	// Options price in USDOnly applicable to options; return "" for other instrument types
+	PriceUsd string `json:"pxUsd"`
+	// Implied volatility of the options orderOnly applicable to options; return "" for other instrument types
+	PriceVol string `json:"pxVol"`
```
c9s, line 54: decided not to use fixedpoint.Value?

bailantaotao: Since the field is only used in options, I ignore it. XD

i can update it

pkg/exchange/okex/okexapi/client.go
```diff
@@ -58,6 +58,10 @@ const (
 	OrderStateFilled          OrderState = "filled"
 )
 
+func (o OrderState) IsWorking() bool {
+	return o == OrderStateLive || o == OrderStatePartiallyFilled
```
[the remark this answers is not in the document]
bailantaotao: Why? The function represents only the working status- either live or partially filled.

pkg/exchange/okex/exchange.go
```diff
@@ -295,15 +298,46 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (*t
...
+
+	nextCursor := int64(0)
+	for {
+		if err := queryOpenOrderLimiter.Wait(ctx); err != nil {
+			return nil, fmt.Errorf("query open orders rate limiter wait error: %w", err)
+		}
+
+		req := e.client.NewGetOpenOrdersRequest().
+			InstrumentID(instrumentID).
+			After(fmt.Sprintf("%d", nextCursor))
```
c9s, line 314: better to use strconv.Itoa

strconv.Itoa is faster than fmt.Sprintf

pkg/exchange/okex/okexapi/client.go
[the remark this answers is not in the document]
c9s: like this:
```
switch o {  case OrderStateLive, OrderStatePartiallyFilled: return true;  default: return false  }
```

c9s (approved):

PR #1502, opened 2024-01-16
REFACTOR: [okx] refactor order trade event by json.Unmarshal

bailantaotao:


pkg/exchange/okex/parse.go
```diff
@@ -391,16 +390,67 @@ func parseAccount(v []byte) (*okexapi.Account, error) {
...
+		Price:         o.FillPrice,
+		Quantity:      o.FillSize,
+		QuoteQuantity: o.FillPrice.Mul(o.FillSize),
+		Symbol:        toGlobalSymbol(o.InstrumentID),
+		Side:          side,
+		IsBuyer:       side == types.SideTypeBuy,
+		IsMaker:       o.ExecutionType == okexapi.LiquidityTypeMaker,
+		Time:          types.Time(o.FillTime.Time()),
+		// charged by the platform is positive in our design, so added the `Neg()`.
+		Fee:           o.FillFee.Neg(),
```
bailantaotao, line 450: The fillFee default is currently negative because it is charged by the platform. However, in our design, the value should be positive, so I added the neg.

bailantaotao: If it's a rebate, then the result will be the opposite.

c9s (approved):

bailantaotao:
I pushed the wrong commit. Don't worry, it's the same as the old one.

PR #1515, opened 2024-01-26
[dca2] add dev mode field for dev

kbearXD:


pkg/strategy/dca2/dev_mode.go
```diff
@@ -0,0 +1,7 @@
+package dca2
+
+// DevMode, if Enabled is true. it means it will check the running field
+type DevMode struct {
+	On            bool `json:"on"`
```
c9s, line 5: use `enabled`

pkg/strategy/dca2/dev_mode.go
```diff
@@ -0,0 +1,7 @@
...
+	IsNewStrategy bool `json:"isNewStrategy"`
```
c9s, line 6: how about `newAccount` ?

pkg/strategy/dca2/strategy.go
```diff
@@ -66,6 +66,9 @@ type Strategy struct {
 	// UseCancelAllOrdersApiWhenClose uses a different API to cancel all the orders on the market when closing a grid
 	UseCancelAllOrdersApiWhenClose bool `json:"useCancelAllOrdersApiWhenClose"`
 
+	// dev mode
+	DevMode DevMode `json:"devMode"`
```
c9s, line 70: would be better to use `DevMode *DevMode`

pkg/strategy/dca2/open_position.go
```diff
@@ -32,6 +33,21 @@ func (s *Strategy) placeOpenPositionOrders(ctx context.Context) error {
 
 	s.debugOrders(createdOrders)
 
+	if s.DevMode.On && s.DevMode.IsNewStrategy {
```
c9s, line 36: and check s.DevMode != nil here

pkg/strategy/dca2/dev_mode.go
kbearXD: Actually, it's not new account. This field is used to make dca2 strategy not recover with long time ago trades.

pkg/strategy/dca2/strategy.go
kbearXD: I think if we use pointer here, we can remove Enabled field.

c9s (approved):

PR #1525, opened 2024-02-05
FEATURE: [binance] add margin request

bailantaotao:
1. The go-binance library has been deprecated, so I added a new margin request.
2. We don't need the timestamp because the value is initialized in client.go


gx578007 (approved):

pkg/exchange/binance/binanceapi/place_margin_order_request.go
```diff
@@ -0,0 +1,26 @@
...
+
+	asset string `param:"asset"`
+
+	// TRUE for Isolated Margin, FALSE for Cross Margin, Default FALSE
+	isIsolated bool `param:"isIsolated"`
+
+	// Only for Isolated margin
+	symbol *string `param:"symbol"`
+
+	amount string `param:"amount"`
```
c9s, line 19: use fixedpoint value?

bailantaotao: OK.

The previous type is because I simply copied the value from another file.
The caller uses `req.Price(order.Market.FormatPrice(order.Price))`

PR #1538, opened 2024-02-21
FIX: [retry] fix and improve QueryOrderUntilFilled status check

c9s:


narumiruna (approved):

pkg/exchange/retry/order.go
```diff
@@ -54,15 +53,17 @@ func QueryOrderUntilFilled(
 			OrderID: strconv.FormatUint(orderId, 10),
 		})
 
-		if err2 != nil || o == nil {
+		if err2 != nil {
```
kbearXD, line 56: will o be nil ?

PR #1540, opened 2024-02-22
FEATURE: [dca2] add metrics for dca2

kbearXD:


pkg/strategy/dca2/strategy.go
```diff
@@ -133,6 +138,23 @@ func (s *Strategy) Subscribe(session *bbgo.ExchangeSession) {
...
+
+	if s.Session != nil {
+		labels["exchange"] = s.Session.Name
+	}
+
+	if s.PrometheusLabels == nil {
+		return labels
+	}
+
+	return mergeLabels(s.PrometheusLabels, labels)
```
c9s, line 155: some label methods are from grid2? could we make it shared code?

pkg/strategy/dca2/metrics.go
```diff
@@ -0,0 +1,127 @@
...
+
+	for k, v := range b {
+		labels[k] = v
+	}
+	return labels
+}
+
+func initMetrics(extendedLabels []string) {
+	if metricsState == nil {
+		logrus.Infof("[initMetrics] metricsState")
```
c9s, line 40: can we use log instead of logrus?

pkg/strategy/dca2/state.go
```diff
@@ -63,17 +71,22 @@ func (s *Strategy) emitNextState(nextState State) {
...
+	defer monitorTicker.Stop()
 
 	for {
 		select {
 		case <-ctx.Done():
 			s.logger.Info("[DCA] runState DONE")
 			return
-		case <-ticker.C:
-			s.logger.Infof("[DCA] triggerNextState current state: %d", s.state)
+		case <-stateTriggerTicker.C:
```
c9s, line 85: you used ticker here, is it still back-testable?

c9s (changes requested):
hold this PR, need to take more time to review this

pkg/strategy/dca2/strategy.go
```diff
@@ -146,6 +168,15 @@ func (s *Strategy) Run(ctx context.Context, _ bbgo.OrderExecutor, session *bbgo.
 		s.Position = types.NewPositionFromMarket(s.Market)
 	}
 
+	// prometheus
```
c9s, line 171: add a meaningful comment? like... avoid registering metrics twice and override the previous one?

pkg/strategy/dca2/strategy.go
```diff
@@ -146,6 +168,15 @@ func (s *Strategy) Run(ctx context.Context, _ bbgo.OrderExecutor, session *bbgo.
...
+	if s.PrometheusLabels != nil {
+		initMetrics(labelKeys(s.PrometheusLabels))
+	}
+	registerMetrics()
```
c9s, line 175: I think you need to move this call to the if block?

c9s (approved):

PR #1541, opened 2024-02-23
FEATURE: [rebalance] add price type

narumiruna:


pkg/types/price_type.go
```diff
@@ -0,0 +1,45 @@
+package types
+
+import (
+	"github.com/c9s/bbgo/pkg/fixedpoint"
+)
+
+type PriceType string
```
c9s, line 7: implement unmarshalJSON on PriceType ? and maybe you can transform the original string into upper case string

c9s (approved):

PR #1544, opened 2024-02-23
FIX: [bitget] batch subscribe channel

bailantaotao:


pkg/exchange/bitget/stream.go
```diff
@@ -87,12 +87,22 @@ func (s *Stream) syncSubscriptions(opType WsEventType) error {
...
 
 	logger.Infof("%s channels: %+v", opType, args)
-	if err := s.Conn.WriteJSON(WsOp{
-		Op:   opType,
-		Args: args,
-	}); err != nil {
-		logger.WithError(err).Error("failed to send request")
-		return err
+
+	maximum := 10
```
c9s, line 91: rename this to batchSize?

pkg/exchange/bitget/stream.go
```diff
@@ -87,12 +87,22 @@ func (s *Stream) syncSubscriptions(opType WsEventType) error {
...
+		}
+
+		if err := s.Conn.WriteJSON(WsOp{
+			Op:   opType,
+			Args: args[begin:end],
+		}); err != nil {
+			logger.WithError(err).Error("failed to send request")
+			return err
+		}
 	}
```
c9s, line 106: can we make this part a function?

ycdesu (approved):

pkg/exchange/bitget/stream.go
bailantaotao: We can do it in the next PR

c9s (approved):

PR #1545, opened 2024-02-23
FEATURE: add universal cancel all orders api helper

c9s:


pkg/strategy/dca2/strategy.go
```diff
@@ -289,30 +291,29 @@ func (s *Strategy) CleanUp(ctx context.Context) error {
...
 	var werr error
 	for {
 		s.logger.Infof("checking %s open orders...", s.Symbol)
 
 		openOrders, err := retry.QueryOpenOrdersUntilSuccessful(ctx, session.Exchange, s.Symbol)
 		if err != nil {
-			s.logger.WithError(err).Errorf("CancelOrdersByGroupID api call error")
-			werr = multierr.Append(werr, err)
+			s.logger.WithError(err).Errorf("unable to query open orders")
+			continue
```
gx578007, line 307: Will this hit api rate limit?

pkg/util/tradingutil/cancel.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	for s := range symbolMap {
+		symbols = append(symbols, s)
+	}
+
+	return symbols
+}
+
+func CollectOpenOrders(ctx context.Context, ex types.Exchange, symbols ...string) ([]types.Order, error) {
```
gx578007, line 104: Where does this method get called?

pkg/util/tradingutil/cancel.go
```diff
@@ -0,0 +1,116 @@
...
+				anyErr = err
+			}
+		}
+
+		if anyErr == nil {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("unable to cancel all orders: %+v", openOrders)
```
gx578007, line 73: How about adding `anyErr`?

pkg/strategy/dca2/strategy.go
c9s: I think we have the protection in the exchange api layer

pkg/util/tradingutil/cancel.go
c9s: Sure

pkg/util/tradingutil/cancel.go
c9s: This will be used from xtri

kbearXD (approved):

PR #1550, opened 2024-02-29
FEATURE: [rebalance] add balance type

narumiruna:


pkg/types/balance_type.go
```diff
@@ -0,0 +1,81 @@
...
+	BalanceTypeInterest  BalanceType = "INTEREST"
+	BalanceTypeNet       BalanceType = "NET"
+	BalanceTypeTotal     BalanceType = "TOTAL"
+	BalanceTypeDebt      BalanceType = "DEBT"
+)
+
+var ErrInvalidBalanceType = errors.New("invalid balance type")
+
+func StrToBalanceType(s string) (b BalanceType, err error) {
+	switch strings.ToLower(s) {
```
c9s, line 26: you can use ToUpper? and if the string matches any type vars you defined in the const, you can simply assign it

c9s (approved):

pkg/types/balance_type.go
narumiruna: done

c9s (approved):

PR #1555, opened 2024-03-05
FIX: [ActiveOrderBook] use origin order if error occurred

bailantaotao:
If errors occur in CancelOrders and QueryOpenOrders, it will result in an infinite loop because the orders have been replaced with an empty slice.


c9s (approved):
thanks!!!

PR #1575, opened 2024-03-12
feature: add ExchangePublic

zenixls2:
This is for exporting bbgo as a library when we want to use ExchangeMarketService + Minimal for a Exchange with only public api enabled (without api keys)

Right now if we pass in empty api key and passphrase, a public api only exchange instance will be created. We also need to reuse some of the utilities inside bbgo to avoid rebuilding the wheels.


zenixls2:
instead of using PublicExchange, we follow the naming convention of Exchange interfaces here to name this new interface as ExchangePublic (Exchange + adj)

PR #1578, opened 2024-03-13
FIX: [okx] remove the query after place order

bailantaotao:


c9s (approved):

pkg/exchange/okex/exchange.go
```diff
@@ -269,16 +269,22 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (*t
...
+	orderID, err := strconv.ParseUint(orders[0].OrderID, 10, 64)
 	if err != nil {
-		return nil, fmt.Errorf("failed to query order by id: %s, clientOrderId: %s, err: %w", orders[0].OrderID, orders[0].ClientOrderID, err)
-	}
-
-	return orderRes, nil
+		return nil, fmt.Errorf("failed to parse response order id: %w", err)
+	}
+
+	timeNow := time.Now()
```
ycdesu, line 277: We specify the time after the network operation, which can take seconds or longer. Running time.Now() upon entering this method would yield a more accurate timestamp

ycdesu (approved):

PR #1580, opened 2024-03-14
FIX: [dca2] must calculate and emit profit at the end of the round

kbearXD:


pkg/strategy/dca2/strategy.go
```diff
@@ -381,6 +385,26 @@ func (s *Strategy) CalculateAndEmitProfit(ctx context.Context) error {
 		return fmt.Errorf("exchange %s doesn't support ExchangeOrderQueryService", s.ExchangeSession.Exchange.Name())
 	}
 
+	maxTry := 10
+	for try := 1; try < maxTry; try++ {
```
ycdesu, line 389: It retries only 9 times

ycdesu (approved):

pkg/strategy/dca2/state.go
```diff
@@ -201,7 +201,7 @@ func (s *Strategy) runTakeProfitReady(ctx context.Context, next State) {
 	// reset position
 
 	// calculate profit stats
-	if err := s.CalculateAndEmitProfit(ctx); err != nil {
+	if err := s.mustCalculateAndEmitProfit(ctx); err != nil {
```
c9s, line 204: when we use "must" it means the function may panic.

pkg/strategy/dca2/strategy.go
c9s: can use backoff here?

c9s (approved):

pkg/strategy/dca2/strategy.go
```diff
@@ -381,6 +386,22 @@ func (s *Strategy) CalculateAndEmitProfit(ctx context.Context) error {
...
+		}
+
+		if s.ProfitStats.FromOrderID == fromOrderID {
+			return fmt.Errorf("FromOrderID (%d) is not updated, retry it", s.ProfitStats.FromOrderID)
+		}
+
+		return nil
+	}
+
+	return retry.GeneralLiteBackoff(ctx, op)
```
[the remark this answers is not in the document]
kbearXD: I will print all the rounds and its orders when calculating.

PR #1583, opened 2024-03-14
FEATURE: [okx] query recent trades

bailantaotao:
The historical trade API has a high rate limiter cost (10/2s), 

so I have introduced a new recent trade API to address this. If the start time is less than or equal to 3 days, we will utilize the recent trade API.


pkg/exchange/okex/exchange.go
```diff
@@ -55,6 +55,7 @@ const (
 	defaultQueryLimit = 100
 
 	maxHistoricalDataQueryPeriod = 90 * 24 * time.Hour
+	threeDaysHistoricalPeriod    = 3 * 24 * time.Hour
```
c9s, line 58: use maxTradeHistoryPeriod ?

pkg/exchange/okex/exchange.go
```diff
@@ -564,33 +567,31 @@ func (e *Exchange) QueryTrades(ctx context.Context, symbol string, options *type
...
-
 	limit := options.Limit
 	if limit > defaultQueryLimit || limit <= 0 {
 		log.Infof("limit is exceeded default limit %d or zero, got: %d, use default limit", defaultQueryLimit, limit)
 		limit = defaultQueryLimit
 	}
-	req.Limit(uint64(limit))
 
-	var newStartTime time.Time
+	timeNow := e.timeNowFunc()
```
c9s, line 576: can be just `now` ?

pkg/exchange/okex/okexapi/get_transaction_history_request.go
```diff
@@ -53,13 +53,13 @@ type Trade struct {
 	PosSide          string           `json:"posSide"`
 }
 
-//go:generate GetRequest -url "/api/v5/trade/fills-history" -type GetTransactionHistoryRequest -responseDataType []Trade
+//go:generate GetRequest -url "/api/v5/trade/fills-history" -type GetTransactionHistoryRequest -responseDataType []Trade -rateLimiter 1+10/2s
```
c9s, line 56: sooo cool!!!

pkg/exchange/okex/exchange.go
bailantaotao: No, by using that `timeNowFunc()`, we can test the function.

pkg/exchange/okex/exchange.go
bailantaotao: We already have one that supports 90 days, called `maxHistoricalDataQueryPeriod`. 

And now we have another one that's only 3 days, so it's called `threeDaysHistoricalPeriod` XD.

PR #1597, opened 2024-03-20
FEATURE: close short positions with ClosePosition

Newtoniano:
This PR adds the ability to close short positions with the ClosePosition method. It adjusts the order quantity based on the available quote balance and the current ask price for the symbol.


c9s (approved):
awesome! you've learned the api so fast!

PR #1601, opened 2024-03-22
add vscode launch setting for easier run / debug

lanphan:


.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
c9s, line 4: is this a valid json comment?

lanphan: json doesn't have comment, but it's a launch file in vscode and it accepts that comment.
In fact, the whole file is generated by vscode when I create new launch file, then I update / add some more configs

c9s: got it, thanks!

c9s (approved):

Issue #1037, opened 2022-12-30
get-order command error


liaochy:
executing get-ordercommand
`../bin/bbgo-slim get-order --order-id 16863344284 --session=binance --symbol=BTCUSDT --dotenv .env.binance.btcusdt`

Result
`FATAL cannot execute command error=<APIError> code=-1105, msg=Parameter 'symbol' was empty.`

cmd/orders.go service.QueryOrder should add symbol to Parameter list

c9s:
looks like they changed the API

lanphan:
this is fixed, you can close it, @c9s

PR #1606, opened 2024-03-27
FIX: issue #1037, get-order command error

lanphan:
fix issue https://github.com/c9s/bbgo/issues/1037

### DIFF
--- a/pkg/cmd/orders.go
+++ b/pkg/cmd/orders.go
@@ -47,6 +47,11 @@ var getOrderCmd = &cobra.Command{
 
 		orderID, err := cmd.Flags().GetString("order-id")
 		if err != nil {
+			return fmt.Errorf("can't get the order-id from flags: %w", err)
+		}
+
+		symbol, err := cmd.Flags().GetString("symbol")
+		if err != nil {
 			return fmt.Errorf("can't get the symbol from flags: %w", err)
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/orders.go
```diff
@@ -46,9 +46,17 @@ var getOrderCmd = &cobra.Command{
...
+		if err != nil {
+			return fmt.Errorf("can't get the order-id from flags: %w", err)
+		}
+
+		symbol, err := cmd.Flags().GetString("symbol")
 		if err != nil {
 			return fmt.Errorf("can't get the symbol from flags: %w", err)
 		}
+		if symbol == "" {
+			return fmt.Errorf("symbol is not found")
```
c9s, line 58: use `symbol is required` ?

c9s (approved):

pkg/cmd/orders.go
lanphan: yes, error returned from binance before fix is

```
FATAL cannot execute command error=<APIError> code=-1105, msg=Parameter 'symbol' was empty.
```

c9s: yeah I mean let's use the error message "symbol is required"

lanphan: symbol is required param in viber, so that I remove that check, both in get-order and list-orders command

PR #1608, opened 2024-03-27
FIX: [xalign] fix buy side max amount

narumiruna:


pkg/strategy/xalign/strategy.go
```diff
@@ -288,6 +284,8 @@ func (s *Strategy) selectSessionForCurrency(
 						Market:      market,
 						TimeInForce: types.TimeInForceGTC,
 					}
+				} else {
+					log.Infof("The amount %f is not greater than the minimal order quantity for %s", q.Float64(), market.Symbol)
```
c9s, line 288: use Warnf ?

narumiruna: done

c9s (approved):

PR #1609, opened 2024-03-27
FEATURE: emit position when position updated and reset

kbearXD:


pkg/strategy/dca2/strategy.go
```diff
@@ -416,6 +419,14 @@ func (s *Strategy) ContinueNextRound() {
 	s.nextRoundPaused = false
 }
 
+func (s *Strategy) GetTakeProfitPrice() fixedpoint.Value {
+	if s.Position.Base == 0 {
```
c9s, line 423: should not compare with 0

gx578007 (approved):

PR #1618, opened 2024-04-16
CHORE: [atrpin] add submitting log

narumiruna:


pkg/strategy/atrpin/strategy.go
```diff
@@ -199,21 +200,3 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
 
 	return nil
 }
-
-func logErr(err error, msgAndArgs ...interface{}) bool {
```
c9s, line 203: i think we need this func, maybe move this to util?

narumiruna: OK, I will drop the commit and create another PR to move `logErr`.

PR #1621, opened 2024-04-19
FEATURE: [dca2] add position callback

kbearXD:


pkg/strategy/dca2/strategy.go
```diff
@@ -100,7 +100,8 @@ type Strategy struct {
 
 	// callbacks
 	common.StatusCallbacks
-	profitCallbacks []func(*ProfitStats)
+	profitCallbacks   []func(*ProfitStats)
+	positionCallbacks []func(*types.Position)
```
c9s, line 104: positionUpdateCallbacks ?

kbearXD: ok!

c9s (approved):

PR #1638, opened 2024-05-20
FEATURE: [dca2] store price quantity pairs of the open-position order…

kbearXD:
…s into persistence


pkg/strategy/dca2/profit_stats.go
```diff
@@ -8,6 +8,11 @@ import (
 	"github.com/c9s/bbgo/pkg/types"
 )
 
+type PriceQuantity struct {
```
c9s, line 11: you can reuse the types.PriceVolume struct ?

PR #1646, opened 2024-05-31
MINOR: add inserted_at column to trades

ycdesu:
A trade may be missed initially and fetched after it has occurred, so we add inserted_at column.

### DIFF
--- a/migrations/mysql/20240531163411_trades_created.sql
+++ b/migrations/mysql/20240531163411_trades_created.sql
@@ -1,0 +1,14 @@
+-- +up
+-- +begin
+ALTER TABLE `trades` ADD COLUMN `inserted_at` DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL AFTER `traded_at`;
+-- +end
+
+-- +begin
+UPDATE `trades` SET `inserted_at`=`traded_at`;
+-- +end
+
+-- +down
+
+-- +begin
+ALTER TABLE `trades` DROP COLUMN `inserted_at`;
+-- +end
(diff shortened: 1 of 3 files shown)


migrations/mysql/20240531163411_trades_created.sql
```diff
@@ -0,0 +1,11 @@
+-- +up
+-- +begin
+ALTER TABLE `trades` ADD COLUMN `inserted_at` TIMESTAMP DEFAULT CURRENT_TIMESTAMP NOT NULL;
```
c9s, line 3: can we use DATETIME(3) ?

pkg/types/trade.go
```diff
@@ -95,6 +95,8 @@ type Trade struct {
 
 	// PnL is the profit and loss value of the executed trade
 	PnL sql.NullFloat64 `json:"pnl" db:"pnl"`
+
+	InsertedAt time.Time `db:"inserted_at"`
```
c9s, line 99: you can use types.Time here, it handles the datetime scan from db

c9s (approved):

pkg/types/trade.go
```diff
@@ -95,6 +95,8 @@ type Trade struct {
 
 	// PnL is the profit and loss value of the executed trade
 	PnL sql.NullFloat64 `json:"pnl" db:"pnl"`
+
+	InsertedAt Time `db:"inserted_at"`
```
c9s, line 99: also add a json tag here?

ycdesu: @c9s It's not returned by the server json. The value is generated by mysql. Do we still have to add json tag?

c9s: yeah we have some api renders the trade object in JSON format, but it's optional, not a problem

c9s (approved):

c9s:
好像還是 test fail 可能要 rebase 一下 我猜可能是 bitget test failed

bailantaotao (approved):
應該要rebase main

c9s (approved):

c9s:
@ycdesu you need to sign CLA~ thanks

PR #1648, opened 2024-06-02
FEATURE: [atrpin] take profit by expected base balance

narumiruna:


pkg/strategy/atrpin/strategy.go
```diff
@@ -133,25 +154,34 @@ func (s *Strategy) Run(ctx context.Context, orderExecutor bbgo.OrderExecutor, se
...
+			s.logger.Infof("position: %+v", position)
+			baseQuantity = position.Base
+		}
+
+		side := types.SideTypeSell
+		takerPrice := ticker.Buy
+		if baseQuantity.Sign() < 0 {
+			side = types.SideTypeBuy
+			takerPrice = ticker.Sell
+			baseQuantity = baseQuantity.Abs()
```
c9s, line 169: I think you can check the ClosePosition() method, and maybe we can reuse some of them

narumiruna: If there is a missing trade, then the position base will be incorrect. Thus I'd like to add the `ExpectedBaseBalance` to solve the problem.

c9s: can we use query trades to fix the position?

the balance balance could be fixed by xalign, and we will need a correct PnL record

narumiruna: we can't fix the balance by xalign, because atrpin will place order

narumiruna: how to query trades?

c9s (approved):

PR #1658, opened 2024-06-18
REFACTOR: countdown util

narumiruna:


pkg/util/countdown.go
```diff
@@ -0,0 +1,24 @@
+package util
+
+import (
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type Countdown struct {
```
c9s, line 9: DailyDataTracker or PeriodicDataTracker

PR #1663, opened 2024-06-25
FIX: [rebalance] round down quantity

narumiruna:


pkg/strategy/rebalance/strategy.go
```diff
@@ -255,9 +256,9 @@ func (s *Strategy) generateOrder(ctx context.Context) (*types.SubmitOrder, error
 		}
 
 		if side == types.SideTypeBuy {
-			quantity = fixedpoint.Min(quantity, balances[s.QuoteCurrency].Available.Div(ticker.Sell))
+			quantity = fixedpoint.Min(quantity, balances[s.QuoteCurrency].Available.Div(ticker.Sell)).Round(market.VolumePrecision, fixedpoint.Down)
```
c9s, line 259: I think you can use:

market.RoundDownQuantityByPrecision

and 

market.RoundUpQuantityByPrecision

c9s (approved):

pkg/strategy/rebalance/strategy.go
narumiruna: done

PR #1667, opened 2024-07-08
FIX: [common] fix profit fixer batch query

c9s:


pkg/strategy/common/profit_fixer.go
```diff
@@ -58,24 +58,37 @@ func (f *ProfitFixer) aggregateAllTrades(ctx context.Context, symbol string, sin
...
+				}
+			}
 		})
 	}
 
 	if err := g.Wait(); err != nil {
 		return nil, err
 	}
 
+	mu.Lock()
```
gx578007, line 88: Is this lock required?

gx578007 (approved):

andycheng123 (approved):

pkg/strategy/common/profit_fixer.go
c9s: not sure, but I want to make it safe

PR #1668, opened 2024-07-09
REFACTOR: support custom order by column

ycdesu:
Now user is able to specify the ordering column.

### DIFF
--- a/pkg/service/trade.go
+++ b/pkg/service/trade.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"context"
+	"fmt"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/service/trade.go
```diff
@@ -304,11 +309,24 @@ func (s *TradeService) Query(options QueryTradesOptions) ([]types.Trade, error)
 		sel = sel.Where(sq.Eq{"exchange": options.Sessions})
 	}
 
+	orderByColumn := "traded_at"
+	if options.OrderByColumn != "" {
+		if options.OrderByColumn != "traded_at" && options.OrderByColumn != "gid" {
```
c9s, line 314: using switch case might be better here?

pkg/service/trade.go
```diff
@@ -304,11 +309,24 @@ func (s *TradeService) Query(options QueryTradesOptions) ([]types.Trade, error)
...
+
+		orderByColumn = options.OrderByColumn
+	}
+
+	ordering := "ASC"
 	if options.Ordering != "" {
-		sel = sel.OrderBy("traded_at " + options.Ordering)
-	} else {
-		sel = sel.OrderBy("traded_at ASC")
+		if options.Ordering != "ASC" && options.Ordering != "DESC" {
```
c9s, line 323: ditto

c9s (approved):

c9s (approved):

PR #1678, opened 2024-07-31
FEATURE: [xalign] add withdraw detection

c9s:


pkg/exchange/max/maxapi/account.go
```diff
@@ -155,7 +155,23 @@ type WithdrawState string
 
 const (
 	WithdrawStateSubmitting WithdrawState = "submitting"
+	WithdrawStateSubmitted  WithdrawState = "submitted"
 	WithdrawStateConfirmed  WithdrawState = "confirmed"
+	WithdrawStatePending    WithdrawState = "pending"
+	WithdrawStateProcessing WithdrawState = "processing"
+	WithdrawStateCanceled   WithdrawState = "canceled"
```
gx578007, line 162: Is this keyword defined by Max because it lacks an "l" and is different from another definition below?

pkg/strategy/xalign/strategy.go
```diff
@@ -391,6 +419,14 @@ func (s *Strategy) align(ctx context.Context, sessions map[string]*bbgo.Exchange
 		}
 	}
 
+	foundActiveTransfer, err := s.detectActiveTransfers(ctx, sessions)
+	if err != nil {
+		log.WithError(err).Errorf("unable to check active transfers")
```
gx578007, line 424: What do you think skipping aligning if any error happens?

gx578007 (approved):

pkg/exchange/max/maxapi/account.go
c9s: Yes they are max only

PR #1700, opened 2024-08-21
Fix: [autobuy] fix error when bollinger settings is not set

narumiruna:


c9s (approved):

pkg/strategy/autobuy/strategy.go
```diff
@@ -64,27 +65,57 @@ func (s *Strategy) InstanceID() string {
...
+
+		if s.Bollinger.BandWidth <= 0 {
+			return fmt.Errorf("bollinger band width must be greater than 0")
+		}
+	}
 	return nil
 }
 
 func (s *Strategy) Defaults() error {
+	if s.OrderType == "" {
```
c9s, line 97: order type needs to be aligned with the price type

if price type = taker, the default limit maker won’t work

narumiruna: so add a warning here?

c9s: the default order type should be "limit", because it would work for all the cases

narumiruna: I will set the default order type to limit

PR #1727, opened 2024-09-02
FIX: update timeInForce for binance margin order

lanphan:
update binance margin order to support IOC and FOK, same behavior with spot order

### DIFF
--- a/pkg/exchange/binance/exchange.go
+++ b/pkg/exchange/binance/exchange.go
@@ -1005,16 +1005,13 @@ func (e *Exchange) submitMarginOrder(ctx context.Context, order types.SubmitOrde
 		}
 	}
 
-	// could be IOC or FOK
-	switch order.Type {
-	case types.OrderTypeLimit, types.OrderTypeStopLimit:
-		req.TimeInForce(binance.TimeInForceTypeGTC)
-	case types.OrderTypeLimitMaker:
-		// do not set TimeInForce for LimitMaker
-	default:
-		if len(order.TimeInForce) > 0 {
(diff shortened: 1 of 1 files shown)


pkg/exchange/binance/exchange.go
```diff
@@ -1005,16 +1005,13 @@ func (e *Exchange) submitMarginOrder(ctx context.Context, order types.SubmitOrde
 		}
 	}
 
-	// could be IOC or FOK
-	switch order.Type {
-	case types.OrderTypeLimit, types.OrderTypeStopLimit:
-		req.TimeInForce(binance.TimeInForceTypeGTC)
-	case types.OrderTypeLimitMaker:
```
c9s, line 1012: keep this condition

pkg/exchange/binance/exchange.go
```diff
@@ -1005,16 +1005,13 @@ func (e *Exchange) submitMarginOrder(ctx context.Context, order types.SubmitOrde
 		}
 	}
 
-	// could be IOC or FOK
-	switch order.Type {
-	case types.OrderTypeLimit, types.OrderTypeStopLimit:
```
c9s, line 1010: remove the limit order type from the case, and keep the stop limit type in the case

lanphan:
@c9s 
I already thought about that, but I keep using same condition with spot because:
1. All spot orders are now ok (noone complains)
2. Document from Binance applies same condition for both spot + margin

Therefore, I copy condition from spot to margin here. In case we want to improve later, we can do it for both

And, I already executed it, it works well now

PR #1728, opened 2024-09-03
FIX: [core] fix memory leak

kbearXD:


pkg/bbgo/marketdatastore.go
```diff
@@ -5,6 +5,9 @@ import "github.com/c9s/bbgo/pkg/types"
 const MaxNumOfKLines = 5_000
 const MaxNumOfKLinesTruncate = 100
 
+const NumberOfKLinesThreshold = 6_000
```
c9s, line 8: we already have MaxNumOfKLines ?

pkg/bbgo/marketdatastore.go
```diff
@@ -5,6 +5,9 @@ import "github.com/c9s/bbgo/pkg/types"
...
+const NumberOfTruncatedKLines = 5_000
```
c9s, line 9: what's the difference with MaxNumOfKLinesTruncate?

pkg/bbgo/marketdatastore.go
```diff
@@ -57,8 +60,13 @@ func (store *MarketDataStore) AddKLine(k types.KLine) {
 	}
 	window.Add(k)
 
-	if len(*window) > MaxNumOfKLines {
-		*window = (*window)[MaxNumOfKLinesTruncate-1:]
+	LenOfWindow := len(*window)
+	CapOfWindow := cap(*window)
```
c9s, line 64: can we fix the variable name case?

c9s (approved):

pkg/bbgo/marketdatastore.go
```diff
@@ -57,8 +59,13 @@ func (store *MarketDataStore) AddKLine(k types.KLine) {
...
 
-	if len(*window) > MaxNumOfKLines {
-		*window = (*window)[MaxNumOfKLinesTruncate-1:]
+	lenOfWindow := len(*window)
+	capOfWindow := cap(*window)
+
+	if lenOfWindow == capOfWindow && capOfWindow > CapacityOfKLineWindowLimit {
+		truncatedWindow := (*window)[(CapacityOfKLineWindowLimit / 2):]
+		copy(*window, truncatedWindow)
+		*window = (*window)[:len(truncatedWindow)]
```
c9s, line 68: can you add a test case for this?

PR #1743, opened 2024-09-13
add new tag disableConfigPrint to prevent printing specific field

lanphan:


pkg/dynamic/print_config.go
```diff
@@ -91,6 +91,10 @@ func PrintConfig(s interface{}, f io.Writer, style *table.Style, withColor bool,
 						continue
 					}
 					if jtag := tt.Tag.Get("json"); jtag != "" && jtag != "-" {
+						disableConfigPrint := tt.Tag.Get("disableConfigPrint")
```
c9s, line 94: I thought you meant the strategy parameter, for this case, i think we can name it "hidden" or "unprintable"

lanphan: ok @c9s, let's use "unprintable" just because we already have "modifiable"

lanphan:
@c9s 
you want to change field name to ignore: true to be consistent with printdump as Railgun said? He already has PR to add documentation for that

lanphan:
@c9s 
I update tag name to "ignore", so that it's consistent with document https://github.com/c9s/bbgo/blob/main/doc/topics/rw-strategy.md

c9s (approved):
looks good, thanks!

PR #1753, opened 2024-09-25
IMPROVE: [xdepthmaker] use order query to update the canceled order, fix depth price, fix symbol column lengths, fix covered position

c9s:


kbearXD (approved):

migrations/sqlite3/20240925160534_fix_symbol_length2.sql
```diff
@@ -0,0 +1,10 @@
+-- +up
+-- +begin
+SELECT 1;
```
gx578007, line 3: This migration file could be dropped because nothing is changed?

c9s: sqlite does not support column modification

PR #1760, opened 2024-10-01
FEATURE: [max] add QueryDepth v3 API to query orderbook

kbearXD:


pkg/exchange/max/exchange.go
```diff
@@ -1134,6 +1134,36 @@ func (e *Exchange) QueryKLines(
...
+
+	depth, err := req.Do(ctx)
+	if err != nil {
+		return snapshot, finalUpdateID, err
+	}
+
+	return convertDepth(symbol, depth)
+}
+
+func convertDepth(symbol string, depth *v3.Depth) (snapshot types.SliceOrderBook, finalUpdateID int64, err error) {
```
c9s, line 1150: move this to convert.go ?

pkg/exchange/max/exchange.go
```diff
@@ -1134,6 +1134,36 @@ func (e *Exchange) QueryKLines(
 	return kLines, nil
 }
 
+func (e *Exchange) QueryDepth(ctx context.Context, symbol string, limit int) (snapshot types.SliceOrderBook, finalUpdateID int64, err error) {
+	req := e.v3client.NewGetDepthRequest()
+	req.Market(symbol)
```
c9s, line 1139: should we sort by price?

c9s (approved):

pkg/exchange/max/exchange.go
kbearXD: default it's true, I think both options are fine ?

PR #1780, opened 2024-10-16
FEATURE: [xmaker] support delayed hedge

c9s:


ycdesu (approved):

pkg/strategy/xmaker/strategy.go
```diff
@@ -1027,7 +1064,7 @@ func (s *Strategy) tryArbitrage(ctx context.Context, quote *Quote, makerBalances
 			return false, nil
 		}
 
-		if qty.IsZero() {
+		if qty.IsZero() || s.makerMarket.IsDustQuantity(qty, sumPv.Price) {
```
gx578007, line 1067: Does `qty.IsZero()` become redundant? Or `qty.IsZero()` is a faster route for most cases?

gx578007 (approved):

PR #1783, opened 2024-10-17
FEATURE: update binance api positionRisk to v3

anywhy:


pkg/exchange/binance/binanceapi/futures_get_position_risks_request.go
```diff
@@ -8,24 +8,30 @@ import (
...
+	MarkPrice              fixedpoint.Value `json:"markPrice"`
+	UnRealizedProfit       fixedpoint.Value `json:"unRealizedProfit"`
+	LiquidationPrice       fixedpoint.Value `json:"liquidationPrice"`
+	IsolatedMargin         string           `json:"isolatedMargin"`
+	Notional               fixedpoint.Value `json:"notional"`
+	MarginAsset            fixedpoint.Value `json:"marginAsset"`
+	IsolatedWallet         string           `json:"isolatedWallet"`
+	InitialMargin          fixedpoint.Value `json:"initialMargin"`
+	MaintMargin            fixedpoint.Value `json:"maintMargin"`
+	PositionInitialMargin  fixedpoint.Value `json:"positionInitialMargin"`
```
c9s, line 25: can you make it aligned with the original field, so the diff could be more clean

the field name matters, so please check the field name is matched with the original name

c9s:
build failed

c9s:
@anywhy can you avoid the merge point? thanks, better to use git rebase rather than "merge"

pkg/exchange/binance/binanceapi/futures_get_position_risks_request.go
```diff
@@ -8,24 +8,30 @@ import (
...
```
c9s, line 25: @anywhy please don't change the order of the field, it is hard to review the difference

c9s (changes requested):
please put the fields in the original order, thanks

PR #1786, opened 2024-10-21
FEATURE: [xalign] detect active depoit. if found, skip align

kbearXD:


pkg/strategy/xalign/strategy.go
```diff
@@ -156,6 +156,38 @@ func (s *Strategy) detectActiveWithdraw(
 	return nil, err2
 }
 
+func (s *Strategy) detectActiveDeposit(
+	ctx context.Context,
+	sessions map[string]*bbgo.ExchangeSession,
+) (*types.Deposit, error) {
+	var err2 error
+	until := time.Now()
+	since := until.Add(-time.Hour * 24)
```
c9s, line 165: using 1 hour should be enough?

kbearXD: I use the same interval as detectActiveWithdraw. I think both it's ok because there will not be too many withdraws and deposits ?

PR #1791, opened 2024-10-24
FEATURE: [xalign] add notification when order quote is over alert amount

kbearXD:


config/xalign.yaml
```diff
@@ -47,3 +47,7 @@ crossExchangeStrategies:
       USDT: 100
       USDC: 100
       TWD: 3000
+    alertAmounts:
+      USDT: 200
+      USDC: 200
+      TWD: 6000
```
c9s, line 53: we need to add a quoteCurrency config so that we can convert BTC, ETH into the quote currency

c9s: and use a single field like `slackAlertQuoteAmount` as the threshold

c9s, line 53: we also need a field named like `slackAlertMentions`

c9s: Check pkg/strategy/autoborrow/strategy.go for example.

pkg/strategy/xalign/strategy.go
```diff
@@ -50,6 +100,7 @@ type Strategy struct {
 	BalanceToleranceRange    fixedpoint.Value            `json:"balanceToleranceRange"`
 	Duration                 types.Duration              `json:"for"`
 	MaxAmounts               map[string]fixedpoint.Value `json:"maxAmounts"`
+	MaxAmountAlert           *MaxAmountAlertConfig       `json:"maxAmountAlert"`
```
c9s, line 103: call it LargeAmountAlert?

pkg/strategy/xalign/strategy.go
```diff
@@ -39,6 +40,55 @@ type QuoteCurrencyPreference struct {
...
+	Side         types.SideType
+	Price        fixedpoint.Value
+	Quantity     fixedpoint.Value
+	Amount       fixedpoint.Value
+}
+
+func (m *MaxAmountAlert) SlackAttachment() slack.Attachment {
+	return slack.Attachment{
+		Color: "red",
+		Title: fmt.Sprintf("xalign amount alert - try to align %s with quote %s amount %f > %f",
```
c9s, line 64: Large Amount Alert

pkg/strategy/xalign/strategy.go
```diff
@@ -50,6 +100,7 @@ type Strategy struct {
 	BalanceToleranceRange    fixedpoint.Value            `json:"balanceToleranceRange"`
 	Duration                 types.Duration              `json:"for"`
 	MaxAmounts               map[string]fixedpoint.Value `json:"maxAmounts"`
```
c9s, line 102: This field can be removed?

c9s (approved):

PR #1796, opened 2024-10-28
IMPROVE: [liqmaker] improve interval checking and tickers

c9s:


ycdesu (approved):

pkg/strategy/liquiditymaker/strategy.go
```diff
@@ -135,6 +136,22 @@ func (s *Strategy) Defaults() error {
 	return nil
 }
 
+func (s *Strategy) liquidityWorker(ctx context.Context, interval types.Interval) {
+	ticker := time.NewTicker(interval.Duration())
+	defer ticker.Stop()
+
+	s.placeLiquidityOrders(ctx)
```
bailantaotao, line 143: place it twice?

c9s: yes before the ticker

c9s: or we will need to wait for 1 hour

PR #1804, opened 2024-11-05
FEATURE: add livenote support

c9s:


pkg/bbgo/livenote.go
```diff
@@ -0,0 +1,26 @@
...
+// The MessageID will be set after the message is posted if it's not set.
+func PostLiveNote(obj livenote.Object) {
+	if len(Notification.liveNotePosters) == 0 {
+		logrus.Warn("no live note poster is registered")
+		return
+	}
+
+	for _, poster := range Notification.liveNotePosters {
+		if err := poster.PostLiveNote(obj); err != nil {
+			logrus.WithError(err).Errorf("unable to post live note: %+v", obj)
```
gx578007, line 19: Will it cause too many error messages?

pkg/notifier/slacknotifier/slack.go
```diff
@@ -66,7 +71,50 @@ func (n *Notifier) worker() {
 	}
 }
 
+func (n *Notifier) PostLiveNote(obj livenote.Object) error {
```
gx578007, line 74: How about making ctx as one of the input parameters?

pkg/bbgo/livenote.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"github.com/sirupsen/logrus"
+
+	"github.com/c9s/bbgo/pkg/livenote"
+)
+
+// PostLiveNote posts a live note to slack or other services
+// The MessageID will be set after the message is posted if it's not set.
+func PostLiveNote(obj livenote.Object) {
+	if len(Notification.liveNotePosters) == 0 {
```
gx578007, line 12: Could Notification be nil?

gx578007 (approved):

PR #1812, opened 2024-11-13
FEAUTRE: [max] reset book if receive both empty asks and bids

kbearXD:


pkg/exchange/max/stream.go
```diff
@@ -245,6 +235,22 @@ func (s *Stream) handleBookEvent(ex *Exchange) func(e max.BookEvent) {
 			})
 			s.depthBuffers[symbol] = f
 		}
+
+		// if we receive orderbook event with both asks and bids are empty, it means we need to rebuild this orderbook
+		needToResetBook := len(e.Asks) == 0 && len(e.Bids) == 0
```
c9s, line 240: namenit "shouldReset" ?

c9s (approved):

PR #1817, opened 2024-11-14
FIX: refactor and redesign connectivity

c9s:


pkg/types/connectivitygroup.go
```diff
@@ -0,0 +1,226 @@
...
+
+const (
+	ConnectivityStateDisconnected ConnectivityState = -1
+	ConnectivityStateUnknown      ConnectivityState = 0
+	ConnectivityStateConnected    ConnectivityState = 1
+	ConnectivityStateAuthed       ConnectivityState = 2
+)
+
+func getConnState(con *Connectivity) ConnectivityState {
+	state := ConnectivityStateDisconnected
```
ycdesu, line 34: Should the default value be Unknown and add a disconnected else case?

c9s: i later figured out that we can start with disconnected state

pkg/types/connectivitygroup.go
```diff
@@ -0,0 +1,226 @@
...
+		select {
+		case <-ctx.Done():
+			return false
+
+		case <-conn.connectedC:
+			continue
+
+		case <-conn.disconnectedC:
+			return true
+		}
```
gx578007, line 192: Is it possible to be blocked here forever? Any timeout required?

pkg/types/connectivitygroup.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (g *ConnectivityGroup) Add(con *Connectivity) {
+	g.mu.Lock()
+	g.connections = append(g.connections, con)
+	g.states[con] = getConnState(con)
+	g.sumState = sumStates(g.states)
+	g.setFlags(g.sumState)
+	g.mu.Unlock()
+
+	_con := con
```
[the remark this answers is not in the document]
c9s: copy 是故意留給 closure 用，我怕直接用 func argument 會有 side effect

pkg/types/connectivitygroup.go
c9s: i need to remove the AnyDisconnected(), channel 會有 triggered once 的問題

gx578007 (approved):

narumiruna (approved):

PR #1828, opened 2024-11-16
FIX: [depth] fix early snapshot id checking

c9s:


pkg/depth/buffer.go
```diff
@@ -151,7 +147,7 @@ func (b *Buffer) fetchAndPush() error {
 
 	if len(b.buffer) > 0 {
 		// the snapshot is too early
-		if finalUpdateID < b.buffer[0].FirstUpdateID {
+		if finalUpdateID < b.buffer[0].FirstUpdateID-1 {
```
c9s, line 150: ☝️ main fix

kbearXD (approved):

PR #1847, opened 2024-12-02
REFACTOR: [xalign] refactor detector

c9s:


pkg/strategy/xalign/detector/deviation.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (d *DeviationDetector[T]) ShouldFix() (bool, time.Duration) {
+	if len(d.records) == 0 {
+		return false, 0
+	}
+
+	last := d.records[len(d.records)-1]
+	firstRecord := d.records[0]
+	sustainedDuration := last.Time.Sub(firstRecord.Time)
```
gx578007, line 98: What if there's only one record, does it matter?

c9s: doesn't matter~

gx578007 (approved):

PR #1857, opened 2024-12-10
FEATURE: [max] unlock depth buffer after emit

kbearXD:


pkg/depth/buffer.go
```diff
@@ -91,16 +92,16 @@ func (b *Buffer) SetSnapshot(snapshot types.SliceOrderBook, firstUpdateID int64,
 		return nil
 	}
 
+	log.Info("setting the snapshot")
 	// set the final update ID so that we will know if there is an update missing
 	b.finalUpdateID = finalUpdateID
 
 	// set the snapshot
 	b.snapshot = &snapshot
+	b.EmitReady(snapshot, nil)
```
c9s, line 101: could be dangerous here, dead lock can happen

c9s (approved):

PR #1876, opened 2025-01-03
FEATURE: add isolation forest algorithm

narumiruna:
https://cs.nju.edu.cn/zhouzh/zhouzh.files/publication/icdm08b.pdf

### DIFF
--- a/pkg/ensemble/iforest/forest.go
+++ b/pkg/ensemble/iforest/forest.go
@@ -1,0 +1,186 @@
+package iforest
+
+import (
+	"math"
+	"math/rand"
+	"sync"
+)
+
+const (
+	defaultNumTrees       = 100
+	defaultSampleSize     = 256
+	defaultScoreThreshold = 0.6
+	defaultDetectionType  = DetectionTypeThreshold
+	offset                = 0.5
+)
+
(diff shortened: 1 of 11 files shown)


pkg/ensemble/iforest/forest.go
```diff
@@ -0,0 +1,175 @@
...
+	Threshold     float64       `json:"threshold"`
+	Proportion    float64       `json:"proportion"`
+	NumTrees      int           `json:"numTrees"`
+	SampleSize    int           `json:"sampleSize"`
+	MaxDepth      int           `json:"maxDepth"`
+}
+
+// SetDefaultValues initializes zero-valued fields in Options with defaults.
+func (o *Options) SetDefaultValues() {
+	if o.DetectionType == "" {
```
ycdesu, line 35: The method name and comment do not indicate that it only sets default values when the value is not zero value. Should we remove all the if conditions?

narumiruna: If we remove all the if conditions, then it will overrides the non-zero options in `NewWithOptions`.

ycdesu (approved):

c9s (approved):

PR #1888, opened 2025-01-20
fix: misuse of unbuffered os.Signal channel as argument to signal.Notify

zenixls2:
signal channel should not be used elsewhere. Reported by go vet.
refer: https://github.com/golang/go/issue/45043

### DIFF
--- a/pkg/cmd/hoptimize.go
+++ b/pkg/cmd/hoptimize.go
@@ -94,11 +94,12 @@ var hoptimizeCmd = &cobra.Command{
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
+		sigs := make(chan os.Signal, 1)
+		signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+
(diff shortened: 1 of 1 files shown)


pkg/cmd/hoptimize.go
```diff
@@ -95,9 +95,7 @@ var hoptimizeCmd = &cobra.Command{
 		defer cancel()
 
 		go func() {
-			c := make(chan os.Signal)
-			signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
-			<-c
+			signal.Notify(make(chan os.Signal), syscall.SIGINT, syscall.SIGTERM)
```
c9s, line 98: but we need to block the call

the example did not change https://pkg.go.dev/os/signal#Notify

zenixls2: modified according to the requirement of blocking the call.
please check again.

c9s (approved):

c9s:
approved!

PR #1902, opened 2025-02-12
FEATURE: [depth] support binance futures orderbook depth buffer

anywhy:


pkg/depth/buffer.go
```diff
@@ -230,29 +259,43 @@ func (b *Buffer) fetchAndPush() error {
 	}
 
 	var pushUpdates []Update
-	for idx, u := range b.buffer {
-		// skip old events
-		if u.FinalUpdateID <= finalUpdateID {
-			continue
+	if b.isFutures {
```
c9s, line 262: what's the difference between futures and non-futures?

pkg/depth/buffer.go
```diff
@@ -230,29 +259,43 @@ func (b *Buffer) fetchAndPush() error {
...
+		for _, u := range b.buffer {
+			if u.FinalUpdateID < finalUpdateID {
+				continue
+			}
+			if u.FirstUpdateID <= finalUpdateID && u.FinalUpdateID >= finalUpdateID {
+				pushUpdates = append(pushUpdates, u)
+			}
```
c9s, line 269: I think this logic is the same as non-futures? it's just reversed ?

c9s: And the field `PreviousUpdateID` is not used here?

anywhy: I see, it s indeed the same logic. I ll merge it.

pkg/depth/buffer.go
anywhy: For futures, packet loss detection is stricter: it requires ``pu == previous u``. If this condition isn't met, it means data might be missing, so you need to reinitialize the order book.

For spot, as long as ``U <= lastUpdateId <= u``, the order book update continues. But if ``U`` > local order book update ID, you have to rebuild it from scratch.

pkg/depth/buffer.go
```diff
@@ -155,48 +160,72 @@ func (b *Buffer) AddUpdate(o types.SliceOrderBook, firstUpdateID int64, finalArg
...
+func (b *Buffer) handleMissingUpdate(u Update) {
+	// drop the prior updates in the buffer since it's corrupted
+	b.buffer = []Update{u}
+	b.resetSnapshot()
+	b.once.Reset()
+	b.once.Do(func() {
+		b.logger.Info("try fetching the snapshot due to missing update")
+		go b.tryFetch()
+	})
+	b.EmitReset()
```
c9s, line 226: EmitReset needs to be called after the mutex Unlock in order to avoid nested locking.

In your case, when calling this method, the mutex is still in the locked mode, did you verify this behavior? does it cause other effect?

pkg/depth/buffer.go
```diff
@@ -253,6 +282,10 @@ func (b *Buffer) fetchAndPush() error {
 	b.buffer = nil
 
 	// set the final update ID so that we will know if there is an update missing
+	if b.isFutures && len(pushUpdates) > 0 {
+		// reset finalUpdateId to first update event
+		finalUpdateID = pushUpdates[0].FinalUpdateID
```
c9s, line 287: sorry, why do we need to reset the finalUpdateId here? I don't understand

pkg/depth/buffer.go
```diff
@@ -236,7 +265,7 @@ func (b *Buffer) fetchAndPush() error {
 			continue
 		}
 
-		if u.FirstUpdateID > finalUpdateID+1 {
+		if u.FirstUpdateID > finalUpdateID+1 || (b.isFutures && idx > 0 && u.PreviousUpdateID != b.buffer[idx-1].FinalUpdateID) {
```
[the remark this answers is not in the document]
anywhy: I was originally planning to validate that the event in the buffer needs to satisfy U <= lastUpdateId <= u, but it seems unnecessary.

pkg/depth/buffer.go
```diff
@@ -163,29 +175,48 @@ func (b *Buffer) AddUpdate(o types.SliceOrderBook, firstUpdateID int64, finalArg
...
-		// drop the prior updates in the buffer since it's corrupted
-		b.buffer = []Update{u}
-		b.resetSnapshot()
-		b.once.Reset()
-		b.once.Do(func() {
-			b.logger.Info("try fetching the snapshot due to missing update")
-			go b.tryFetch()
-		})
+	if b.isFutures {
+		previousUpdateID := b.finalUpdateID
```
c9s, line 186: could you please document how do we need previousUpdateID for futures ? thanks!

anywhy: ok

c9s (approved):

PR #1909, opened 2025-02-18
FIX: [exchange] fix bugs and revise Coinbase request structs

dboyliao:
- add missing `slug` in the struct tags.
- update data types for fields.
- bug fix in pagination query function


dboyliao:
Despite the `Stream` for the websocket API is WIP, the coverage of REST API is above 70% (See attached pic)

截圖 2025-02-21 下午2 29 21

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
bailantaotao, line 1: ignore this file?

pkg/exchange/coinbase/api/v1/get_candles_request.go
```diff
@@ -17,15 +21,27 @@ type Candle struct {
 	Volume fixedpoint.Value `json:"volume"`
 }
 
-type GetCandlesResponse []Candle
+type GetCandlesResponse []RawCandle
+
+func (rc *RawCandle) Candle() *Candle {
+	values := *rc
```
bailantaotao, line 27: Verify that the rc is not nil and that the slice length is as expected. ( e.q. len == 6 ). Otherwise, the service will panic.

.vscode/settings.json
dboyliao: I think it's helpful to all the vscode users so I add it here.
It will help Go extension to load the local env file (just as instructed in the README) while running tests in vscode.
Should I ignore this?

bailantaotao (approved):

PR #1911, opened 2025-02-20
refactor: using slices.Contains to simplify the code

battmdpkq:
There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.


battmdpkq:
It seems to be unrelated to this change.

c9s (approved):

PR #1914, opened 2025-02-26
FIX: [exchange] fix bugs and add integration tests for Coinbase Exchange

dboyliao:
截圖 2025-02-26 下午2 12 55

### DIFF
--- a/pkg/exchange/coinbase/exchage.go
+++ b/pkg/exchange/coinbase/exchage.go
@@ -315,15 +315,15 @@ func (e *Exchange) QueryKLines(ctx context.Context, symbol string, interval type
 	if options.EndTime != nil {
 		req.End(*options.EndTime)
 	}
-	res, err := req.Do(ctx)
+	rawCandles, err := req.Do(ctx)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to get klines(%v): %v", interval, symbol)
 	}
-	candles := make([]api.Candle, 0, len(res))
-	for _, c := range res {
-		candle, err := c.Candle()
+	candles := make([]api.Candle, 0, len(rawCandles))
+	for _, rawCandle := range rawCandles {
+		candle, err := rawCandle.Candle()
 		if err != nil {
-			log.Warnf("invalid raw candle detected, skipping: %v", c)
+			log.Warnf("invalid raw candle detected, skipping: %v", rawCandle)
 			continue
 		}
 		candles = append(candles, *candle)
--- a/pkg/exchange/coinbase/exchage.go
+++ b/pkg/exchange/coinbase/exchage.go
@@ -145,18 +145,20 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
 		req.Price(order.Price)
 	}
 
-	// set time in force
-	switch order.TimeInForce {
-	case types.TimeInForceGTC:
-		req.TimeInForce("GTC")
-	case types.TimeInForceIOC:
-		req.TimeInForce("IOC")
-	case types.TimeInForceFOK:
-		req.TimeInForce("FOK")
-	case types.TimeInForceGTT:
-		req.TimeInForce("GTT")
-	default:
-		return nil, fmt.Errorf("unsupported time in force: %v", order.TimeInForce)
+	// set time in force, using default if not set
+	if len(order.TimeInForce) > 0 {
+		switch order.TimeInForce {
+		case types.TimeInForceGTC:
+			req.TimeInForce("GTC")
+		case types.TimeInForceIOC:
+			req.TimeInForce("IOC")
+		case types.TimeInForceFOK:
+			req.TimeInForce("FOK")
+		case types.TimeInForceGTT:
+			req.TimeInForce("GTT")
+		default:
+			return nil, fmt.Errorf("unsupported time in force: %v", order.TimeInForce)
+		}
 	}
 	// client order id
 	if len(order.ClientOrderID) > 0 {
@@ -286,7 +288,7 @@ func (e *Exchange) QueryTickers(ctx context.Context, symbol ...string) (map[stri
 	for _, s := range symbol {
 		ticker, err := e.QueryTicker(ctx, s)
 		if err != nil {
-			return nil, errors.Wrap(err, "failed to get tickers")
+			return nil, errors.Wrapf(err, "failed to get ticker for %v", s)
 		}
 		tickers[s] = *ticker
 	}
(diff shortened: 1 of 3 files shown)


pkg/exchange/coinbase/exchage.go
```diff
@@ -145,18 +145,20 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
...
-	case types.TimeInForceIOC:
-		req.TimeInForce("IOC")
-	case types.TimeInForceFOK:
-		req.TimeInForce("FOK")
-	case types.TimeInForceGTT:
-		req.TimeInForce("GTT")
-	default:
-		return nil, fmt.Errorf("unsupported time in force: %v", order.TimeInForce)
+	// set time in force, using default if not set
+	if len(order.TimeInForce) > 0 {
```
bailantaotao, line 149: Could you set a default timeInForce beforehand?

pkg/exchange/coinbase/exchage.go
```diff
@@ -308,25 +310,36 @@ func (e *Exchange) QueryKLines(ctx context.Context, symbol string, interval type
 		log.Warnf("limit %d is greater than the maximum limit 300, set to 300", options.Limit)
 		options.Limit = DefaultKLineLimit
 	}
-	granity := interval.String()
+	granity := fmt.Sprintf("%d", interval.Seconds())
```
bailantaotao, line 313: use `string(interval) `?

pkg/exchange/coinbase/exchage.go
```diff
@@ -308,25 +310,36 @@ func (e *Exchange) QueryKLines(ctx context.Context, symbol string, interval type
...
-		rawCandles = rawCandles[:options.Limit]
-	}
-	klines := make([]types.KLine, 0, len(rawCandles))
-	for _, rawCandle := range rawCandles {
-		candle := rawCandle.Candle()
-		klines = append(klines, toGlobalKline(symbol, granity, candle))
+	candles := make([]api.Candle, 0, len(res))
+	for _, c := range res {
+		candles = append(candles, *c.Candle())
+	}
```
bailantaotao, line 328: Most of us are accustomed to using pointers, for example make([]*api.Candle, 0, len(res)), so you don't have to dereference.

```
candles := make([]*api.Candle, 0, len(res))
	for _, c := range res {
		candles = append(candles, c.Candle())
	}
```

pkg/exchange/coinbase/exchage.go
```diff
@@ -308,25 +310,36 @@ func (e *Exchange) QueryKLines(ctx context.Context, symbol string, interval type
...
+	numCandles := len(candles)
+	klines := make([]types.KLine, 0, numCandles)
+	if numCandles > 0 {
+		for idx, candle := range candles {
+			kline := toGlobalKline(symbol, interval, &candle)
+			klines = append(klines, kline)
+			if idx > 0 {
+				klines[idx-1].StartTime = kline.EndTime
+			}
```
bailantaotao, line 340: 如果你都帶入 granity，那這個 start time和 end time的計算應該可以維持原邏輯？如果coinbase回傳非granity的間隔，那應該很有問題...

bailantaotao (approved):

pkg/exchange/coinbase/exchage.go
dboyliao: `string(interval)` will be `"1m"` but the Coinbase API is expecting `"60"`.

pkg/exchange/coinbase/exchage.go
dboyliao: As your review for PR #1909, I will apply the error handling here for invalid raw candles.
In that case, I think we don't need to use pointers.

pkg/exchange/coinbase/exchage.go
dboyliao: I want to leave it unset if it's not given by to user.
So it will fallback to Coinbase API defaults.

I want it to be consistent with Coinbase API so I didn't set a default value here.

pkg/exchange/coinbase/exchage.go
dboyliao: 有道理
重看了一下 doc: https://docs.cdp.coinbase.com/exchange/reference/exchangerestapi_getproductcandles
API 回傳的應該是 start time
我應該用下一根的 start time 當 end time 才對~
當時寫的時候應該不小心搞反了

dboyliao: 喔 不對
我應該用 `granity` 換算才對

dboyliao: Done.

pkg/exchange/coinbase/exchage.go
```diff
@@ -317,17 +319,20 @@ func (e *Exchange) QueryKLines(ctx context.Context, symbol string, interval type
...
+			log.Warnf("invalid raw candle detected, skipping: %v", rawCandle)
 			continue
 		}
-		klines = append(klines, toGlobalKline(symbol, granity, candle))
+		candles = append(candles, *candle)
+	}
+
+	klines := make([]types.KLine, 0, len(candles))
+	for _, candle := range candles {
+		kline := toGlobalKline(symbol, interval, &candle)
```
ycdesu, line 334: Bbgo uses go 1.21.6. Before go 1.22, the variable between each for loop iteration is the same variable. That means the `&candle` always points to the same address unless you use go 1.22. Please ensure this behavior is expected.

https://go.dev/blog/loopvar-preview

dboyliao: 截圖 2025-02-26 下午2 23 28
Since I do not keep the pointer (`&candle`) in the body of `toGlobalKline()`, I think it will do what is expected in go 1.21.6.

pkg/exchange/coinbase/exchage.go
gx578007: how about renaming the varialbe "granity" to "granularity"?

dboyliao: Opps, my bad.
Mispelling.
Will fix it.

ycdesu (approved):

pkg/exchange/coinbase/exchage.go
dboyliao: 截圖 2025-02-26 下午3 06 05
截圖 2025-02-26 下午3 06 14

I compile and run it with go1.21.6
As shown in the screenshot, there should be no issue where pointers outlive the loop.

pkg/exchange/coinbase/exchage.go
dboyliao: Done.

PR #1916, opened 2025-02-26
FEATURE: [exchange] Coinbase Websocket connection

dboyliao:
Implement `OnConnect` and websocket auth.

Reference implementation (Python):
https://github.com/coinbase-samples/exchange-scripts-py/blob/main/websocket/auth_ws_channel.py

### DIFF
--- a/pkg/exchange/coinbase/stream.go
+++ b/pkg/exchange/coinbase/stream.go
@@ -116,20 +116,15 @@ func (s *Stream) handleConnect() {
 	subProductsMap := make(map[string][]string)
 	for _, sub := range s.Subscriptions {
 		strChannel := string(sub.Channel)
-		if _, ok := subProductsMap[strChannel]; !ok {
-			subProductsMap[strChannel] = []string{}
-		}
 		// "rfq_matches" allow empty symbol
 		if sub.Channel != "rfq_matches" && len(sub.Symbol) == 0 {
 			continue
 		}
-		products := subProductsMap[strChannel]
-		products = append(products, sub.Symbol)
+		products := append(subProductsMap[strChannel], sub.Symbol)
 		subProductsMap[strChannel] = products
 	}
 	subCmds := []websocketCommand{}
 	signature, ts := s.generateSignature()
-	authEnabled := !s.PublicOnly && len(s.apiKey) > 0 && len(s.passphrase) > 0 && len(s.secretKey) > 0
 	for channel, productIDs := range subProductsMap {
 		var subType string
 		switch channel {
@@ -147,7 +142,7 @@ func (s *Stream) handleConnect() {
 				},
 			},
 		}
-		if authEnabled {
+		if s.authEnabled() {
 			subCmd.Signature = &signature
 			subCmd.Key = &s.apiKey
 			subCmd.Passphrase = &s.passphrase
@@ -161,6 +156,10 @@ func (s *Stream) handleConnect() {
 			log.WithError(err).Errorf("subscription error: %v", subCmd)
 		}
 	}
+}
+
+func (s *Stream) authEnabled() bool {
+	return !s.PublicOnly && len(s.apiKey) > 0 && len(s.passphrase) > 0 && len(s.secretKey) > 0
 }
 
 func (s *Stream) generateSignature() (string, string) {
--- a/pkg/exchange/coinbase/stream.go
+++ b/pkg/exchange/coinbase/stream.go
@@ -12,6 +12,7 @@ import (
 )
 
 const wsFeedURL = "wss://ws-feed.exchange.coinbase.com"
+const rfqMatchChannel = "rfq_matches"
 
 //go:generate callbackgen -type Stream
 type Stream struct {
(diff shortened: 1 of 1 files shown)


pkg/exchange/coinbase/stream.go
```diff
@@ -25,10 +37,156 @@ type Stream struct {
...
+	// subscribe to channels
+	if len(s.Subscriptions) == 0 {
+		return
+	}
+
+	subProductsMap := make(map[string][]string)
+	for _, sub := range s.Subscriptions {
+		strChannel := string(sub.Channel)
+		if _, ok := subProductsMap[strChannel]; !ok {
+			subProductsMap[strChannel] = []string{}
```
ycdesu, line 124: `[]string{}` is redundant. We don't have to initialize slice since `append` handles everything.

image

pkg/exchange/coinbase/stream.go
```diff
@@ -25,10 +37,155 @@ type Stream struct {
...
+
+	// public handlers
+	s.OnConnect(s.handleConnect)
+
 	return &s
 }
+
+func (s *Stream) dispatchEvent(e interface{}) {
+	switch e := e.(type) {
+	case *StatusMessage:
```
bailantaotao, line 63: if you use pointer here, your parser should return pointer address

pkg/exchange/coinbase/stream.go
```diff
@@ -25,10 +37,155 @@ type Stream struct {
...
+
+	subProductsMap := make(map[string][]string)
+	for _, sub := range s.Subscriptions {
+		strChannel := string(sub.Channel)
+		// "rfq_matches" allow empty symbol
+		if sub.Channel != "rfq_matches" && len(sub.Symbol) == 0 {
+			continue
+		}
+		products := append(subProductsMap[strChannel], sub.Symbol)
+		subProductsMap[strChannel] = products
```
bailantaotao, line 128: simplify to
```
subProductsMap[strChannel] = append(subProductsMap[strChannel], sub.Symbol)
```

pkg/exchange/coinbase/stream.go
```diff
@@ -25,10 +37,155 @@ type Stream struct {
...
+			subCmd.Timestamp = &ts
+		}
+		subCmds = append(subCmds, subCmd)
+	}
+	for _, subCmd := range subCmds {
+		err := s.Conn.WriteJSON(subCmd)
+		if err != nil {
+			log.WithError(err).Errorf("subscription error: %v", subCmd)
+		}
+	}
```
bailantaotao, line 162: Do we meet its requirements?

https://docs.cdp.coinbase.com/exchange/docs/websocket-rate-limits#subscription-limits

pkg/exchange/coinbase/stream.go
bailantaotao: Likely
```
func (s *Stream) parseMessage(data []byte) (msg interface{}, err error) {
	var baseMsg messageBaseType
	err = json.Unmarshal(data, &baseMsg)
	if err != nil {
		return
	}

	switch baseMsg.Type {
	case "heartbeat":
		var heartbeatMsg HeartbeatMessage
		err = json.Unmarshal(data, &heartbeatMsg)
                 return &heartbeatMsg, err
```

pkg/exchange/coinbase/stream.go
dboyliao: Done.

pkg/exchange/coinbase/stream.go
dboyliao: Done.

pkg/exchange/coinbase/stream.go
dboyliao: The subscription may be rejected if we exceed the limits.
I think the program should keep on running so I simply add a warning log here.

pkg/exchange/coinbase/stream.go
```diff
@@ -21,10 +33,146 @@ type Stream struct {
...
+	// subscribe to channels
+	if len(s.Subscriptions) == 0 {
+		return
+	}
+
+	subProductsMap := make(map[string][]string)
+	for _, sub := range s.Subscriptions {
+		strChannel := string(sub.Channel)
+		// "rfq_matches" allow empty symbol
+		if sub.Channel != "rfq_matches" && len(sub.Symbol) == 0 {
```
bailantaotao, line 112: define rfq_matches as constant on global

bailantaotao: What's the meaning of rfa_matchers?

pkg/exchange/coinbase/stream.go
```diff
@@ -21,10 +33,146 @@ type Stream struct {
...
+			subCmd.Key = &s.apiKey
+			subCmd.Passphrase = &s.passphrase
+			subCmd.Timestamp = &ts
+		}
+		subCmds = append(subCmds, subCmd)
+	}
+	for _, subCmd := range subCmds {
+		err := s.Conn.WriteJSON(subCmd)
+		if err != nil {
+			log.WithError(err).Errorf("subscription error: %v", subCmd)
```
bailantaotao, line 147: Should we return on error?

bailantaotao (approved):

pkg/exchange/coinbase/stream.go
dboyliao: It seems to me that websocket subscription is not of all-or-nothing semantic.
The user can still work on the feeds of the successful subscriptions despite that there are failed ones.

pkg/exchange/coinbase/stream.go
dboyliao: `rfq` stands for "Request For Quote".

I will make it a global const.

dboyliao: For details on rfq, please refer to the following link:
https://help.coinbase.com/en/exchange/trading-and-funding/request-for-quote-rfq

PR #1920, opened 2025-03-05
FIX: [exchange] (coinbase) fill missing SubmitOrder

dboyliao:
As title.

### DIFF
--- a/pkg/exchange/coinbase/convert.go
+++ b/pkg/exchange/coinbase/convert.go
@@ -12,6 +12,13 @@ import (
 
 func toGlobalOrder(cbOrder *api.Order) types.Order {
 	return types.Order{
+		SubmitOrder: types.SubmitOrder{
+			ClientOrderID: cbOrder.ClientOID,
+			Side:          cbOrder.Side.GlobalSideType(),
+			Quantity:      cbOrder.Size,
+			Price:         cbOrder.Price,
+			StopPrice:     cbOrder.StopPrice,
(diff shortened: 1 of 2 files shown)


narumiruna (approved):

pkg/exchange/coinbase/convert.go
```diff
@@ -12,12 +12,20 @@ import (
 
 func toGlobalOrder(cbOrder *api.Order) types.Order {
 	return types.Order{
+		SubmitOrder: types.SubmitOrder{
+			ClientOrderID: cbOrder.ClientOID,
+			Side:          cbOrder.Side.GlobalSideType(),
+			Quantity:      cbOrder.Size,
+			Price:         cbOrder.Price,
+			StopPrice:     cbOrder.StopPrice,
```
c9s, line 20: OrderType field is missing

and TimeInForce

pkg/exchange/coinbase/convert.go
```diff
@@ -12,12 +12,20 @@ import (
 
 func toGlobalOrder(cbOrder *api.Order) types.Order {
 	return types.Order{
+		SubmitOrder: types.SubmitOrder{
+			ClientOrderID: cbOrder.ClientOID,
+			Side:          cbOrder.Side.GlobalSideType(),
```
c9s, line 17: use toGlobalSide() instead of method,

because binding method will import the global types package from the api package.

PR #1924, opened 2025-03-05
FEAT: [coinbase] implement stream message handlers

dboyliao:
In this PR:
1. Implement handlers to bridge Coinbase events to standard events
2. Modify on-connection handler: publish balance snapshot event on connection.

### DIFF
--- a/pkg/exchange/coinbase/stream_handlers.go
+++ b/pkg/exchange/coinbase/stream_handlers.go
@@ -277,7 +277,7 @@ func (s *Stream) handleDoneMessage(msg *DoneMessage) {
 		return
 	}
 	isWorking := true
-	if msg.Reason == "cancelled" {
+	if msg.Reason == "canceled" {
 		isWorking = false
 	}
 	orderUpdate := types.Order{
--- a/pkg/exchange/coinbase/stream_handlers.go
+++ b/pkg/exchange/coinbase/stream_handlers.go
@@ -16,8 +16,7 @@ type channelType struct {
 }
 
 func (c channelType) String() string {
-	return fmt.Sprintf(`(Name:       %s,
-ProductIDs: %s)`, c.Name, c.ProductIDs)
+	return fmt.Sprintf("(channelType Name:%s, ProductIDs: %s)", c.Name, c.ProductIDs)
 }
 
 type authMsg struct {
@@ -28,10 +27,7 @@ type authMsg struct {
 }
 
 func (msg authMsg) String() string {
-	return fmt.Sprintf(`(Signature:  %s,
-Key:        ****,
-Passphrase: ****,
-Timestamp:  %s)`, msg.Signature, msg.Timestamp)
+	return fmt.Sprintf("(authMsg Signature:%s, Timestamp:%s)", msg.Signature, msg.Timestamp)
 }
 
 type subscribeMsgType1 struct {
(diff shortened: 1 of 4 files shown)


pkg/exchange/coinbase/stream_handlers.go
```diff
@@ -0,0 +1,344 @@
...
+					Timestamp:  ts,
+				},
+			}
+		}
+		subCmds = append(subCmds, subCmd)
+	}
+	for _, subCmd := range subCmds {
+		err := s.Conn.WriteJSON(subCmd)
+		if err != nil {
+			log.WithError(err).Errorf("subscription error: %v", subCmd)
```
ycdesu, line 115: Long with %v prints all fields of the struct. The struct contains exchange credentials. We shouldn't log the credentials. Note that error logs may also be sent to 3 party systems such as rollbar or sentry. It leaks the credential.

c9s (approved):

pkg/exchange/coinbase/stream_handlers.go
dboyliao: Fixed.
I also add a test on it.

pkg/exchange/coinbase/stream_handlers.go
```diff
@@ -0,0 +1,371 @@
...
+	"github.com/c9s/bbgo/pkg/types"
+)
+
+type channelType struct {
+	Name       string   `json:"name"`
+	ProductIDs []string `json:"product_ids,omitempty"`
+}
+
+func (c channelType) String() string {
+	return fmt.Sprintf(`(Name:       %s,
```
ycdesu, line 19: Looks like you add a `tab` before the value?
image

You can insert a tab using the `\t` directive, but in this case, I guess it's either a typo or you intend to align each line manually. However, when reading logs on different platforms such as Kibana, they may parse the logs differently, which can affect the alignment and make it look different from what you intended.

pkg/exchange/coinbase/stream_handlers.go
```diff
@@ -0,0 +1,371 @@
...
+	Signature  string `json:"signature,omitempty"`
+	Key        string `json:"key,omitempty"`
+	Passphrase string `json:"passphrase,omitempty"`
+	Timestamp  string `json:"timestamp,omitempty"`
+}
+
+func (msg authMsg) String() string {
+	return fmt.Sprintf(`(Signature:  %s,
+Key:        ****,
+Passphrase: ****,
```
ycdesu, line 33: We don't have to print all fields of the struct, so we could just ignore key and passphrase

pkg/exchange/coinbase/stream_handlers.go
```diff
@@ -0,0 +1,371 @@
...
+	Channels []channelType `json:"channels"`
+
+	authMsg
+}
+
+func (msg subscribeMsgType1) String() string {
+	return fmt.Sprintf(`(Type:     %s,
+Channels: %s,
+%s)`, msg.Type, msg.Channels, msg.authMsg.String())
+}
```
ycdesu, line 48: Another `tab`

pkg/exchange/coinbase/stream_handlers.go
dboyliao: Got it.
I will simplify the format string.

pkg/exchange/coinbase/stream_handlers.go
dboyliao: ok

pkg/exchange/coinbase/stream_handlers.go
```diff
@@ -0,0 +1,344 @@
...
+
+	// emit balance snapshot on connection
+	go func() {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+		defer cancel()
+
+		// query account balances
+		balances, err := s.exchange.QueryAccountBalances(ctx)
+		if err != nil {
+			log.WithError(err).Warn("failed to query account balances, the balance snapshot is initialized with empty balances")
```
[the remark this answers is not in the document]
dboyliao: After consult with @c9s, we prefer to emit an empty balance.
Will fix it accordingly.

ycdesu (approved):

PR #1926, opened 2025-03-07
FIX: [coinbase] submit order logic

dboyliao:
- bug fix
    - `size` is not required for all order
    - use `price * quantity` as `funds`, which is required parameter for market order on Coinbase Exchange.
    - update stop-limit order logic: using stop order on Coinbase Exchange. 
        - as a result, we only support long position on stop-limit order.

### DIFF
--- a/pkg/exchange/coinbase/exchage.go
+++ b/pkg/exchange/coinbase/exchage.go
@@ -167,7 +167,12 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
 	case types.OrderTypeMarket:
 		switch order.Side {
 		case types.SideTypeBuy:
-			funds := order.Price.Mul(order.Quantity)
+			// buy with market price
+			ticker, err := e.QueryTicker(ctx, order.Market.Symbol)
+			if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/exchange/coinbase/exchage.go
```diff
@@ -132,6 +136,16 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
 		req.OrderType("limit")
 	case types.OrderTypeMarket:
 		req.OrderType("market")
+	case types.OrderTypeStopLimit:
+		req.OrderType("stop")
+		switch order.Side {
+		case types.SideTypeBuy:
+			req.Stop("entry") // trigger order when the price <= stop price
+		case types.SideTypeSell:
+			req.Stop("loss") // trigger order when the price >= stop price
```
c9s, line 145: this is strategy dependent i think

dboyliao: I see it over here:
https://docs.cdp.coinbase.com/exchange/reference/exchangerestapi_postorders#stop-orders

So that's why I think it's about stop-limit order under the context of bbgo.

c9s (approved):

PR #1927, opened 2025-03-07
FIX: [coinbase] fix stream handlers and order placing logics

dboyliao:
The purpose of this PR is to:
1. minor refactoring, moving type conversion logics from api modules to application module
2. renaming `ActiveMessage` as `ActivateMessage` to be aligned with Coinbase documentation.

These are preliminary works for the upcoming PR #1928.

### DIFF
--- a/pkg/exchange/coinbase/convert.go
+++ b/pkg/exchange/coinbase/convert.go
@@ -3,6 +3,7 @@ package coinbase
 import (
 	"hash/fnv"
 	"math"
+	"strings"
 	"time"
 
 	api "github.com/c9s/bbgo/pkg/exchange/coinbase/api/v1"
@@ -18,6 +19,25 @@ func toGlobalSide(cbSide api.SideType) types.SideType {
 		return types.SideTypeSell
 	}
 	return types.SideTypeNone
+}
+
+func toGlobalOrderStatus(order *api.Order) types.OrderStatus {
+	switch order.Status {
+	case api.OrderStatusRejected:
+		return types.OrderStatusRejected
+	case api.OrderStatusReceived, api.OrderStatusOpen, api.OrderStatusPending:
+		return types.OrderStatusNew
+	case api.OrderStatusDone:
+		switch order.DoneReason {
+		case "filled":
+			return types.OrderStatusFilled
+		case "canceled":
+			return types.OrderStatusCanceled
+		case "rejected":
+			return types.OrderStatusRejected
+		}
+	}
+	return types.OrderStatus(strings.ToUpper(string(order.Status)))
 }
 
 func toGlobalOrder(cbOrder *api.Order) types.Order {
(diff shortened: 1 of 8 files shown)


pkg/exchange/coinbase/convert.go
```diff
@@ -20,6 +21,31 @@ func toGlobalSide(cbSide api.SideType) types.SideType {
...
+		return types.OrderStatusReceived
+	case api.OrderStatusOpen, api.OrderStatusPending:
+		return types.OrderStatusNew
+	case api.OrderStatusDone:
+		switch order.DoneReason {
+		case "filled":
+			if order.FilledSize.Eq(order.Size) {
+				return types.OrderStatusFilled
+			} else {
+				return types.OrderStatusPartiallyFilled
```
c9s, line 38: return types.OrderStatusCanceled for this line?

pkg/exchange/coinbase/convert.go
```diff
@@ -20,6 +21,31 @@ func toGlobalSide(cbSide api.SideType) types.SideType {
 	return types.SideTypeNone
 }
 
+func toGlobalOrderStatus(order *api.Order) types.OrderStatus {
+	switch order.Status {
+	case api.OrderStatusRejected:
+		return types.OrderStatusRejected
+	case api.OrderStatusReceived:
+		return types.OrderStatusReceived
```
c9s, line 29: Is Received = Update?

this means if an order is working, then you will receive incremental "ExecutedQuantity" ? you might need update the status to PartiallyFilled and add the same check like LINE35-39?

c9s (approved):

pkg/exchange/coinbase/convert.go
kbearXD, line 29: The status "received" in coinbase means that the order you submitted is valid and it will go into the matching engine. I think it's closed to "NEW" in bbgo.

The status "open" in coinbase means that the order is on the orderbook, so if you submit a order and it will be fully taker order. there will no "open" status. I think we need to check the remaining size to know the status is bbgo's PartiallyFilled or New

pkg/exchange/coinbase/convert.go
dboyliao: https://docs.cdp.coinbase.com/exchange/docs/websocket-channels#done

It should be `filled`.
No partially filled for done message.
Will fix it.

pkg/exchange/coinbase/convert.go
dboyliao: After discussion with @kbearXD, I think received is more close to new in bbgo.
In that case, a dedicated status for received is not necessary.
I will remove it and map received status on Coinbase to new in bbgo.

kbearXD (approved):

c9s (approved):

PR #1928, opened 2025-03-08
FIX: [coinbase] refactoring and fix order-related streaming event handling

dboyliao:
1. update `.SubmitOrder` logic, to simplify the streaming handlers logic
    - Do not support `limit order with funds`
    - Only support stop limit order (via `stop` order on Coinbase Exchange) at long position .
2. By `1.`, update order-related event handlers accordingly.

### DIFF
--- a/pkg/exchange/coinbase/stream_handlers.go
+++ b/pkg/exchange/coinbase/stream_handlers.go
@@ -10,9 +10,24 @@ import (
 	"github.com/c9s/bbgo/pkg/types"
 )
 
+type ChannelName = string
+
+const (
+	rfqMatchChannel    ChannelName = "rfq_matches"
+	statusChannel      ChannelName = "status"
+	auctionChannel     ChannelName = "auctionfeed"
+	matchesChannel     ChannelName = "matches"
+	tickerChannel      ChannelName = "ticker"
+	tickerBatchChannel ChannelName = "ticker_batch"
+	level2Channel      ChannelName = "level2"
+	level2BatchChannel ChannelName = "level2_batch"
+	fullChannel        ChannelName = "full"
+	balanceChannel     ChannelName = "balance"
+)
+
 type channelType struct {
-	Name       string   `json:"name"`
-	ProductIDs []string `json:"product_ids,omitempty"`
+	Name       ChannelName `json:"name"`
+	ProductIDs []string    `json:"product_ids,omitempty"`
 }
 
 func (c channelType) String() string {
@@ -71,7 +86,7 @@ func (s *Stream) handleConnect() {
 	for _, sub := range s.Subscriptions {
 		strChannel := string(sub.Channel)
 		// rfqMatchChannel allow empty symbol
-		if sub.Channel != rfqMatchChannel && sub.Channel != "status" && len(sub.Symbol) == 0 {
+		if strChannel != rfqMatchChannel && strChannel != "status" && len(sub.Symbol) == 0 {
 			continue
 		}
 		subProductsMap[strChannel] = append(subProductsMap[strChannel], sub.Symbol)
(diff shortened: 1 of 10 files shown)


c9s (approved):

pkg/exchange/coinbase/stream_handlers.go
```diff
@@ -366,3 +499,26 @@ func (s *Stream) checkAndUpdateSequenceNumber(msgType MessageType, currentSeqNum
...
+func (s *Stream) updateWorkingOrders(orders ...types.Order) {
+	s.lockWorkingOrderMap.Lock()
+	defer s.lockWorkingOrderMap.Unlock()
+
+	for _, order := range orders {
+		exisiting, ok := s.workingOrdersMap[order.UUID]
+		if ok {
+			if !order.IsWorking {
+				// order is already in the map and not working, remove it
+				delete(s.workingOrdersMap, order.UUID)
```
c9s, line 512: calling "delete" in the order iteration could be dangerous.

c9s: oh it's different map, then it's ok

pkg/exchange/coinbase/stream_test.go
```diff
@@ -49,3 +54,199 @@ func Test_SubCmdString(t *testing.T) {
...
+			break
+		}
+	}
+}
+
+func TestBalance(t *testing.T) {
+	accountsStr := os.Getenv("COINBASE_ACCOUNT_IDS")
+	accounts := strings.Split(accountsStr, ",")
+
+	c := make(chan bool)
```
c9s, line 217: you can use `make(chan struct[})` and send `c <- struct{}{}`

pkg/exchange/coinbase/stream_test.go
```diff
@@ -49,3 +54,199 @@ func Test_SubCmdString(t *testing.T) {
...
+			if triggered {
+				return
+			}
+			triggered = true
+			assert.NotNil(t, m)
+			t.Logf("get balance message: %v", *m)
+			c <- true
+		})
+		err := stream.Connect(context.Background())
+		assert.NoError(t, err)
```
c9s, line 235: you need to add `<- c` here?

pkg/exchange/coinbase/stream_test.go
```diff
@@ -49,3 +54,199 @@ func Test_SubCmdString(t *testing.T) {
...
+	})
+	<-c
+}
+
+func getTestStreamOrSkip(t *testing.T) *Stream {
+	if isCI, _ := strconv.ParseBool(os.Getenv("CI")); isCI {
```
c9s, line 241: IntegrationTestWithPassphraseConfigured already checks the params for you, if the env vars are not defined, then the test will be skipped.

dboyliao: Really?
I  check the `IntegrationTestWithPassphraseConfigured`.
It does not check the `"CI"` env variable.

pkg/exchange/coinbase/stream_test.go
dboyliao: I do `<-c` at line 237 already.

ycdesu (approved):

pkg/exchange/coinbase/stream_test.go
```diff
@@ -49,3 +54,199 @@ func Test_SubCmdString(t *testing.T) {
...
+		}
+		triggered := false
+		stream.OnTickerMessage(func(m *TickerMessage) {
+			if triggered {
+				return
+			}
+			triggered = true
+			assert.NotNil(t, m)
+			// t.Logf("get ticker message: %v", *m)
+			c <- *m
```
c9s, line 95: select {
   case c <- m:
   default:
}

select {
   case c <- *m:
}

PR #1936, opened 2025-03-16
FIX: [sync service] inconsistency of the query conditions on existing trades

dboyliao:
The query on existing trade records from the DB was based on trade id and limit.
Whereas, the query via the exchange API is based on timestamp and trade id.
The inconsistency on the query conditions may cause error on corner cases.
There are chances the ids map may fail to filter out existing trade records and resulting in duplicate key error while inserting.

In this PR, I add an extra query for the record with the last trade id.
And I update the ids map accordingly so that self-trade records on both side will be included.

### DIFF
--- a/pkg/service/sync_task.go
+++ b/pkg/service/sync_task.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"reflect"
 	"sort"
+	"strconv"
 	"time"
 
 	"github.com/Masterminds/squirrel"
+	"github.com/c9s/bbgo/pkg/types"
 	"github.com/jmoiron/sqlx"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


c9s (approved):

pkg/service/sync_task.go
```diff
@@ -208,6 +216,46 @@ func buildIdMap(sel SyncTask, recordSliceRef reflect.Value) map[string]struct{}
 		id := sel.ID(entryRef.Interface())
 		ids[id] = struct{}{}
 	}
-
 	return ids
 }
+
+func patchSelfTrade(ctx context.Context, db *sqlx.DB, sel SyncTask, recordSliceRef reflect.Value, ids *map[string]struct{}) error {
```
c9s, line 222: map itself is already a pointer, so you can just pass it

dboyliao: Done.

kbearXD (approved):

PR #1938, opened 2025-03-17
FEATURE: [bbgo] pass logger as parameter into bbgo.Sync

kbearXD:


c9s (commented):
can we use Isolation for the logger instead of changing the interface?

c9s (approved):

ycdesu (approved):

PR #1949, opened 2025-03-26
FEATURE: [depthmaker] add depth RESTful API fallback support

c9s:


kbearXD (approved):

pkg/exchange/max/exchange.go
```diff
@@ -1149,7 +1149,16 @@ func (e *Exchange) QueryDepth(
...
 	req := e.v3client.NewGetDepthRequest()
 	req.Market(symbol)
-	req.Limit(limit)
+
+	// maximum limit is 300
+	if limit > 300 {
+		limit = 300
+	}
+
+	if limit > 0 {
```
gx578007, line 1158: Need to check `limit <= 0`?

dboyliao (approved):

pkg/exchange/max/exchange.go
c9s: ok

c9s: fixed

PR #1960, opened 2025-03-28
IMPROVE: [bitget] improve bitget time in force support

c9s:
Add IOC support on bitget

### DIFF
--- a/pkg/exchange/bitget/exchange.go
+++ b/pkg/exchange/bitget/exchange.go
@@ -319,11 +319,27 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
 		return nil, fmt.Errorf("time-in-force %s not supported", order.TimeInForce)
 	}
 
+	// bitget's "force" field is mandatory, so we need to set it.
 	switch order.Type {
 	case types.OrderTypeLimitMaker:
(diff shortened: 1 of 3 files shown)


narumiruna (approved):

gx578007 (approved):

pkg/exchange/bitget/exchange.go
```diff
@@ -319,11 +319,26 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
 		return nil, fmt.Errorf("time-in-force %s not supported", order.TimeInForce)
 	}
 
+	// bitget's "force" field is mandatory, so we need to set it.
 	switch order.Type {
 	case types.OrderTypeLimitMaker:
 		req.Force(v2.OrderForcePostOnly)
+	case types.OrderTypeMarket:
+		// From the doc: It is invalid when orderType is market)
```
kbearXD, line 327: Should we return error here?
If someone submit order with market, stop limit, stop market, it will submit no price order ?

c9s: 好像一定要給 time in force

c9s: i fixed it by assigning it to GTC

PR #1965, opened 2025-04-02
FEATURE:[okx]support okx futures order and fix examples

anywhy:


c9s (approved):

symbols.go
```diff
@@ -0,0 +1,1032 @@
+// Code generated by go generate; DO NOT EDIT.
```
dboyliao, line 1: This file is not used.
Please remove it.

I think it's generate by the `//go:generate sh ...` in the `pkg/exchange/okex/convert.go` at line 18

bailantaotao (approved):

dboyliao (commented):
Please add tests to pass the `codecov/patch` check.

symbols.go
anywhy: ok

pkg/exchange/okex/convert.go
```diff
@@ -11,12 +11,21 @@ import (
 )
 
 func toGlobalSymbol(symbol string) string {
+	symbol = strings.TrimSuffix(symbol, "-SWAP")
 	return strings.ReplaceAll(symbol, "-", "")
 }
 
 //go:generate sh -c "echo \"package okex\nvar spotSymbolMap = map[string]string{\n\" $(curl -s -L 'https://www.okx.com/api/v5/public/instruments?instType=SPOT' | jq -r '.data[] | \"\\(.instId | sub(\"-\" ; \"\") | tojson ): \\( .instId | tojson),\n\"') \"\n}\" > symbols.go"
```
dboyliao, line 18: Please remove this line as well.

anywhy: done

PR #1969, opened 2025-04-08
FIX: [logging] add more informative logging to StreamOrderBook

dboyliao:


pkg/types/orderbook.go
```diff
@@ -252,6 +257,17 @@ func (sb *StreamOrderBook) emitChange(signalType BookSignalType, bookTime time.T
 	}
 }
 
+func (sb *StreamOrderBook) SetLogger(logger logrus.FieldLogger) {
+	sb.logger = logger
+}
+
+func (sb *StreamOrderBook) log(msg string, args ...interface{}) {
```
ycdesu, line 264: The message is logged in `info` level. It's better to rename the method to `logInfo` or other names, so user could know its side effect immediately,

ycdesu (approved):

pkg/types/orderbook.go
dboyliao: Fixed.

PR #1974, opened 2025-04-10
FIX: [xgap] prevent order placement on price collision with existing orders

dboyliao:


pkg/strategy/xgap/strategy.go
```diff
@@ -394,6 +394,14 @@ func (s *Strategy) placeOrders(ctx context.Context) {
 
 	log.Infof("%s adjusted quantity: %f", s.Symbol, quantity.Float64())
 
+	bid, ask, _ := s.tradingBook.BestBidAndAsk()
+	if price.Compare(bid.Price) <= 0 || price.Compare(ask.Price) >= 0 {
+		log.Infof(
```
c9s, line 401: use s.logger.Warnf ?

c9s (approved):

pkg/strategy/xgap/strategy.go
dboyliao: ok

kbearXD (approved):

PR #1975, opened 2025-04-11
FIX: [bitget] change MinTradeAmount and MaxTradeAmount types to string and handle conversion error

dboyliao:


bailantaotao (approved):

pkg/exchange/bitget/convert.go
```diff
@@ -29,6 +29,21 @@ func toGlobalMarket(s v2.Symbol) types.Market {
 		log.Warnf("The market symbol status %s is not online: %s", s.Symbol, s.Status)
 	}
 
+	var minQuantity, maxQuantity fixedpoint.Value
+	if v, err := fixedpoint.NewFromString(s.MinTradeAmount); err == nil {
+		minQuantity = v
+	} else {
+		log.Warnf("fail to convert min trade amount %s for %s, setting it to 0: %v", s.MinTradeAmount, s.Symbol, err)
+		minQuantity = fixedpoint.NewFromFloat(0.0)
```
gx578007, line 37: Maybe use the min fixedpoint like "1e-8"?

pkg/exchange/bitget/convert.go
```diff
@@ -29,6 +29,21 @@ func toGlobalMarket(s v2.Symbol) types.Market {
...
+	}
+
+	if v, err := fixedpoint.NewFromString(s.MaxTradeAmount); err == nil {
+		maxQuantity = v
+	} else {
+		log.Warnf("fail to convert max trade amount %s for %s, setting it to unlimited (Inf): %v", s.MaxTradeAmount, s.Symbol, err)
```
gx578007, line 43: Could you hint the input `s.MaxTradeAmount` can overflow in the log message.

dboyliao: I include the `err` in the log message.
It should be clear enough it's overflow or anything else that cause the parsing failure.

pkg/exchange/bitget/convert.go
dboyliao: Considering trading, I think `0` is a safer value to use as min amount.

ycdesu (approved):

PR #1977, opened 2025-04-11
REFACTOR: [xgap] balance checks on orders and skip large quantity for making spread orders

dboyliao:


pkg/util/tradingutil/orders.go
```diff
@@ -0,0 +1,53 @@
+package tradingutil
+
+import (
+	"fmt"
+
+	"github.com/c9s/bbgo/pkg/fixedpoint"
+	"github.com/c9s/bbgo/pkg/types"
+)
+
+func CheckQuantityByBalance(
```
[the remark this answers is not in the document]
dboyliao: Done

pkg/util/tradingutil/orders.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+func HasSufficientBalance(
+	baseCurrency string,
+	quoteCurrency string,
+	side types.SideType,
+	price, quantity fixedpoint.Value,
+	tradingMarket types.Market,
+	balances types.BalanceMap,
+) (bool, string) {
```
c9s, line 17: i think the problem of this function you got too many arguments, better to avoid it

For side buy, you don't need Base asset balance, you only need Quote asset balance and the buy price to calculate the actual quantity.

For side sell, you only need Base asset balance, quote balance and price are not used for this case

pkg/types/position.go
```diff
@@ -218,14 +218,22 @@ func (p *Position) GetBase() (base fixedpoint.Value) {
...
-// GetBaseAndAverageCost locks the mutex and return the base quantity and average cost
-func (p *Position) GetBaseAndAverageCost() (base fixedpoint.Value, averageCost fixedpoint.Value) {
+type PositionStatus struct {
+	Base        fixedpoint.Value
+	AverageCost fixedpoint.Value
+	Type        PositionType
+}
+
+// GetStatus locks the mutex and return the position status
+func (p *Position) GetStatus() PositionStatus {
```
c9s, line 228: why?

pkg/types/position.go
```diff
@@ -218,14 +218,22 @@ func (p *Position) GetBase() (base fixedpoint.Value) {
 	return base
 }
 
-// GetBaseAndAverageCost locks the mutex and return the base quantity and average cost
-func (p *Position) GetBaseAndAverageCost() (base fixedpoint.Value, averageCost fixedpoint.Value) {
```
c9s, line 222: you are actually changing an existing public API which could be used from somewhere else, it could be a private repository or a private strategy where we can't scan the reference

c9s: we need to be careful when changing an existing public api

pkg/util/tradingutil/orders.go
```diff
@@ -0,0 +1,53 @@
...
+	// 1. check if the currency is correct
+	_, okBase := balances[baseCurrency]
+	_, okQuote := balances[quoteCurrency]
+	if !okBase || !okQuote {
+		return false, fmt.Sprintf("invalid currency paire: %s/%s", baseCurrency, quoteCurrency)
```
c9s, line 22: maybe return type error? and use fmt.Errorf() for the error message

pkg/types/position.go
dboyliao: I think the issue lies in the need for a thread-safe way to access position-related information,
which is used to determine trading conditions.

Perhaps, just like we discussed offline,
renaming it to "snapshot" would make it easier to understand.

pkg/util/tradingutil/orders.go
dboyliao: ok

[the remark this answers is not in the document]
dboyliao: Opps

pkg/util/tradingutil/orders.go
dboyliao: I want to have a unified way to check the balance when submitting an order with `SubmitOrder`.
Maybe it will be better if I pass `SubmitOrder` instead.

pkg/util/tradingutil/orders.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	// 2. check if there is available balance
+	switch side {
+	case types.SideTypeSell:
+		balance := balances[baseCurrency]
+		if balance.Available.Compare(quantity) < 0 {
```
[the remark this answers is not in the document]
dboyliao: I'm not thinking about a guarantee on placing order.
Instead, I'd like to leave it to the caller to decide what to do next if there is an error concerning the balances.
That's why I return an error here.

pkg/strategy/xgap/strategy.go
```diff
@@ -237,17 +237,28 @@ func (s *Strategy) CrossRun(ctx context.Context, _ bbgo.OrderExecutionRouter, se
...
+	order := types.SubmitOrder{
+		Symbol:      s.Symbol,
+		Side:        types.SideTypeSell,
+		Type:        types.OrderTypeLimit,
+		Quantity:    fixedpoint.Max(bestBid.Volume, s.tradingMarket.MinQuantity),
+		Price:       bestBid.Price,
+		Market:      s.tradingMarket,
+		TimeInForce: types.TimeInForceIOC,
+	}
+	if err := tradingutil.HasSufficientBalance(
```
c9s, line 250: HasSufficientBalance sounds you will return bool, but it returns error

pkg/strategy/xgap/strategy.go
```diff
@@ -237,17 +237,28 @@ func (s *Strategy) CrossRun(ctx context.Context, _ bbgo.OrderExecutionRouter, se
...
+		order,
+		balances,
+	); err != nil {
+		return fmt.Errorf(
+			"cannot make spread for %s: %s",
+			s.tradingMarket.BaseCurrency,
+			err,
+		)
 	}
+	orderForms := []types.SubmitOrder{order}
```
c9s, line 261: we will always have only one make spread order, hence no need to create a slice?

pkg/strategy/xgap/strategy.go
```diff
@@ -260,56 +271,82 @@ func (s *Strategy) makeSpread(ctx context.Context, bestBid, bestAsk types.PriceV
...
 }
 
-func getAdjustPositionOrder(symbol string, tradingMarket types.Market, currentPosition *types.Position, bestBid, bestAsk types.PriceVolume) (orderForms []types.SubmitOrder) {
-	basePosition, avgCost := currentPosition.GetBaseAndAverageCost()
-	if basePosition.Eq(fixedpoint.Zero) {
+// getAdjustPositionOrders returns the order forms that can adjust the current position if possible.
+// When in a short position, it will try to place a buy order.
+// When in a long position, it will try to place a sell order.
+// It's based on the spread and the condition of the current position.
+func getAdjustPositionOrders(
```
c9s, line 278: you are generating an new order, not returning an existing order, so using prefix "get" could be confusing

you may use "build" or "new" as the prefix

pkg/strategy/xgap/strategy.go
dboyliao: I adopt this from the original code.

pkg/strategy/xgap/strategy.go
dboyliao: Fixed

pkg/util/tradingutil/orders.go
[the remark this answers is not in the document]
dboyliao: fixed

pkg/util/tradingutil/orders.go
dboyliao: Updated function signature:
```
func HasSufficientBalance(
	tradingMarket types.Market,
	submitOrder types.SubmitOrder,
	balances types.BalanceMap,
) (bool, error)
```

pkg/types/position.go
dboyliao: I leave the core API unchanged and move the locking util function to a private module in `xgap` folder instead.

pkg/strategy/xgap/strategy.go
```diff
@@ -271,56 +282,82 @@ func (s *Strategy) makeSpread(ctx context.Context, bestBid, bestAsk types.PriceV
...
 			Type:        types.OrderTypeLimit,
-			Quantity:    fixedpoint.Max(bestBid.Volume, tradingMarket.MinQuantity),
-			Price:       bestBid.Price,
+			Quantity:    quantity,
+			Price:       price,
 			Market:      tradingMarket,
 			TimeInForce: types.TimeInForceIOC,
-		})
+		}
+		if ok, err := tradingutil.HasSufficientBalance(
```
c9s, line 337: pull out this check to the caller

pkg/strategy/xgap/strategy.go
```diff
@@ -271,56 +282,82 @@ func (s *Strategy) makeSpread(ctx context.Context, bestBid, bestAsk types.PriceV
...
 			bestBid.Price.String(),
-			avgCost.String(),
+			positionSnapshot.AverageCost.String(),
 		)
-		orderForms = append(orderForms, types.SubmitOrder{
+		tryPlaceOrder = true
+		pv = bestBid
+		side = types.SideTypeSell
+	}
+	if tryPlaceOrder {
```
c9s, line 325: you can check if the pv is assigned instead of introducing a new var

dboyliao: ok.
I add a `.IsZero()` method instead.

pkg/strategy/xgap/strategy.go
dboyliao: ok

c9s (approved):

ycdesu (approved):

PR #1982, opened 2025-04-15
REFACTOR: [xgap] quantity validation checks on adjusting position orders

dboyliao:


pkg/strategy/xgap/strategy.go
```diff
@@ -328,29 +331,50 @@ func (s *Strategy) placeOrders(ctx context.Context) {
...
-			)
-			if balanceOk, balanceErr := tradingutil.HasSufficientBalance(
-				s.tradingMarket,
-				adjPosOrder,
-				balances,
-			); adjPosIsPossible && balanceOk {
+			if orderPlaced := func() bool {
+				positionSnapshot := getPositionSnapshot(s.Position)
+				s.logger.Infof("position snapshot: %+v", positionSnapshot)
+				adjPosIsPossible, adjPosOrder := buildAdjustPositionOrder(
```
c9s, line 339: can be shorter as "canAdjust" or "canAdj"

pkg/strategy/xgap/strategy.go
```diff
@@ -328,29 +331,50 @@ func (s *Strategy) placeOrders(ctx context.Context) {
...
+					s.logger.WithError(balanceErr).Warnf("cannot place position adjusting order: %+v", adjPosOrder)
+					return false
+				}
+
 				s.logger.Infof("adjust position order forms: %+v", adjPosOrder)
 				createdOrders, err := s.OrderExecutor.SubmitOrders(ctx, adjPosOrder)
 				if err != nil {
-					log.WithError(err).Errorf("order submit error for adjust position, created orders: %+v", createdOrders)
+					s.logger.WithError(err).Errorf("order submit error for adjust position, created orders: %+v", createdOrders)
+					if err := s.OrderExecutor.GracefulCancel(ctx, createdOrders...); err != nil {
```
c9s, line 369: if you got error, you don't have to cancel it.

pkg/strategy/xgap/strategy.go
```diff
@@ -328,29 +331,50 @@ func (s *Strategy) placeOrders(ctx context.Context) {
 	balances := s.tradingSession.GetAccount().Balances()
 	// try to use the bid/ask price from the trading book
 	if hasPrice {
-		// try to place orders that can adjust the position
+		// try to place orders that can adjust the position if possible
 		if s.Position != nil {
```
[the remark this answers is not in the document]
dboyliao: So the logic is like this:
1. If position is not nil, try to place an adjusting position order if possible
2. If an adjusting position order is not possible, proceed to place self-trade orders within the gap.

pkg/strategy/xgap/strategy.go
```diff
@@ -519,14 +544,32 @@ func (s *Strategy) placeOrders(ctx context.Context) {
 			Market:      s.tradingMarket,
 			TimeInForce: types.TimeInForceIOC,
 		},
+	} {
+		if s.tradingMarket.IsDustQuantity(order.Quantity, order.Price) {
```
c9s, line 548: this check can be done before the loop?

c9s: because the price and the quantity are the same

pkg/strategy/xgap/strategy.go
dboyliao: Got it.
I forget that the order is an IOC order.

pkg/strategy/xgap/strategy.go
dboyliao: Oh, you are right.

c9s (approved):

pkg/strategy/xgap/strategy.go
```diff
@@ -323,34 +326,59 @@ func buildAdjustPositionOrder(
...
+		return false
+	}
+	// sanity checks on the order
+	// check dust quantity
+	if s.tradingMarket.IsDustQuantity(adjPosOrder.Quantity, adjPosOrder.Price) {
+		s.logger.Warnf("adjust position order is of dust quantity: %+v", adjPosOrder)
+		return false
+	}
+	// check balance
+	balanceOk, balanceErr := tradingutil.HasSufficientBalance(
```
ycdesu, line 348: Don't we have to check the balance error? Or it implies if balanceOk is false, the error will always exist?

dboyliao: I think what I try to achieve here is to make placing order more robust.
In addition to that, I add some logging suggesting insufficient balance which can help debugging in production.

pkg/strategy/xgap/strategy.go
```diff
@@ -306,10 +306,13 @@ func buildAdjustPositionOrder(
 		pv = bestBid
 		side = types.SideTypeSell
 	}
+	price := pv.Price
+	quantity := fixedpoint.Min(
+		positionSnapshot.Base.Abs(),
```
gx578007, line 311: What if position is less than min quantity?

dboyliao: Good point.

I think I should only making the quantity be no larger than the position in this function.
Then leave the check on dust quantity to the caller, which will be the check at L343.

dboyliao: fixed.

gx578007 (approved):

ycdesu (approved):

PR #1986, opened 2025-04-16
FEATURE: [coinbase] Implement cancel orders by symbol (with minor fix on balances)

dboyliao:


c9s (approved):

pkg/exchange/coinbase/exchange.go
```diff
@@ -451,3 +452,24 @@ func (e *Exchange) queryOrderTradesByPagination(ctx context.Context, q types.Ord
 	}
 	return cbTrades, nil
 }
+
+// Utils
```
c9s, line 456: fix comment?

pkg/exchange/coinbase/exchange.go
```diff
@@ -451,3 +452,24 @@ func (e *Exchange) queryOrderTradesByPagination(ctx context.Context, q types.Ord
...
+func (e *Exchange) CancelOrdersBySymbol(ctx context.Context, symbol string) ([]types.Order, error) {
+	cbOrders, err := e.queryOrdersByPagination(ctx, symbol, []string{})
```
c9s, line 458: why query before canceling?

pkg/exchange/coinbase/exchange.go
dboyliao: done.

pkg/exchange/coinbase/exchange.go
dboyliao: Since the coinbase API will only return order ids that are canceled, I need to do the queries to retrieve price, quantity, ...etc.
However, I think you get a point.
I should delete the orders first then fetch the data I need afterward.
In such way, the result should be consistent.

pkg/exchange/coinbase/exchange_test.go
```diff
@@ -104,6 +104,37 @@ func Test_OrdersAPI(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func Test_CancelOrdersBySymbol(t *testing.T) {
```
c9s, line 107: convention: TestExchange_CancelOrdersBySymbol

pkg/exchange/coinbase/exchange.go
```diff
@@ -451,3 +452,29 @@ func (e *Exchange) queryOrderTradesByPagination(ctx context.Context, q types.Ord
...
+	}
+	if len(canceledOrderIds) == 0 {
+		return nil, nil
+	}
+
+	// Once the order is cancelled, it can not be queried by the API (always 404 error)
+	// As a result, we simply construct empty orders here (without price, quantity, side ...etc)
+	orders := make([]types.Order, 0, len(canceledOrderIds))
+	for _, orderID := range canceledOrderIds {
+		orders = append(orders, types.Order{
```
c9s, line 472: use QueryOrder here?

c9s: oh 404

c9s (approved):

pkg/exchange/coinbase/exchange_test.go
dboyliao: ok

kbearXD (approved):

PR #1993, opened 2025-04-21
REFACTOR: [session] Refactor max borrowable update

dboyliao:


pkg/bbgo/session.go
```diff
@@ -156,6 +157,9 @@ type ExchangeSession struct {
 	logger log.FieldLogger
 
 	priceSolver *pricesolver.SimplePriceSolver
+
+	maxBorrowableUpdateAssets map[string]struct{}
```
c9s, line 161: marginAssets

pkg/bbgo/session.go
```diff
@@ -1170,3 +1176,65 @@ func (session *ExchangeSession) FormatOrders(orders []types.SubmitOrder) (format
 
 	return formattedOrders, err
 }
+
```
c9s, line 1179: AddMarginAssets(....)

c9s (approved):

pkg/bbgo/session.go
```diff
@@ -1170,3 +1193,73 @@ func (session *ExchangeSession) FormatOrders(orders []types.SubmitOrder) (format
...
+// AddMarginAssets adds the margin assets to the session
+// IPPORTANT: After adding the margin assets, it will trigger an update on the max borrowable amount.
+// It's recommended to add callback for the update on the max borrowable amount first before calling this method.
+func (session *ExchangeSession) AddMarginAssets(
+	ctx context.Context,
+	assets ...string,
+) {
+	if session.Margin {
```
c9s, line 1204: you should use `if !session.Margin { warn }` as the function guard

pkg/bbgo/session.go
```diff
@@ -1170,3 +1193,73 @@ func (session *ExchangeSession) FormatOrders(orders []types.SubmitOrder) (format
...
+	defer t.Stop()
+	defer session.logger.Info("max borrowable updating worker exit")
+
+	ex, ok := session.Exchange.(types.MarginBorrowRepayService)
+	if !ok {
+		session.logger.Warnf("cannot start worker for max borrowable on exchange %s", session.ExchangeName)
+		return
+	}
+	session.logger.Infof("start max borrowable updating worker for %s", session.ExchangeName)
+UpdateLoop:
```
c9s, line 1237: avoid using label, it's package-wide symbol

pkg/bbgo/session.go
```diff
@@ -1170,3 +1193,73 @@ func (session *ExchangeSession) FormatOrders(orders []types.SubmitOrder) (format
...
+	for {
```
c9s, line 1238: you can place one `session.doUpdateMaxBorrowable(ctx, ex)` call before the `for` loop.

pkg/bbgo/session.go
```diff
@@ -1170,3 +1193,73 @@ func (session *ExchangeSession) FormatOrders(orders []types.SubmitOrder) (format
...
+		}
+		for _, asset := range assets {
+			session.marginAssets[asset] = struct{}{}
+		}
+		ex, ok := session.Exchange.(types.MarginBorrowRepayService)
+		if !ok {
+			session.logger.Warnf("exchange %s does not support max borrowable query", session.ExchangeName)
+			return
+		}
+		session.doUpdateMaxBorrowable(ctx, ex)
```
c9s, line 1218: can move this call to LINE 1237

pkg/bbgo/session.go
```diff
@@ -1170,3 +1193,73 @@ func (session *ExchangeSession) FormatOrders(orders []types.SubmitOrder) (format
...
+	}
+}
+
+// worker goroutine to update the max borrowable amount
+func (session *ExchangeSession) updateMaxBorrowable(
```
c9s, line 1223: call it `marginInfoUpdater()`

pkg/bbgo/session.go
```diff
@@ -1170,3 +1193,73 @@ func (session *ExchangeSession) FormatOrders(orders []types.SubmitOrder) (format
...
+	ctx context.Context,
+	interval time.Duration,
+) {
+	t := time.NewTicker(timejitter.Milliseconds(interval, 500))
+	defer t.Stop()
+	defer session.logger.Info("max borrowable updating worker exit")
```
c9s, line 1229: this is more like a debug message

pkg/bbgo/testdata/strategy.yaml
```diff
@@ -5,11 +5,13 @@ sessions:
     envVarPrefix: MAX
     takerFeeRate: 0
     makerFeeRate: 0
+    maxBorrowableUpdateInterval: 3m
```
c9s, line 8: call it `marginInfoUpdaterInterval`

pkg/bbgo/session.go
```diff
@@ -1170,3 +1193,73 @@ func (session *ExchangeSession) FormatOrders(orders []types.SubmitOrder) (format
...
+		select {
+		case <-ctx.Done():
+			break UpdateLoop
+		case <-t.C:
+			session.doUpdateMaxBorrowable(ctx, ex)
+		}
+	}
+}
+
+func (session *ExchangeSession) doUpdateMaxBorrowable(
```
c9s, line 1248: you don't have to prefix "do", only if it conflicts with an existing name

pkg/bbgo/session.go
```diff
@@ -122,6 +125,9 @@ type ExchangeSession struct {
 
 	UseHeikinAshi bool `json:"heikinAshi,omitempty" yaml:"heikinAshi,omitempty"`
 
+	// Margin Assets Configs
+	MaxBorrowableUpdateInterval string `json:"maxBorrowableUpdateInterval" yaml:"maxBorrowableUpdateInterval"`
```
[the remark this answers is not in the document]
dboyliao: But it making it hard to pass the `config_test.go`.

pkg/bbgo/session.go
```diff
@@ -1170,3 +1193,74 @@ func (session *ExchangeSession) FormatOrders(orders []types.SubmitOrder) (format
...
+}
+
+// worker goroutine to update the max borrowable amount
+func (session *ExchangeSession) marginInfoUpdater(
+	ctx context.Context,
+	interval time.Duration,
+) {
+	t := time.NewTicker(timejitter.Milliseconds(interval, 500))
+	defer t.Stop()
+	defer session.logger.Debug("max borrowable updating worker exit")
```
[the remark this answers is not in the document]
dboyliao: I refactorred it.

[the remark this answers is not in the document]
dboyliao: I wrap the margin update logics into a new private struct, `marginInfoUpdaterType`.
Conceptually, `marginInfoUpdaterType` is private to exchange session.
APIs of `marginInfoUpdaterType` are exposed via exchange session.

pkg/bbgo/margin_info_updater.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/c9s/bbgo/pkg/fixedpoint"
+	"github.com/c9s/bbgo/pkg/types"
+	"github.com/c9s/bbgo/pkg/util/timejitter"
+)
+
+type MaxBorrowableCallbackType func(asset string, amount fixedpoint.Value)
+
+//go:generate callbackgen -type MarginInfoUpdater
+type marginInfoUpdaterType struct {
```
c9s, line 15: should be MarginInfoUpdater?

pkg/bbgo/margin_info_updater.go
```diff
@@ -0,0 +1,91 @@
...
+	BorrowRepayService types.MarginBorrowRepayService
+
+	assets                 map[string]struct{}
+	maxBorrowableCallbacks []MaxBorrowableCallbackType
+}
+
+func NewMarginInfoUpdaterFromSession(
+	session *ExchangeSession,
```
c9s, line 23: you only need session.Exchange for this MarginInfoUpdater, the margin flag checking can be done outside of this component

pkg/bbgo/margin_info_updater.go
```diff
@@ -0,0 +1,91 @@
...
+) *marginInfoUpdaterType {
+	if !session.Margin || session.Exchange == nil {
+		return nil
+	}
```
c9s, line 27: ditto

pkg/bbgo/session.go
```diff
@@ -465,6 +473,22 @@ func (session *ExchangeSession) Init(ctx context.Context, environ *Environment)
...
 
+	// session-wide max borrowable updating worker
+	if session.MarginInfoUpdaterInterval == "" {
+		session.MarginInfoUpdaterInterval = "30m"
+	}
+	if session.marginInfoUpdater == nil {
+		session.marginInfoUpdater = NewMarginInfoUpdaterFromSession(session)
+	}
+	if session.marginInfoUpdater != nil {
+		interval, err := time.ParseDuration(session.MarginInfoUpdaterInterval)
```
c9s, line 484: duration parsing can be done by using types.Duration

pkg/bbgo/margin_info_updater.go
dboyliao: I want it to be private.
And ideally expose it via `NewMarginInfoUpdaterFromSession` or any other constructor if necessary.

pkg/bbgo/margin_info_updater.go
dboyliao: The symantic for `NewMarginInfoUpdaterFromSession` is that it is the constructor of the private type `marginInfoUpdaterType`.
It will try to construct the struct and return `nil` if it's not viable.

The caller only need to check if the value returned is `nil` or not.

pkg/bbgo/margin_info_updater.go
dboyliao: Just like what I reply above, I'd rather keep the check here.

pkg/bbgo/session.go
```diff
@@ -465,6 +473,22 @@ func (session *ExchangeSession) Init(ctx context.Context, environ *Environment)
...
 	})
 
+	// session-wide max borrowable updating worker
+	if session.MarginInfoUpdaterInterval == "" {
+		session.MarginInfoUpdaterInterval = "30m"
+	}
+	if session.marginInfoUpdater == nil {
+		session.marginInfoUpdater = NewMarginInfoUpdaterFromSession(session)
+	}
+	if session.marginInfoUpdater != nil {
```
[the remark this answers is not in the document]
dboyliao: Since `NewMarginInfoUpdaterFromSession` may return a `nil`, which represent the case where margin info updater is not viable, we need to do the check here.

pkg/bbgo/margin_info_updater.go
```diff
@@ -0,0 +1,91 @@
...
+	return &m
+}
+
+// Run starts the update workers for the given interval.
+func (m *marginInfoUpdaterType) Run(
+	ctx context.Context,
+	interval types.Duration,
+) {
+	go m.maxBorrowableUpdater(ctx, interval)
```
c9s, line 40: I’d prefer using the go keyword outside the caller—it makes it clearer that a new goroutine is being started from the session.go file.

pkg/bbgo/margin_info_updater.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/c9s/bbgo/pkg/fixedpoint"
+	"github.com/c9s/bbgo/pkg/types"
+	"github.com/c9s/bbgo/pkg/util/timejitter"
+)
+
+type MaxBorrowableCallbackType func(asset string, amount fixedpoint.Value)
```
c9s, line 12: can be just `MaxBorrowableCallback`

pkg/bbgo/session.go
```diff
@@ -207,6 +213,8 @@ func NewExchangeSession(name string, exchange types.Exchange) *ExchangeSession {
 		Session: session,
 	}
 
+	session.marginInfoUpdater = NewMarginInfoUpdaterFromSession(session)
```
c9s, line 216: add the margin flag check here:

if session.Margin {

}

make it clearer that we have a flag checking to wrap the component, don't hide the condition in the constructor.

we always assume a constructor always returns an object.

pkg/bbgo/margin_info_updater.go
dboyliao: ok, fixed.

pkg/bbgo/margin_info_updater.go
dboyliao: ok, fixed.

pkg/bbgo/margin_info_updater.go
dboyliao: fixed.

pkg/bbgo/margin_info_updater.go
dboyliao: But I think using the `go` keyword here is better.
Think of the case when we need to concurrently run more updating goroutine.
Let's say we need another updater, `maxMarginRatioUpdater`.
We only need to add `go m.maxMarginRatioUpdater()` here.

pkg/bbgo/session.go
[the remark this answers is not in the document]
dboyliao: fixed.

pkg/bbgo/margin_info_updater.go
dboyliao: done.

pkg/bbgo/margin_info_updater.go
dboyliao: fixed.

pkg/bbgo/session.go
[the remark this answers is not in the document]
dboyliao: fixed.

pkg/bbgo/session.go
dboyliao: fixed.

pkg/bbgo/testdata/strategy.yaml
dboyliao: done.

ycdesu (approved):

PR #2002, opened 2025-04-24
FEATURE: [core] Implement KLine builder, emitting kline events by market trades

dboyliao:


dboyliao:
I just be told we have `SerialMarketDataStore` in bbgo.
I will update this PR with it.
Don't review it now.

dboyliao:
It's ready for review.

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,190 @@
...
+	// klinesMap is a map from interval to accumulated kline of that interval
+	klinesMap map[types.Interval]*types.KLine
+
+	kLineCallbacks       []func(kline types.KLine)
+	kLineClosedCallbacks []func(kline types.KLine)
+	logger               *logrus.Entry
+
+	// the following fields should be protected by lock when being updated
+	mu                                                 sync.Mutex
+	open, high, low, close, volume, quoteVolume, price fixedpoint.Value
```
c9s, line 30: quote volume needs to be a float64 number, it's a huge number, especially in the BTCUSDT market

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,190 @@
...
+
+func (c *kLineBuilder) Run(ctx context.Context) {
+	if len(c.intervals) == 0 {
+		return
+	}
+	minDuration := c.minInterval.Duration()
+	for interval := range c.intervals {
+		if interval.Duration()%minDuration != 0 {
+			c.logger.Errorf("interval %s is not a multiple of minInterval %s", interval, c.minInterval)
+			panic(interval)
```
c9s, line 100: use fmt.Errorf and pass the error to panic instead of Errorf logging here

it's a logical exception (user configuration faults)

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+	}
+	c.logger.Infof("kline updater started for %s (%+v)", c.symbol, c.intervals)
+	c.updateTime = types.Time(time.Now())
```
c9s, line 104: we can't use time.Now when in replay mode (back testing), it could mess up the "real" data time

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,190 @@
...
+	if waitDuration > 0 {
+		select {
+		case <-time.After(waitDuration):
+		case <-ctx.Done():
+			return
+		}
+	}
+	// start ticker
+	intervalDuration := c.minInterval.Duration()
+	ticker := time.NewTicker(intervalDuration)
```
c9s, line 122: you need to consider back testing mode here, in the back testing mode we can't really wait fir the ticker, that's why SerialMarketDataStore was implemented in that way

pkg/core/klinebuilder.go
dboyliao: How about volume?

pkg/core/klinebuilder.go
dboyliao: ok

pkg/core/klinebuilder.go
dboyliao: I see.
I will check if it's in backtest or not when binding to a stream, just like serial data store.
And if it's in backtest mode, I won't start the goroutine.

pkg/core/klinebuilder.go
dboyliao: Oh, I know what's wrong here.
`quoteVolume` is the quote volume of the quote quantity of the last trade.
So, it is ok to use `fixedpoint.Value` here.

And due to the definition of `types.KLine`, I really can't use `float64` for the accumulated quote volume for a long interval.

dboyliao: Or we'd have to update all codes where the `types.KLine` is applied.

pkg/core/klinebuilder.go
dboyliao: Done.

`update` won't run if it's in backtest mode.

pkg/exchange/coinbase/stream_handlers.go
```diff
@@ -25,6 +26,20 @@ const (
 	balanceChannel     types.Channel = "balance"
 )
 
+// minInterval: for the kline updater
+var minInterval = types.Interval("1s")
```
bailantaotao, line 30: use const

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,207 @@
...
+		case tickTime := <-ticker.C:
+			startTime := tickTime.Add(-1 * intervalDuration).Round(intervalDuration)
+
+			// construct kline and reset converter
+			c.mu.Lock()
+			lastTradeTime := c.updateTime
+			klineMinInterval := types.KLine{
+				Symbol:    c.symbol,
+				StartTime: types.Time(startTime),
+				EndTime:   types.Time(tickTime),
```
bailantaotao, line 155: tickTime.Add(-1 millisecond)? 

The start time and end time cannot overlap

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,207 @@
...
+		if !ok {
+			// interval not in map and kline is nil
+			kline = &types.KLine{}
+			kline.Set(klineMinInterval)
+			c.klinesMap[interval] = kline
+		} else {
+			kline.Merge(klineMinInterval)
+		}
+		kline.Interval = interval
+		expectEndTime := kline.StartTime.Time().Add(interval.Duration())
```
bailantaotao, line 197: -1 time.millisecond?

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,207 @@
...
+	// start ticker
+	intervalDuration := c.minInterval.Duration()
+	ticker := time.NewTicker(intervalDuration)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case tickTime := <-ticker.C:
```
bailantaotao, line 146: is ticker frequency 1s?

pkg/exchange/coinbase/stream_handlers.go
dboyliao: In the `init()`, it will be updated according to supported intervals on Coinbase.
In this way, if there is any updates to the supported intervals on Coinbase, we won't need to change the min interval here.
It will be updated accordingly.

pkg/core/klinebuilder.go
dboyliao: No, it will be the minimum of supported interval on Coinbase.
See `init()` please.

pkg/exchange/coinbase/stream_handlers.go
```diff
@@ -129,7 +155,28 @@ func (s *Stream) handleConnect() {
...
+			for _, option := range options {
+				if !s.exchange.IsSupportedInterval(option.Interval) {
+					logStream.Errorf("unsupported interval %s for symbol %s", option.Interval, symbol)
+					panic(option.Interval)
+				}
+				logStream.Infof("subscribe for kline events of symbol %s with interval %s", symbol, option.Interval)
+				klConverter.Subscribe(option.Interval)
+			}
+			klConverter.BindStream(s)
+			klConverter.OnKLine(func(kline types.KLine) {
```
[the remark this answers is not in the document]
dboyliao: Good point.

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,207 @@
...
+	} else {
+		c.high = c.price
+	}
+	if !c.low.IsZero() {
+		if c.price.Compare(c.low) < 0 {
+			c.low = c.price
+		}
+	} else {
+		c.low = c.price
+	}
```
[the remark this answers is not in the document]
dboyliao: Nice.
Fixing it.

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+	} else {
+		c.high = c.price
+	}
+	if !c.low.IsZero() {
+		if c.price.Compare(c.low) < 0 {
+			c.low = c.price
+		}
+	} else {
+		c.low = c.price
```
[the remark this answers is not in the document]
dboyliao: `low`, `high`, ...etc will be zero when `c` is being "reset".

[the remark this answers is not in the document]
dboyliao: When the builder was reset, we should use whatever the price is as its low.

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,207 @@
...
+
+	c.price = trade.Price
+	c.close = trade.Price
+	if !c.high.IsZero() {
+		if c.price.Compare(c.high) > 0 {
+			c.high = c.price
+		}
+	} else {
+		c.high = c.price
+	}
```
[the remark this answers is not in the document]
dboyliao: Oh, yeah.
Just like my comment below, the `high` will be zero when the builder is reset.
I was thinking it's more clear to do the update with a `if` here.
But you are right, it can be simplified.

pkg/core/klinebuilder.go
dboyliao: No need to do that.
`startTime := tickTime.Add(-1 * intervalDuration).Round(intervalDuration)`
Since `.Round(...)` will round the time to it's neareast multiple of the given duration, `startTime` will not overlap with `tickTime`.

pkg/core/klinebuilder.go
dboyliao: I believe we don't need to do that here.

Ideally, `klineMinInterval` should end at a timestamp which is a multiple of `minInterval`.
`expectEndTime` here is the ideal end time of the `klineMinInterval`.

So basically, L181 is checking:
1. if the `klineMinInterval.EndTime`, which in fact is the ticker time, is after `expectEndTime`.
2. or the `lastUpdateTime`, which is the time of the last market trade, is after `expectEndTime`.

Either case, the kline should be closed for that interval.

pkg/core/klinebuilder.go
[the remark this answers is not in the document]
dboyliao: I add a comment on it for clearity.

pkg/exchange/coinbase/stream_handlers.go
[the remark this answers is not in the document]
dboyliao: Done.

pkg/exchange/coinbase/stream_handlers.go
```diff
@@ -129,7 +155,28 @@ func (s *Stream) handleConnect() {
...
+				s.EmitKLine(kline)
+			})
+			klConverter.OnKLineClosed(func(kline types.KLine) {
+				s.EmitKLineClosed(kline)
```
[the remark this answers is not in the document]
dboyliao: Done.

pkg/core/klinebuilder.go
dboyliao: Oh, I don't realize what you mean until now.
You mean the next kline start time should not overlap with the end time of the previous one right?
I will update the end time as expected end time minus one millisecond.

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,221 @@
...
+	running bool
+
+	exchange types.Exchange
+}
+
+// NewKLineBuilder: constructor of kLineBuilder
+//   - symbol: symbol to trace on
+//   - minInterval: unit interval, related to your signal timeframe.
+//     All the supported intervals of the binding stream should be multiple of this interval.
+func NewKLineBuilder(symbol string, minInterval types.Interval, isBackTesting bool) *kLineBuilder {
```
c9s, line 39: you don't need to pass isBackTesting to this builder, we just need to make this component backtest-able by ensuring not using async in the process

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,221 @@
...
+		symbol:        symbol,
+		minInterval:   minInterval,
+		isBackTesting: isBackTesting,
+		logger:        logger,
+		intervals:     make(map[types.Interval]struct{}),
+		state:         NewKLineBuilderState(symbol),
+	}
+}
+
+func (kb *kLineBuilder) Subscribe(interval types.Interval) {
```
c9s, line 54: call this Subscribe as "AddInterval" and let it return error

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,221 @@
...
+			interval,
+		)
+		kb.logger.Error(err)
+		panic(err)
+	}
+	kb.intervals[interval] = struct{}{}
+}
+
+func (kb *kLineBuilder) BindStream(stream types.Stream) {
+	if kb.isBackTesting {
```
c9s, line 68: can't handle backtest like this, would prefer pull out this logic to another layer

pkg/core/klinebuilder_state.go
```diff
@@ -0,0 +1,87 @@
...
+	if trade.Symbol != kb.symbol {
+		return
+	}
+	klineDelta := types.KLine{
+		Symbol:      trade.Symbol,
+		Open:        trade.Price,
+		Close:       trade.Price,
+		High:        trade.Price,
+		Low:         trade.Price,
+		EndTime:     types.Time(time.Now()), // use current time to fix the timezone issue (temporary)
```
c9s, line 63: in the back test mode, we don't have use "time.Now()" because it's a simulation

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,221 @@
...
+				kline.Set(lastKline)
+				kline.StartTime = oriStarTime
+				okForEmit = true
+			} else if kb.exchange != nil {
+				// try to query kline from RESTful API
+				queryStartTime := kline.StartTime.Time().Add(-kline.Interval.Duration())
+				queryEndTime := queryStartTime.Add(2 * kline.Interval.Duration())
+				ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+				defer cancel()
+				klines, err := kb.exchange.QueryKLines(
```
c9s, line 167: we can't call QueryKlines in the back-test mode, and it seems like you're mixing 2 difference scenario in the same component?

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,221 @@
...
+			if found {
+				kline.Set(lastKline)
+				kline.StartTime = oriStarTime
+				okForEmit = true
+			} else if kb.exchange != nil {
+				// try to query kline from RESTful API
+				queryStartTime := kline.StartTime.Time().Add(-kline.Interval.Duration())
+				queryEndTime := queryStartTime.Add(2 * kline.Interval.Duration())
+				ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+				defer cancel()
```
c9s, line 166: be careful of the defer, it will not be called at the end of loop, it's executed at the end of function call.

pkg/core/klinebuilder.go
```diff
@@ -0,0 +1,221 @@
...
+	kb.mu.Lock()
+	defer kb.mu.Unlock()
+
+	kb.state.AddTrade(trade)
+}
+
+func (kb *kLineBuilder) run(ctx context.Context) {
+	startTime := time.Now()
+	startTimeTruncated := startTime.Truncate(kb.minInterval.Duration())
+	intervalDuration := kb.minInterval.Duration()
```
c9s, line 107: since minInterval's purpose is different (for the ticker), we can rename it to tickerInterval, and ensures all the intervals we used are greater than the ticker interval.

pkg/core/klinebuilder_state.go
```diff
@@ -0,0 +1,87 @@
...
+
+func NewKLineBuilderState(symbol string) *kLineBuilderState {
+	return &kLineBuilderState{
+		symbol:        symbol,
+		lastKLinesMap: make(map[types.Interval]*types.KLine),
+		kLinesMap:     make(map[types.Interval]*types.KLine),
+	}
+}
+
+// Reset will reset the kline of the given interval to a new open kline with the given start time
```
c9s, line 26: doc comment uses simple form, "Reset resets"

pkg/core/klinebuilder_state.go
```diff
@@ -0,0 +1,87 @@
...
+		Volume:      trade.Quantity,
+		QuoteVolume: trade.QuoteQuantity,
+	}
+	for _, oldKline := range kb.kLinesMap {
+		updateKlineInPlace(oldKline, &klineDelta)
+	}
+
+}
+
+func updateKlineInPlace(existing, delta *types.KLine) {
```
c9s, line 73: why not just pass types.Trade ? because you only need the price and the volume

pkg/core/klinebuilder_state.go
```diff
@@ -0,0 +1,87 @@
...
+	kline, ok := kb.kLinesMap[interval]
+	return kline, ok
+}
+
+func (kb *kLineBuilderState) GetLastKLine(interval types.Interval) (*types.KLine, bool) {
+	lastKLine, ok := kb.lastKLinesMap[interval]
+	return lastKLine, ok
+}
+
+func (kb *kLineBuilderState) AddTrade(trade types.Trade) {
```
c9s, line 50: when AddTrade is called, you can return the updated klines to the caller, and these updated klines can be pushed to the stream

c9s: e.g. 
```
itv1 --------|---itv2-------|---itv3
td1-td2-----|---td3-------|---td4
```
itv* is the ticker interval, not the kline interval, the ticker interval should be smaller than all the kline intervals, which is for pushing the updated kline periodically.

when td3 is received, we need to return "closed k1" and "new k2"

dboyliao:
Close this PR in favor of #2024.

PR #2007, opened 2025-04-27
FEATURE: [okx] Support updating the local cache of symbols when calling QueryMarkets

anywhy:


pkg/exchange/okex/convert.go
```diff
@@ -16,7 +16,7 @@ func toGlobalSymbol(symbol string) string {
 }
 
 //go:generate go run gensymbols.go
-func toLocalSymbol(symbol string, instType ...okexapi.InstrumentType) string {
+func ToLocalSymbol(symbol string, instType ...okexapi.InstrumentType) string {
```
c9s, line 19: please avoid export converter function since it's only used in the okx package

anywhy: Got it — if we don’t export the ``converter`` method, we won’t be able to call OKX-specific functions like ``SetAccountLeverage`` inside the strategy, since we need to convert from the global symbol to the local symbol first.
So, could we maybe add this kind of special function directly to the Exchange instead?

c9s: yes, you can add public methods on Exchange, and in your strategy, you cast the exchange interface into *okx.Exchange to use these methods

c9s (approved):

bailantaotao (approved):

PR #2009, opened 2025-04-28
FIX: [binance] apply new ws endpoint and use ed25519 authentication for spot trading

dboyliao:
`listenKey` is deprecated:
https://developers.binance.com/docs/binance-spot-api-docs/CHANGELOG#user-data-streams

### DIFF
--- a/pkg/exchange/binance/binanceapi/client.go
+++ b/pkg/exchange/binance/binanceapi/client.go
@@ -172,17 +172,12 @@ func (c *RestClient) NewAuthenticatedRequest(
 func (c *RestClient) newAuthenticatedRequest(
 	ctx context.Context, method, refURL string, params url.Values, payload interface{}, signFunc func(string) string,
 ) (*http.Request, error) {
-	if c.IsUsingEd25519Auth() {
-		// Ed25519 authentication
-		if len(c.PrivateKey) == 0 {
-			return nil, errEmptyPrivateKey
-		}
-	} else {
-		// HMAC authentication
-		if len(c.Key) == 0 {
-			return nil, errNoApiKey
-		}
+	if len(c.Key) == 0 {
+		return nil, errNoApiKey
+	}
 
+	if !c.IsUsingEd25519Auth() {
+		// using HMAC authentication
 		if len(c.Secret) == 0 {
 			return nil, errNoApiSecret
 		}
--- a/pkg/exchange/binance/binanceapi/client.go
+++ b/pkg/exchange/binance/binanceapi/client.go
@@ -94,7 +94,7 @@ func (c *RestClient) Auth(key, secret string, privateKey ed25519.PrivateKey) {
 }
 
 func (c *RestClient) IsUsingEd25519Auth() bool {
-	return c.PrivateKey == nil
+	return len(c.PrivateKey) > 0
 }
 
 // NewRequest create new API request. Relative url can be provided in refURL.
--- a/pkg/exchange/binance/binanceapi/client.go
+++ b/pkg/exchange/binance/binanceapi/client.go
@@ -3,6 +3,7 @@ package binanceapi
 import (
 	"bytes"
 	"context"
+	"crypto/ed25519"
 	"crypto/hmac"
 	"crypto/sha256"
 	"crypto/tls"
@@ -52,6 +53,7 @@ type RestClient struct {
 	requestgen.BaseAPIClient
 
 	Key, Secret string
+	PrivateKey  ed25519.PrivateKey
 
 	recvWindow int
 	timeOffset int64
@@ -78,10 +80,15 @@ func NewClient(baseURL string) *RestClient {
 	return client
 }
 
-func (c *RestClient) Auth(key, secret string) {
+func (c *RestClient) Auth(key, secret string, privateKey ed25519.PrivateKey) {
 	c.Key = key
 	// pragma: allowlist nextline secret
 	c.Secret = secret
+	c.PrivateKey = privateKey
+}
+
+func (c *RestClient) Ed25519Auth() bool {
+	return c.PrivateKey == nil
 }
 
 // NewRequest create new API request. Relative url can be provided in refURL.
@@ -141,12 +148,31 @@ func (c *RestClient) SendRequest(req *http.Request) (*requestgen.Response, error
 
 // newAuthenticatedRequest creates new http request for authenticated routes.
 func (c *RestClient) NewAuthenticatedRequest(ctx context.Context, method, refURL string, params url.Values, payload interface{}) (*http.Request, error) {
-	if len(c.Key) == 0 {
-		return nil, errors.New("empty api key")
+	signFunc := c.signHMAC
+	if c.Ed25519Auth() {
+		signFunc = c.signEd25519
 	}
+	return c.newAuthenticatedRequest(ctx, method, refURL, params, payload, signFunc)
+}
 
-	if len(c.Secret) == 0 {
-		return nil, errors.New("empty api secret")
+// newAuthenticatedRequest creates new http request for authenticated routes.
+// See the doc links for details:
+// - https://developers.binance.com/docs/binance-spot-api-docs/rest-api/endpoint-security-type
+// - https://developers.binance.com/docs/binance-spot-api-docs/rest-api/general-information-on-endpoints
+func (c *RestClient) newAuthenticatedRequest(ctx context.Context, method, refURL string, params url.Values, payload interface{}, signFunc func(string) string) (*http.Request, error) {
+	if c.Ed25519Auth() {
+		// Ed25519 authentication
+		if len(c.PrivateKey) == 0 {
+			return nil, errors.New("empty private key")
+		}
+	} else {
+		// HMAC authentication
+		if len(c.Key) == 0 {
+			return nil, errors.New("empty api key")
+		}
+		if len(c.Secret) == 0 {
+			return nil, errors.New("empty api secret")
+		}
 	}
 
 	rel, err := url.Parse(refURL)
@@ -172,7 +198,7 @@ func (c *RestClient) NewAuthenticatedRequest(ctx context.Context, method, refURL
 	}
 
 	toSign := rawQuery + string(body)
-	signature := sign(c.Secret, toSign)
+	signature := signFunc(toSign)
 
 	// sv is the extra url parameters that we need to attach to the request
 	sv := url.Values{}
@@ -204,15 +230,20 @@ func (c *RestClient) NewAuthenticatedRequest(ctx context.Context, method, refURL
 	return req, nil
 }
 
-// sign uses sha256 to sign the payload with the given secret
-func sign(secret, payload string) string {
-	var sig = hmac.New(sha256.New, []byte(secret))
+// signHMAC uses sha256 to sign the payload with the given secret
+func (c *RestClient) signHMAC(payload string) string {
+	var sig = hmac.New(sha256.New, []byte(c.Secret))
 	_, err := sig.Write([]byte(payload))
 	if err != nil {
 		return ""
 	}
 
 	return fmt.Sprintf("%x", sig.Sum(nil))
+}
+
+// signEd25519 uses ed25519 to sign the payload with the given private key
+func (c *RestClient) signEd25519(payload string) string {
+	return GenerateSignatureEd25519(payload, c.PrivateKey)
 }
 
 func currentTimestamp() int64 {
--- a/pkg/exchange/binance/binanceapi/client.go
+++ b/pkg/exchange/binance/binanceapi/client.go
@@ -27,6 +27,12 @@ const RestBaseURL = "https://api.binance.com"
 const SandboxRestBaseURL = "https://testnet.binance.vision"
 const DebugRequestResponse = false
 
+var errEmptyPrivateKey = errors.New("empty private key")
+
+var errNoApiKey = errors.New("empty api key")
+
+var errNoApiSecret = errors.New("empty api secret")
+
 var dialer = &net.Dialer{
 	Timeout:   30 * time.Second,
 	KeepAlive: 30 * time.Second,
@@ -87,12 +93,14 @@ func (c *RestClient) Auth(key, secret string, privateKey ed25519.PrivateKey) {
 	c.PrivateKey = privateKey
 }
 
-func (c *RestClient) Ed25519Auth() bool {
+func (c *RestClient) IsUsingEd25519Auth() bool {
 	return c.PrivateKey == nil
 }
 
 // NewRequest create new API request. Relative url can be provided in refURL.
-func (c *RestClient) NewRequest(ctx context.Context, method, refURL string, params url.Values, payload interface{}) (*http.Request, error) {
+func (c *RestClient) NewRequest(
+	ctx context.Context, method, refURL string, params url.Values, payload interface{},
+) (*http.Request, error) {
 	rel, err := url.Parse(refURL)
 	if err != nil {
 		return nil, err
@@ -147,9 +155,11 @@ func (c *RestClient) SendRequest(req *http.Request) (*requestgen.Response, error
 }
 
 // newAuthenticatedRequest creates new http request for authenticated routes.
-func (c *RestClient) NewAuthenticatedRequest(ctx context.Context, method, refURL string, params url.Values, payload interface{}) (*http.Request, error) {
+func (c *RestClient) NewAuthenticatedRequest(
+	ctx context.Context, method, refURL string, params url.Values, payload interface{},
+) (*http.Request, error) {
 	signFunc := c.signHMAC
-	if c.Ed25519Auth() {
+	if c.IsUsingEd25519Auth() {
 		signFunc = c.signEd25519
 	}
 	return c.newAuthenticatedRequest(ctx, method, refURL, params, payload, signFunc)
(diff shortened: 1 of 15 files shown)


pkg/exchange/binance/stream.go
```diff
@@ -80,6 +94,12 @@ func NewStream(ex *Exchange, client *binance.Client, futuresClient *futures.Clie
 		futuresClient:  futuresClient,
 		depthBuffers:   make(map[string]*depth.Buffer),
 	}
+	var privatePEM string = os.Getenv("BINANCE_STREAM_PRIVATE_PEM")
```
c9s, line 97: need to use prefix `BINANCE_API_*`, so maybe call it `BINANCE_API_PRIVATE_KEY_PEM`

pkg/exchange/binance/stream.go
```diff
@@ -80,6 +94,12 @@ func NewStream(ex *Exchange, client *binance.Client, futuresClient *futures.Clie
...
+	privateKey, err := prasePrivateKey(privatePEM)
```
c9s, line 98: parse

pkg/exchange/binance/stream.go
c9s, line 97: btw you should load the private key pem in the Exchange constructor, in order to handle fallback to HMAC

pkg/exchange/binance/stream.go
```diff
@@ -132,12 +152,28 @@ func NewStream(ex *Exchange, client *binance.Client, futuresClient *futures.Clie
 
 	stream.OnDisconnect(stream.handleDisconnect)
 	stream.OnConnect(stream.handleConnect)
-	stream.OnListenKeyExpired(func(e *ListenKeyExpired) {
-		stream.Reconnect()
-	})
 	return stream
 }
 
+func prasePrivateKey(pemString string) (ed25519.PrivateKey, error) {
```
c9s, line 158: parse

pkg/exchange/binance/stream.go
```diff
@@ -132,12 +152,28 @@ func NewStream(ex *Exchange, client *binance.Client, futuresClient *futures.Clie
...
+	block, _ := pem.Decode([]byte(pemString))
```
c9s, line 159: error is not checked

pkg/exchange/binance/stream.go
```diff
@@ -132,12 +152,28 @@ func NewStream(ex *Exchange, client *binance.Client, futuresClient *futures.Clie
...
+	}
+
+	// Parse the private key
+	privateKeyInterface, err := x509.ParsePKCS8PrivateKey(block.Bytes)
+	if err != nil {
+		return nil, err
+	}
+	privateKey, ok := privateKeyInterface.(ed25519.PrivateKey)
+	if !ok {
+		err := errors.New("Not an ED25519 private key")
```
c9s, line 171: fmt.Errorf("Not an ED25519 private key: %T given", privateKeyInterface)

pkg/exchange/binance/stream.go
```diff
@@ -146,36 +182,61 @@ func (s *Stream) handleDisconnect() {
...
-		ID:     1,
-	})
+	err := s.Conn.WriteJSON(cmd)
 
 	if err != nil {
 		log.WithError(err).Error("subscribe error")
 	}
 }
 
+func (s *Stream) generateSignature() (string, int64) {
```
c9s, line 231: I guess you need to pull out this method as a pure function that takes one parameter -- the private key, because you will need to use it later for the RESTful API client.

dboyliao: Yes.
I didn't think of backward compatibility.
I'll revise the PR as an whole.

pkg/exchange/binance/stream.go
dboyliao: The second return value of `pem.Decode` is not `error`.

dboyliao: It's the resting bytes not yet decoded.

pkg/exchange/binance/stream.go
dboyliao: done.

pkg/exchange/binance/stream.go
c9s: got it~

pkg/exchange/binance/stream.go
```diff
@@ -146,13 +192,26 @@ func (s *Stream) handleDisconnect() {
...
 func (s *Stream) handleConnect() {
+	var err error
+	if len(s.privateKey) == 0 {
+		log.Warnf("Ed25519 private key is not set, using deprecated HMAC authentication")
+		err = s.handleConnectHMAC()
+	} else {
+		err = s.handleConnectED25519()
+	}
+	if err != nil {
+		log.WithError(err).Error("subscribe error")
```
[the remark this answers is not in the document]
dboyliao: I'll handle the authentication in the body of `handleConnectHMAC` and `handleConnectED25519`.
In both of them, I'll skip authentication if it's a public data stream.

pkg/exchange/binance/stream.go
dboyliao: I move it to `pkg/exchange/coinbase/util.go` so the RESTful API can share this function with the stream.

pkg/exchange/binance/stream.go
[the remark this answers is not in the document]
dboyliao: Ok, I think I will rename them as `authConnectHMAC` and `authConnectEd25519` for clarity.
And skip auth if it's public stream.

pkg/exchange/binance/exchange.go
```diff
@@ -101,6 +104,12 @@ func New(key, secret string) *Exchange {
 	if util.IsPaperTrade() {
 		binance.UseTestnet = true
 	}
+
+	var privatePEM string = os.Getenv("BINANCE_API_PRIVATE_KEY_PEM")
+	privateKey, err := util.ParseEd25519PrivateKey(privatePEM)
+	if err != nil {
+		panic(err)
+	}
```
c9s, line 112: you might need a flag "ed25519Auth bool" something like this

pkg/exchange/binance/stream.go
```diff
@@ -41,6 +52,7 @@ type Stream struct {
 
 	client        *binance.Client
 	futuresClient *futures.Client
+	privateKey    ed25519.PrivateKey
```
c9s, line 55: you can read the privateKey field from the given exchange object directly since it's in the same package

c9s: you can also add an accessor like "getEd25519PrivateKey" method on the exchange struct

pkg/exchange/binance/stream.go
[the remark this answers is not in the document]
c9s: authConnectHMAC and authConnectEd25519 sounds good

pkg/exchange/binance/stream.go
```diff
@@ -146,34 +159,57 @@ func (s *Stream) handleDisconnect() {
...
 	}
+}
+
+func (s *Stream) authConnectHMAC() error {
+	go s.EmitAuth()
+	return nil
+}
 
+func (s *Stream) authConnectEd25519() error {
+	signature, timestamp := generateSignature(s.client.APIKey, s.privateKey)
```
c9s, line 183: you might need to move this function to the binanceapi package later for reuse

pkg/exchange/binance/util.go
```diff
@@ -0,0 +1,17 @@
+package binance
+
+import (
+	"crypto/ed25519"
+	"encoding/base64"
+	"fmt"
+	"time"
+)
+
+func generateSignature(apiKey string, privateKey ed25519.PrivateKey) (string, int64) {
```
c9s, line 10: extract time.Now() as a function parameter, so that you can add a test against this function

pkg/util/ed25519.go
```diff
@@ -0,0 +1,31 @@
...
+	"crypto/ed25519"
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
+	"fmt"
+)
+
+func ParseEd25519PrivateKey(pemString string) (ed25519.PrivateKey, error) {
+	if len(pemString) == 0 {
+		return nil, nil
```
c9s, line 13: from an utility function's perspective, empty pem string is unexpected, you need to return an error

pkg/util/ed25519.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+	block, _ := pem.Decode([]byte(pemString))
+	if block == nil {
+		return nil, errors.New("failed to parse PEM block containing the private key")
+	}
+
+	// Parse the private key
+	privateKeyInterface, err := x509.ParsePKCS8PrivateKey(block.Bytes)
```
c9s, line 21: we seldom use the "Interface" in the variable name in golang, you can call it "pkcs8Key" here or just "privateKey"

pkg/exchange/binance/stream.go
dboyliao: I already move the `generateSignature` to `pkg/exchange/binance/util.go`.
The RESTful client should have access to it.

pkg/util/ed25519.go
dboyliao: How about `anyPrivateKey`?

pkg/exchange/binance/stream.go
dboyliao: Done.

pkg/exchange/binance/stream.go
dboyliao: Oh, the `binanceapi` package.
I see.
Will do.

dboyliao: Done.

pkg/exchange/binance/exchange.go
```diff
@@ -150,6 +174,14 @@ func New(key, secret string) *Exchange {
 	return ex
 }
 
+func (e *Exchange) GetEd25519PrivateKey() ed25519.PrivateKey {
+	return e.ed25519PrivateKey
+}
+
+func (e *Exchange) GetEd25519Auth() bool {
```
c9s, line 181: ditto, it's for the stream, they are in the same package, so you don't have to export them as public methods, it could be private methods

pkg/exchange/binance/stream.go
```diff
@@ -146,34 +163,63 @@ func (s *Stream) handleDisconnect() {
...
+}
 
-		// spawn a goroutine to emit auth event to prevent blocking the main event loop
-		go s.EmitAuth()
-		return
+func (s *Stream) authConnectEd25519() error {
+	timestamp := time.Now().UnixNano() / 1e6
+	params := map[string]string{
+		"apiKey":    s.client.APIKey,
+		"timestamp": fmt.Sprintf("%d", timestamp),
```
c9s, line 191: always prefer using strconv.FormatInt instead of Sprintf when you can use trconv.FormatInt

pkg/exchange/binance/stream.go
```diff
@@ -146,34 +163,63 @@ func (s *Stream) handleDisconnect() {
...
 	}
+	signature := api.GenerateSignatureEd25519(params, s.privateKey)
+	return s.Conn.WriteJSON(LoginCommand{
+		WebSocketCommand: WebSocketCommand{
+			Method: "session.login",
+			ID:     1,
```
c9s, line 197: can ID be reused ?

pkg/exchange/binance/stream.go
```diff
@@ -146,34 +163,63 @@ func (s *Stream) handleDisconnect() {
...
+		},
+		Params: LoginParams{
+			APIKey:    s.client.APIKey,
+			Signature: signature,
+			Timestamp: timestamp,
+		},
+	})
+}
 
+func (s *Stream) subscribe() error {
```
c9s, line 207: maybe name it `sendSubscribe` or `writeSubscribe`

pkg/util/ed25519.go
```diff
@@ -0,0 +1,31 @@
...
+
+	// Parse the private key
+	pkcs8Key, err := x509.ParsePKCS8PrivateKey(block.Bytes)
+	if err != nil {
+		return nil, err
+	}
+	privateKey, ok := pkcs8Key.(ed25519.PrivateKey)
+	if !ok {
+		err := fmt.Errorf("not an ED25519 private key: %T given", pkcs8Key)
+		return nil, err
```
c9s, line 28: you can:

```
return nil, fmt.Errorf("not an ED25519 private key: %T given", pkcs8Key)
```

pkg/exchange/binance/stream.go
dboyliao: Why? Is there a performance or security issue on it?

pkg/exchange/binance/binanceapi/util.go
```diff
@@ -0,0 +1,20 @@
+package binanceapi
+
+import (
+	"crypto/ed25519"
+	"encoding/base64"
+	"fmt"
+	"strings"
+)
+
+func GenerateSignatureEd25519(params map[string]string, privateKey ed25519.PrivateKey) string {
```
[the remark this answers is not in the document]
dboyliao: Ok.
I think I can change it as `[]string` containing values like `"k=v"`.

pkg/exchange/binance/stream.go
dboyliao: According to the doc, it seems to be ok if we fill in random number:
https://developers.binance.com/docs/binance-spot-api-docs/websocket-api/request-format

截圖 2025-04-29 下午5 41 32

pkg/exchange/binance/stream.go
dboyliao: I have a better idea, how about `subscribeMarketDataSource`?
Since `subscribe` is only used for market data stream, naming it this way make it more clear in my opinion.

pkg/exchange/binance/binanceapi/util.go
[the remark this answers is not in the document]
dboyliao: Done.

pkg/util/ed25519.go
dboyliao: done.

pkg/exchange/binance/stream.go
dboyliao: Ok, `writeSubscriptions` it is!

c9s (approved):

pkg/exchange/binance/binanceapi/client.go
```diff
@@ -140,13 +155,37 @@ func (c *RestClient) SendRequest(req *http.Request) (*requestgen.Response, error
...
+// See the doc links for details:
+// - https://developers.binance.com/docs/binance-spot-api-docs/rest-api/endpoint-security-type
+// - https://developers.binance.com/docs/binance-spot-api-docs/rest-api/general-information-on-endpoints
+func (c *RestClient) newAuthenticatedRequest(
+	ctx context.Context, method, refURL string, params url.Values, payload interface{}, signFunc func(string) string,
+) (*http.Request, error) {
+	if c.IsUsingEd25519Auth() {
+		// Ed25519 authentication
+		if len(c.PrivateKey) == 0 {
+			return nil, errEmptyPrivateKey
```
gx578007, line 178: It should be impossible to happen while you have passed `IsUsingEd25519Auth`.

pkg/exchange/binance/exchange.go
```diff
@@ -35,11 +35,22 @@ const DefaultDepthLimit = 5000
 const BinanceUSBaseURL = "https://api.binance.us"
 const BinanceTestBaseURL = "https://testnet.binance.vision"
 const BinanceUSWebSocketURL = "wss://stream.binance.us:9443"
+
 const WebSocketURL = "wss://stream.binance.com:9443"
-const WebSocketTestURL = "wss://testnet.binance.vision"
-const FutureTestBaseURL = "https://testnet.binancefuture.com"
 const FuturesWebSocketURL = "wss://fstream.binance.com"
-const FuturesWebSocketTestURL = "wss://stream.binancefuture.com"
+const TestNetFuturesWebSocketURL = "wss://stream.binancefuture.com"
```
dboyliao, line 41: This one does not look like testnet url.
Is it on purpose or a typo?

pkg/exchange/binance/exchange.go
```diff
@@ -67,21 +91,37 @@ func init() {
...
+	privateKey ed25519.PrivateKey
+
+	usingEd25519 bool
+}
+
+func (e *ed25519authentication) getEd25519PrivateKey() ed25519.PrivateKey {
+	return e.privateKey
+}
+
+func (e *ed25519authentication) isUsingEd25519Auth() bool {
```
dboyliao, line 113: ditto.

pkg/exchange/binance/exchange.go
c9s: It's just a rename from the above line<

pkg/exchange/binance/exchange.go
```diff
@@ -67,21 +91,37 @@ func init() {
...
 	return ok && v
 }
 
+type ed25519authentication struct {
+	privateKey ed25519.PrivateKey
+
+	usingEd25519 bool
+}
+
+func (e *ed25519authentication) getEd25519PrivateKey() ed25519.PrivateKey {
```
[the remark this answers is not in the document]
c9s: yeah we can remove it

pkg/exchange/binance/binanceapi/client.go
c9s: I just updated the validations

pkg/exchange/binance/exchange.go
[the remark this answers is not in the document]
c9s: removed

pkg/exchange/binance/exchange.go
c9s: removed

gx578007 (approved):

PR #2014, opened 2025-04-30
FIX:[coinbase] convert private symbols to local symbols when connecting user data stream

dboyliao:


ycdesu (approved):
It will be nice to have a unit test for this case

gx578007 (approved):

PR #2017, opened 2025-05-05
REFACTOR: (coinbase) wrapping connection logics into methods for better maintainability.

dboyliao:
In this PR, I refactor the `handleConnect` method of the coinbase `Stream` object as following:
1. wrapping connection logics of user and market data stream into two separate methods:
    - `buildMapForUserDataStream`
    - `buildMapForMarketStream`
2. `writeSubscribeJson`: method that write the subscription JSON to the websocket connection
3. Instead of just logging on subscription error, use `panic` to abort the process
    - the rationale is that we should prevent the process running since user/market data streams are important data feeds for the trading strategies. If there is any error on subscriptions, the stream is compromised and should be explicitly stopped instead of just logging a message.

### DIFF
--- a/pkg/exchange/coinbase/stream_handlers.go
+++ b/pkg/exchange/coinbase/stream_handlers.go
@@ -77,206 +77,16 @@ func (msg subscribeMsgType2) String() string {
 }
 
 func (s *Stream) handleConnect() {
-	subProductsMap := make(map[types.Channel][]string)
+	// channel2LocalSymbolsMap is a map from channel to local symbols
+	channel2LocalSymbolsMap := make(map[types.Channel][]string)
 
 	// user data strea, subscribe to user channel for the user order/trade updates
 	if !s.PublicOnly {
-		if !s.authEnabled {
-			panic("user channel requires authentication")
-		}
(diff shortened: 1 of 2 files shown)


kbearXD (approved):

pkg/exchange/coinbase/stream_handlers.go
```diff
@@ -77,67 +77,131 @@ func (msg subscribeMsgType2) String() string {
 }
 
 func (s *Stream) handleConnect() {
-	subProductsMap := make(map[types.Channel][]string)
+	// channel2LocalSymbolsMap is a map from channel to local symbols
+	channel2LocalSymbolsMap := make(map[types.Channel][]string)
 
 	// user data strea, subscribe to user channel for the user order/trade updates
```
gx578007, line 83: Please also help fix the typo in the comment: strea"m".

dboyliao: Opps, thanks.

dboyliao: Done.

gx578007 (approved):

PR #2019, opened 2025-05-05
REFACTOR: [types] move `MarketDataStore`/`SerialMarketDataStore` to `types` module

dboyliao:


pkg/types/serialmarketdatastore.go
```diff
@@ -43,7 +44,7 @@ func (store *SerialMarketDataStore) Subscribe(interval types.Interval) {
 	store.Subscription = append(store.Subscription, interval)
 }
 
-func (store *SerialMarketDataStore) BindStream(ctx context.Context, stream types.Stream) {
+func (store *SerialMarketDataStore) BindStream(ctx context.Context, stream Stream) {
 	if store.UseMarketTrade {
 		if IsBackTesting {
```
c9s, line 49: perhaps remove this condition to simplify the package import

pkg/bbgo/environment.go
```diff
@@ -49,6 +49,7 @@ var BackTestService *service.BacktestService
 func SetBackTesting(s *service.BacktestService) {
 	BackTestService = s
 	IsBackTesting = s != nil
+	types.IsBackTesting = IsBackTesting
```
c9s, line 52: so let's remove this line

c9s (approved):

pkg/types/serialmarketdatastore.go
dboyliao: I'll comment out this first.
Let's see if we need to add it back in the future.

pkg/bbgo/environment.go
dboyliao: ok

ycdesu (approved):

PR #2022, opened 2025-05-06
FEATURE: [okx] support okx futures channel subscribe

anywhy:


c9s:
Hi @anywhy could you please also take a look at this comment? thanks
https://github.com/c9s/bbgo/pull/1902#discussion_r2052158428

pkg/exchange/okex/kline_stream.go
```diff
@@ -32,14 +34,24 @@ func NewKLineStream() *KLineStream {
...
+	s.isFutures = true
+}
+
 func (s *KLineStream) handleConnect() {
 	var subs []WebsocketSubscription
+
+	instType := okexapi.InstrumentTypeSpot
+	if s.isFutures {
+		instType = okexapi.InstrumentTypeSwap
+	}
```
c9s, line 47: see my PR to call the new method to get the instrument type
https://github.com/c9s/bbgo/pull/2023/files#diff-14a80c286a74b4b35bf3fa52f7e522f1db4e9c7f13ce1706b64b95547b193e5eR805

pkg/exchange/okex/stream.go
```diff
@@ -79,17 +79,18 @@ func NewStream(client *okexapi.RestClient, balanceProvider types.ExchangeAccount
 
 func (s *Stream) UseFutures() {
 	s.isFutures = true
```
c9s, line 81: you can get the FuturesSettings from the exchange instance

anywhy: got it, I'll give it a try

c9s (approved):

pkg/exchange/okex/stream_test.go
```diff
@@ -25,6 +26,9 @@ func getTestClientOrSkip(t *testing.T) *Stream {
 	}
 
 	exchange := New(key, secret, passphrase)
+	if len(isFutures) == 1 {
```
c9s, line 29: can you assign the field from the caller?

pkg/exchange/okex/stream_test.go
```diff
@@ -25,6 +26,9 @@ func getTestClientOrSkip(t *testing.T) *Stream {
...
+		exchange.IsFutures = isFutures[0]
+	}
 	return NewStream(exchange.client, exchange)
```
c9s, line 32: let's allocate stream from the caller

anywhy: done

pkg/exchange/okex/kline_stream.go
```diff
@@ -11,12 +11,15 @@ import (
 type KLineStream struct {
 	types.StandardStream
 
+	exchange types.Exchange
+
 	kLineEventCallbacks []func(candle KLineEvent)
 }
 
-func NewKLineStream() *KLineStream {
+func NewKLineStream(exchange types.Exchange) *KLineStream {
```
c9s, line 19: you don't have to use types.Exchange here, you can use *Exchange since it's all for okx

pkg/exchange/okex/kline_stream.go
```diff
@@ -84,11 +90,23 @@ func (s *KLineStream) dispatchEvent(e interface{}) {
...
+		_ = syncSubscriptions(s.StandardStream.Conn, s.StandardStream.Subscriptions, WsEventTypeUnsubscribe, instType)
 	}
 	s.Resubscribe(func(old []types.Subscription) (new []types.Subscription, err error) {
 		// clear the subscriptions
 		return []types.Subscription{}, nil
 	})
 }
+
+func (s *KLineStream) getInstrumentType() okexapi.InstrumentType {
+	e := s.exchange.(*Exchange)
```
c9s, line 104: call s.exchange.getInstrumentType()

pkg/exchange/okex/kline_stream.go
anywhy: done

pkg/exchange/okex/kline_stream.go
anywhy: ``getInstrumentType()`` is a private function, so if we do it this way, we’ll have to expose it.

c9s: it's in the same package

pkg/exchange/okex/kline_stream.go
```diff
@@ -84,11 +90,22 @@ func (s *KLineStream) dispatchEvent(e interface{}) {
...
+		_ = syncSubscriptions(s.StandardStream.Conn, s.StandardStream.Subscriptions, WsEventTypeUnsubscribe, instType)
 	}
 	s.Resubscribe(func(old []types.Subscription) (new []types.Subscription, err error) {
 		// clear the subscriptions
 		return []types.Subscription{}, nil
 	})
 }
+
+func (s *KLineStream) getInstrumentType() okexapi.InstrumentType {
+	if s.exchange.IsMargin {
```
c9s, line 104: you can call s.exchange.getInstrumentType, since it's in the same package, calling its private method is doable.

pkg/exchange/okex/stream.go
```diff
@@ -338,3 +331,14 @@ func (s *Stream) dispatchEvent(e interface{}) {
 
 	}
 }
+
+func (s *Stream) getInstrumentType() okexapi.InstrumentType {
+	e := s.exchange.(*Exchange)
```
c9s, line 336: you can use the original type *Exchange of the field exchange here so you don't need to cast the type

pkg/exchange/okex/kline_stream.go
anywhy: done

pkg/exchange/okex/kline_stream.go
anywhy: done

c9s (approved):

PR #2025, opened 2025-05-07
FEATURE: [okx]  support global order quantity convert to okx contract size

anywhy:


pkg/exchange/okex/exchange.go
```diff
@@ -170,6 +170,8 @@ func (e *Exchange) QueryMarkets(ctx context.Context) (types.MarketMap, error) {
 			// OKEx does not offer minimal notional, use 1 USD here.
 			MinNotional: fixedpoint.One,
 			MinAmount:   fixedpoint.One,
+
+			ContractValue: fixedpoint.MustNewFromString(instrument.ContractValue),
```
c9s, line 174: you can update the type instrument.ContractValue , so we can parse it directly from JSON

c9s (approved):

c9s:
i left one comment

c9s (approved):

pkg/exchange/okex/exchange.go
```diff
@@ -350,14 +355,7 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (*t
...
 		// Because our order.Quantity unit is base coin, so we indicate the target currency to Base.
-		switch order.Side {
-		case types.SideTypeSell:
-			orderReq.Size(order.Market.FormatQuantity(order.Quantity))
-			orderReq.TargetCurrency(okexapi.TargetCurrencyBase)
-		case types.SideTypeBuy:
-			orderReq.Size(order.Market.FormatQuantity(order.Quantity))
-			orderReq.TargetCurrency(okexapi.TargetCurrencyBase)
-		}
+		orderReq.TargetCurrency(okexapi.TargetCurrencyBase)
```
c9s, line 358: @anywhy did you test this on both spot and futures?

pkg/exchange/okex/exchange.go
```diff
@@ -350,14 +355,7 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (*t
...
 		// target currency = Default is quote_ccy for buy, base_ccy for sell
 		// Because our order.Quantity unit is base coin, so we indicate the target currency to Base.
-		switch order.Side {
-		case types.SideTypeSell:
-			orderReq.Size(order.Market.FormatQuantity(order.Quantity))
-			orderReq.TargetCurrency(okexapi.TargetCurrencyBase)
-		case types.SideTypeBuy:
-			orderReq.Size(order.Market.FormatQuantity(order.Quantity))
-			orderReq.TargetCurrency(okexapi.TargetCurrencyBase)
-		}
```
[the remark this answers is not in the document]
bailantaotao: Set `orderReq.Size(order.Market.FormatQuantity(order.Quantity))`.

pkg/types/market.go
```diff
@@ -265,6 +267,17 @@ func (m Market) AdjustQuantityByMaxAmount(quantity, currentPrice, maxAmount fixe
 	return m.TruncateQuantity(quantity)
 }
 
+// AdjustQuantityToContractSize adjusts the quantity to contract size
+func (m Market) AdjustQuantityToContractSize(quantity fixedpoint.Value) fixedpoint.Value {
+	// Validate input parameters
+	if quantity.Sign() <= 0 || m.ContractValue.Sign() <= 0 || m.StepSize.Sign() <= 0 {
```
c9s, line 273: if m.ContractValue is not defined, use the original quantity?

c9s: same thing for m.StepSize

c9s (approved):

c9s:
This affects the behavior of syncing order history.

Under OKX’s margin mode (specifically spot & futures mode), it’s possible to place orders using the cash mode. However, if you try to query order history using the margin mode, you won’t be able to find those orders placed via cash mode, because they’re categorized under “Spot”.

When using OKX margin, it seems that operating in Spot mode is generally easier—mainly because market orders can specify size using the base currency instead of the quote.

That said, when syncing, we’ll need a workaround to handle this edge case: the query for order history should be forced to use “spot” instead of “margin”.

PR #2031, opened 2025-05-14
FEATURE: [coinbase] integrate TickKLineDriver with kline stream subscription

dboyliao:


pkg/exchange/coinbase/stream_handlers.go
```diff
@@ -242,13 +243,22 @@ func (s *Stream) buildMapForMarketStream(channel2LocalSymbolsMap map[types.Chann
 
 		// binding serial market store
 		for symbol, options := range klineOptionsMap {
-			store := types.NewSerialMarketDataStore(symbol, minInterval, true)
+			klineDriver := klinedriver.NewTickKLineDriver(
+				symbol, time.Second*5,
```
zenixls2, line 247: minInterval is defined as 1sec. Does 5 seconds here having the same meaning?

zenixls2:
Is SerialMarketDataStore used elsewhere?
If not, maybe time to remove the implementation? 
cc @c9s

pkg/exchange/coinbase/stream_handlers.go
dboyliao: Oh, `minInterval` was 1 minute not 1 sec.
`minInterval` is initialized as 1 sec and updated to the minimal supported interval on Coinbase in the `init()`.

5 seconds does have similar meaning here, which is the klines emitting frequency for the driver.

dboyliao:
As far as I know, SerialMarketDataStore is not used elsewhere.
But I'm not sure if somebody is using it in their own project or not.
That's why I still keep it whereas the kline driver does pretty much the same thing.

pkg/exchange/coinbase/stream_handlers.go
dboyliao: You just remind me `minInterval` is no longer required.
I'll remove it as well.

dboyliao:
As far as I know, `SerialMarketDataStore` is still used in several strategies, such as `drift` and `elliottwave`.
Though it should be able to be replaced by the tick kline driver, let's leave it as is for now.

zenixls2 (approved):

c9s (approved):

PR #2040, opened 2025-05-20
FEATURE: [grid2][dca2] cancel all orders again to avoid the submit or…

kbearXD:
…ders still working


ycdesu (approved):
We assume we will place all orders within 10 seconds. However, the context timeouts are longer than 10 seconds. It would be better to use channel to ensure the executions are in order.

PR #2043, opened 2025-05-21
feature: add Ordinary Least Square for series

zenixls2:
Add OLS. Need test cases to compare with the result from python.

### DIFF
--- a/pkg/types/indicator.go
+++ b/pkg/types/indicator.go
@@ -234,7 +234,7 @@ func switchIface(b interface{}) Series {
 // Divid two series, result[i] = a[i] / b[i]
 func Div(a interface{}, b interface{}) SeriesExtend {
 	aa := switchIface(a)
-	if 0 == b {
+	if b == 0 {
 		panic("Divid by zero exception")
 	}
 	bb := switchIface(b)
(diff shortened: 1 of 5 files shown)


pkg/types/trade_stats.go
```diff
@@ -19,7 +19,7 @@ type IntervalProfitCollector struct {
 	Interval  Interval      `json:"interval"`
 	Profits   *floats.Slice `json:"profits"`
 	Timestamp *floats.Slice `json:"timestamp"`
-	tmpTime   time.Time     `json:"tmpTime"`
```
zenixls2, line 22: by linter

pkg/types/indicator.go
```diff
@@ -408,6 +408,41 @@ func Array(a Series, limit ...int) (result []float64) {
...
+const (
+	RegressionTypeConstant              RegressionType = "c"
+	RegressionTypeConstantTrend         RegressionType = "ct"
+	RegressionTypeConstantTrendSeasonal RegressionType = "ctt"
+	RegressionTypeNone                  RegressionType = "n"
+)
+
+// Ordinary Least Squares fit result, only support 1d array
+func OSL(a SeriesExtend, b SeriesExtend, n int) (float64, float64) {
+	if a.Length() < n {
```
dboyliao, line 422: How about setting `n` to `a.Length()` if `a.Length() < n`?
So that you can continue the calculation for a result still.

pkg/types/indicator.go
```diff
@@ -408,6 +408,41 @@ func Array(a Series, limit ...int) (result []float64) {
...
+		return 0, 0
+	}
+	if b.Length() < n {
```
dboyliao, line 425: ditto.

pkg/types/indicator.go
```diff
@@ -408,6 +408,41 @@ func Array(a Series, limit ...int) (result []float64) {
...
+		return 0, 0
+	}
+	numerator := 0.0
```
dboyliao, line 428: Before starting the calculation, maybe it's better to add a check on the equality of the length of `a` and `b`?

c9s (approved):

pkg/types/indicator.go
zenixls2: Not necessary. We only process last n elements of the time series.

pkg/types/indicator.go
zenixls2: done.

pkg/types/indicator.go
zenixls2: you're correct. The other implementations are following the same rule.
done.

dboyliao (approved):

PR #2047, opened 2025-05-23
REFACTOR: [types] Add UUID to OrderQuery

dboyliao:


pkg/exchange/coinbase/exchange.go
```diff
@@ -379,10 +379,16 @@ func (e *Exchange) QueryKLines(ctx context.Context, symbol string, interval type
...
+	if len(q.UUID) == 0 {
+		req = req.OrderID(q.OrderID)
+	} else {
+		req = req.OrderID(q.UUID)
+	}
+
 	cbOrder, err := req.Do(ctx)
 	if err != nil {
-		return nil, errors.Wrapf(err, "failed to get order: %v", q.OrderID)
+		return nil, errors.Wrapf(err, "failed to get order: %v", q.UUID)
```
kbearXD, line 391: we may query by q.OrderID, should we add q.OrderID into the error message ?

dboyliao: The order id is a hash value of `q.UUID`.
I was thinking knowing `q.UUID` is sufficient to know `q.OrderID`.

But I think you get a point.
If I choose to support `q.OrderID` as query condition as well, I should include it into the message as well.
I will include the entire `q` into the message with `%+v`.

pkg/types/order.go
```diff
@@ -260,6 +260,7 @@ type OrderQuery struct {
 	Symbol        string
 	OrderID       string
 	ClientOrderID string
+	UUID          string
```
c9s, line 263: call it OrderUUID? Align with the OrderID field

dboyliao: done

kbearXD (approved):

PR #2048, opened 2025-05-23
FEATURE: [xmaker][exchange] support mock exchange for hedging 

c9s:


pkg/exchange/mock/exchange.go
```diff
@@ -0,0 +1,379 @@
...
+	tradeId uint64
+
+	openOrders map[uint64]types.Order
+
+	markets types.MarketMap
+
+	mu sync.Mutex
+
+	// trading symbol
+	symbol string
```
dboyliao, line 48: `symbol` is not used?

c9s: it's used in the next PR~

PR #2058, opened 2025-06-02
FEATURE: [indicatorv2] Premium Stream

dboyliao:


pkg/indicator/v2/premium_signal.go
```diff
@@ -0,0 +1,90 @@
...
+	symbol1, symbol2 string
+	interval         types.Interval
+	price1, price2   float64
+	premiumMargin    float64
+}
+
+func (s *PremiumSignalStream) BackFill(klines []types.KLine) {
+	if s.kLineStream == nil {
+		update1 := types.KLineWith(s.symbol1, s.interval, func(kline types.KLine) {
+			s.price1 = kline.Close.Float64()
```
c9s: You can use PriceStream if you only need one price from the KLine

pkg/indicator/v2/premium_signal.go
```diff
@@ -0,0 +1,90 @@
...
+	} else {
+		premiumType = Neutral
+	}
+	// push and emit
+	s.PushAndEmit(float64(premiumType))
+}
+
+func PremiumSignal(
+	stream *KLineStream,
+	symbol1, symbol2 string,
```
c9s: you don't need symbol since KLineStream already handles it

pkg/indicator/v2/premium_signal.go
```diff
@@ -0,0 +1,90 @@
...
+	interval types.Interval,
```
c9s: you don't need this too

pkg/indicator/v2/premium_signal.go
```diff
@@ -0,0 +1,90 @@
...
+		Float64Series: types.NewFloat64Series(),
+		premiumMargin: premiumMargin,
+		interval:      interval,
+		symbol1:       symbol1,
+		symbol2:       symbol2,
+		kLineStream:   stream,
+	}
+	// stream can be nil (ex: backtesting)
+	if s.kLineStream != nil {
+		s.kLineStream.AddSubscriber(
```
c9s: you can check EWMA for the usage of AddSubscriber

pkg/indicator/v2/premium_signal.go
dboyliao: done.

pkg/indicator/v2/premium_signal.go
```diff
@@ -0,0 +1,48 @@
+package indicatorv2
+
+import (
+	"math"
+
+	"github.com/c9s/bbgo/pkg/types"
+)
+
+type PremiumSignalStream struct {
```
c9s: call it PremiumStream?

PR #2066, opened 2025-06-11
feature: add np.random functions to generate rand arrays

zenixls2:
Add utility functions to generate floats with random numbers.


zenixls2:
also add percentile function.

c9s (approved):

PR #2080, opened 2025-07-01
FEATURE: [core] Add slack upload api support

c9s:


pkg/notifier/slacknotifier/slack_test.go
```diff
@@ -20,3 +23,25 @@ func Test_objectName(t *testing.T) {
 		assert.Equal(t, "A", objectName(obj))
 	})
 }
+
+func TestSlack(t *testing.T) {
+	apiToken := os.Getenv("SLACK_API_TOKEN")
+	channel := os.Getenv("SLACK_CHANNEL")
```
ycdesu, line 29: Should we ignore this test if those credentials are not configured?

pkg/notifier/telegramnotifier/telegram.go
```diff
@@ -201,8 +211,15 @@ func (n *Notifier) NotifyTo(channel string, obj interface{}, args ...interface{}
...
 }
 
-func (n *Notifier) SendPhoto(buffer *bytes.Buffer) {
-	n.SendPhotoTo("", buffer)
+func (n *Notifier) Upload(file *types.UploadFile) {
+	select {
+	case n.taskC <- notifyTask{
+		photoBuffer: file.Data,
+	}:
+	case <-time.After(1 * time.Second):
```
ycdesu, line 219: If `n.taskC` is full for 1 second, the `notifyTask` will be dropped without any logging message. It may be very difficult to debug this case if n.taskC is full

pkg/notifier/slacknotifier/slack_test.go
c9s: you are right, I need to add this check

c9s: fixed

pkg/notifier/telegramnotifier/telegram.go
c9s: yeah I need to find an another time to revisit this code and improve them

ycdesu (approved):

gx578007 (approved):

PR #2081, opened 2025-07-02
REFACTOR: [tradingdesk] refactor with TradingManager

narumiruna:


pkg/strategy/tradingdesk/strategy.go
```diff
@@ -76,68 +66,10 @@ func (s *Strategy) Run(ctx context.Context, _ bbgo.OrderExecutor, session *bbgo.
...
-
-	s.OrderExecutorMap[symbol] = executor
-	return executor, nil
-}
-
 // OpenPosition opens a new position with risk-based position sizing.
 // The position size is calculated based on MaxLossLimit, stop loss price, and available balance.
 func (s *Strategy) OpenPosition(ctx context.Context, param OpenPositionParam) error {
-	executor, err := s.getOrCreateOrderExecutor(param.Symbol)
+	executor, err := s.TradingManagerMap.GetOrderExecutor(ctx, s.Environment, s.Session, param.Symbol, s.ID(), s.InstanceID())
```
c9s: should be "getTradingManager" and then call GetOrderExecutor on it?

c9s (approved):

pkg/strategy/tradingdesk/manager.go
```diff
@@ -0,0 +1,207 @@
...
+	if side == types.SideTypeBuy {
+		// For long positions, risk is current price - stop loss price
+		return currentPrice.Sub(stopLossPrice)
+	}
+	// For short positions, risk is stop loss price - current price
+	return stopLossPrice.Sub(currentPrice)
+}
+
+// createInvalidStopLossError creates an appropriate error message for invalid stop loss prices
+func (m *TradingManager) createInvalidStopLossError(side types.SideType, currentPrice fixedpoint.Value) error {
```
c9s: this method doesn't need TradingManager, it could be a simple function

pkg/strategy/tradingdesk/manager.go
```diff
@@ -0,0 +1,207 @@
...
+
+// getAvailableBalance returns the available balance for the appropriate currency based on order side
+func (m *TradingManager) getAvailableBalance(side types.SideType) (fixedpoint.Value, error) {
+	account := m.Session.GetAccount()
+
+	var currency string
+	if side == types.SideTypeBuy {
+		currency = m.Market.QuoteCurrency // For buy orders, we need quote currency balance
+	} else {
+		currency = m.Market.BaseCurrency // For sell orders, we need base currency balance
```
c9s: we don't use BaseCurrency for the futures account,

we only deposit USDT in the futures account, and there will be no base assets

c9s: you might need 2 modes:
1. futures mode (use quote balance and the current price to calcualte how much base we can submit)
2. spot mode (check base currency)

pkg/strategy/tradingdesk/open_position_param.go
```diff
@@ -12,7 +12,7 @@ type OpenPositionParam struct {
 	Confidence      fixedpoint.Value `json:"confidence"`
 	Side            types.SideType   `json:"side"`
 	Quantity        fixedpoint.Value `json:"quantity"`
-	StopLossPrice   fixedpoint.Value `json:"stopLossPrice"`
+	StopPrice       fixedpoint.Value `json:"stopPrice"`
```
c9s: why not use StopLossPrice?

pkg/strategy/tradingdesk/manager.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		return err
+	}
+
+	order := types.SubmitOrder{
+		Symbol:    param.Symbol,
+		Side:      param.Side,
+		Type:      types.OrderTypeMarket,
+		Quantity:  quantity,
+		StopPrice: param.StopPrice,
```
c9s: when using StopPrice, it means you are placing a stop loss order, not opening a new position.

you need to open the position first, and then submit a stop loss order with the stop price

c9s: Eventually you will need 3 orders to submit:

1. market order to open a position
2. stop loss order for setting up the stop loss price (with ClosePosition option)
3. take profit order for setting up the take price (with ClosePosition option)

PR #2084, opened 2025-07-07
FEATURE: [Coinbase] Implement ExchangeTradeHistoryService and Fix Pagination

dboyliao:


pkg/exchange/coinbase/exchange.go
```diff
@@ -485,3 +475,93 @@ func (e *Exchange) DefaultFeeRates() types.ExchangeFee {
...
 }
+
+// ExchangeTradeHistoryService
+// QueryClosedOrders queries closed orders for the given symbol and time range.
+// startTime and endTime are inclusive.
+func (e *Exchange) QueryClosedOrders(ctx context.Context, symbol string, startTime, endTime time.Time, lastOrderID uint64) ([]types.Order, error) {
+	if lastOrderID > 0 {
+		log.Warning("lastOrderID is not supported for Coinbase, it will be ignored")
+	}
+	cbOrders, err := e.queryOrdersByPagination(ctx, symbol, []string{"done"})
```
c9s: startTime and endTime are not applied here?

pkg/exchange/coinbase/exchange.go
```diff
@@ -485,3 +475,93 @@ func (e *Exchange) DefaultFeeRates() types.ExchangeFee {
...
+			options.StartTime.Format(time.RFC3339Nano),
+		)
+	}
+	if options.EndTime != nil {
+		req.EndDate(
+			options.EndTime.Format(time.RFC3339Nano),
+		)
+	}
+	if options.LastTradeID > 0 {
+		req.Before(options.LastTradeID)
```
c9s, line 550: the LastTradeID is used in ascending ordering, so it needs to be used as "After"

dboyliao: I've tested it and checked the response headers. It is `Before` in this case on Coinbase.

As you can see in the page below:
https://docs.cdp.coinbase.com/exchange/rest-api/pagination
The pagination id for `before` is for the page "newer" than the current page.
To my understanding, the `LastTradeID` is indicating the last trade id in the DB , we should return the trades "newer" to the last trade.
Hence, it should be `before` not `after`.

pkg/exchange/coinbase/exchange.go
```diff
@@ -485,3 +498,90 @@ func (e *Exchange) DefaultFeeRates() types.ExchangeFee {
...
+	if err != nil {
+		return nil, err
+	}
+
+	limit := int(options.Limit)
+	// pagination done if
+	// 1. limit > 0 and we have enough trades
+	// 2. the trades are less than PaginationLimit -> have reached the end of the pagination
+	donePagination := (limit > 0 && len(cbTrades) >= limit) || (len(cbTrades) < PaginationLimit)
+	for {
```
c9s: you can write it as `for !donePagination {` ?

pkg/exchange/coinbase/exchange.go
dboyliao: Done

pkg/exchange/coinbase/exchange.go
dboyliao: Opps, I forgot that syntax for a `for` loop.
Will fix it.

pkg/exchange/coinbase/exchange.go
```diff
@@ -485,3 +492,87 @@ func (e *Exchange) DefaultFeeRates() types.ExchangeFee {
...
+
+func (e *Exchange) queryProductTradesByPagination(
+	ctx context.Context, symbol string, options *types.TradeQueryOptions,
+) (cbTrades api.TradeSnapshot, err error) {
+	defer func() {
+		if err == nil && options.Limit > 0 {
+			cbTrades = cbTrades[:options.Limit]
+		}
+	}()
+	req := e.client.NewGetOrderTradesRequest().Limit(PaginationLimit)
```
gx578007, line 537: Is it possible that PaginationLimit is equal to 0? If so, how will it impact the termination condition of pagination?

gx578007 (approved):

pkg/exchange/coinbase/exchange.go
dboyliao: `PaginationLimit` is defined as a `const` and it is assumed to be a positive number.
We can adjust the pagination size easily as needed.

If 0 is a major concern, I can add a check on it in `init()`.
However, I don't see the need here for now.

PR #2086, opened 2025-07-09
FEATURE: [coinbase] implement transfer history query service

dboyliao:


pkg/exchange/coinbase/api/v1/get_transfers_request.go
```diff
@@ -0,0 +1,92 @@
...
+const (
+	TransferTypeDeposit          TransferType = "deposit"
+	TransferTypeWithdraw         TransferType = "withdraw"
+	TransferTypeInternalDeposit  TransferType = "internal_deposit"
+	TransferTypeInternalWithdraw TransferType = "internal_withdraw"
+)
+
+// Coinbase use a custom time format for transfers
+// We handle it with a custom type TransferTime
+type TransferTime time.Time
```
c9s, line 22: Maybe move to types/time.go and name it like RFC3339NanoTimezone ?

dboyliao: Not likely.
There is a missing `T` in the timestamp returned by the API.
The example of the time string by the Coinbase `/transfers` API: `"2025-02-01 13:14:32.602546+00"`

kbearXD (approved):

PR #2088, opened 2025-07-10
🔧 fix(coinbase): allow empty asset parameter for transfer history queries and update tests

dboyliao:


pkg/exchange/coinbase/convert.go
```diff
@@ -236,7 +236,7 @@ func toGlobalWithdraw(transfer *api.Transfer) types.Withdraw {
 		// completed_at is not zero -> completed
 		withdraw.Status = types.WithdrawStatusCompleted
 	default:
-		withdraw.Status = types.WithdrawStatusUnknown
+		withdraw.Status = types.WithdrawStatusProcessing
```
c9s, line 239: Will this cause problem?

dboyliao: No, as far as I know.
1. `bbgo sync` filter on the status for `rejected` withdraws. The update made in this PR does not break the functionality of `bbgo sync`
2. In bbgo, only `xalign` and `deposit2transfer` depend on the deposit/withdraw status and they use it to find completed or pending transfers. So, the update made in this PR should no affect these strategies to work properly.

ycdesu (approved):

PR #2092, opened 2025-07-14
FIX: [coinbase] fix trades history pagination and update tests

dboyliao:


pkg/exchange/coinbase/exchange.go
```diff
@@ -539,41 +539,56 @@ func (e *Exchange) queryProductTradesByPagination(
...
-	cbTrades, err = req.Do(ctx)
+	cbTradesDirty, err := req.Do(ctx)
 	if err != nil {
 		return nil, err
 	}
-
+	seenTrades := make(map[uint64]struct{})
+	for _, trade := range cbTradesDirty {
+		seenTrades[trade.TradeID] = struct{}{}
+	}
```
c9s: not sure why do we need this?

dboyliao: Ya, you're right.
It's my experiment codes which I forgot to delete it.
Will fix it.

ycdesu (approved):

PR #2093, opened 2025-07-15
FIX: [coinbase] add executed quantity and update time to global order conversion

dboyliao:


pkg/exchange/coinbase/convert.go
```diff
@@ -41,24 +41,32 @@ func toGlobalOrderStatus(order *api.Order) types.OrderStatus {
...
+		OriginalStatus:   string(cbOrder.Status),
+		CreationTime:     cbOrder.CreatedAt,
+		IsWorking:        isWorkingOrder(cbOrder.Status),
+		ExecutedQuantity: cbOrder.FilledSize,
 	}
+	if cbOrder.Status == api.OrderStatusDone {
+		order.UpdateTime = cbOrder.DoneAt
+	} else {
+		order.UpdateTime = types.Time(time.Now())
+	}
```
gx578007: Be careful that this could lead to an order is done before it has been partially filled.

dboyliao: fixed

ycdesu (approved):

PR #2095, opened 2025-07-15
FEATURE: [tradingdesk] store expiryTime when ttl is defined and create close trigger

c9s:


dboyliao (approved):

pkg/strategy/tradingdesk/manager.go
```diff
@@ -164,6 +168,14 @@ func (m *TradingManager) OpenPosition(ctx context.Context, params OpenPositionPa
 
 	m.logger.Infof("created orders: %+v", createdOrders)
 
+	if params.TimeToLive > 0 {
+		expiryTime := time.Now().Add(params.TimeToLive)
+		m.ExpiryTime = &expiryTime
+
+		closeTrigger := m.createCloseTrigger(ctx, params.TimeToLive)
+		go closeTrigger()
```
gx578007: If any error after this happens and causes early return, will it try to close the position even there's no open position?

c9s: yes, it always tries to close
the position

gx578007 (approved):

PR #2102, opened 2025-07-17
FEATURE: [tradingdesk] add slack blocks and fix ClosePosition conversion

c9s:


pkg/strategy/tradingdesk/manager.go
```diff
@@ -517,3 +516,119 @@ func fallbackStopLossTakeProfit(
...
+
+	return fixedpoint.Zero
+}
+
+// SlackBlocks generates Slack message blocks with position details, take profit, stop loss prices, and unrealized profit.
+func (m *TradingManager) SlackBlocks() []slack.Block {
+	var blocks []slack.Block
+
+	// Query the latest price
+	ticker, err := m.session.Exchange.QueryTicker(context.Background(), m.Position.Symbol)
```
dboyliao, line 536: Isn't it better to use a timeout context here?

dboyliao (approved):

pkg/strategy/tradingdesk/manager.go
c9s: it should be ok, we've set the timeout for http.Client

ycdesu (approved):

PR #2107, opened 2025-07-21
FEATURE: [migrations] Add order uuid column in `orders` and `trades` tables

dboyliao:


migrations/mysql/20250721140324_orders_add_uuid.sql
```diff
@@ -0,0 +1,9 @@
+-- +up
+-- +begin
+ALTER TABLE `orders` ADD COLUMN `uuid` VARCHAR(255) NOT NULL DEFAULT '';
```
c9s: why using 255 chars here?

and it should be an optional field?

dboyliao: If `uuid` is not given, it will be default to `''`.
I think it's effectively the same as being optional.

And I just want to set a number that's long enough to accommodate a unique id string.

dboyliao: fixed

pkg/types/trade.go
```diff
@@ -193,6 +194,7 @@ func (trade Trade) String() string {
 		trade.ID,
 		trade.Time.Time().Format(time.StampMilli),
 		trade.Liquidity(),
+		trade.OrderUUID,
```
c9s: UUID is optional, I think you need to add a check or the text will be too long

c9s (approved):

pkg/types/trade.go
dboyliao: ok

dboyliao: done.

kbearXD (approved):

c9s (approved):

PR #2108, opened 2025-07-22
FEATURE: [coinbase] Add order execution metrics for submission and cancellation tracking

dboyliao:


ycdesu (approved):

pkg/exchange/coinbase/exchange.go
```diff
@@ -201,21 +201,27 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
 		req.ClientOrderID(order.ClientOrderID)
 	}
 
-	timeNow := time.Now()
+	// Record metrics timing
+	requestTime := time.Now()
 	res, err := req.Do(ctx)
+	duration := time.Since(requestTime)
+	recordOrderSubmissionMetrics(order, duration, err)
```
c9s: why do you need to pass error

dboyliao: I was thinking it would be easier to see the counts of different error.

dboyliao: Also, I need the error to get the response status code.
So that I can update corresponding metrics.

pkg/exchange/coinbase/metrics.go
```diff
@@ -0,0 +1,145 @@
...
+		"success": success,
+	}).Inc()
+}
+
+// Helper function to record order cancellation metrics
+func recordOrderCancelMetrics(order types.Order, duration time.Duration, err error) {
+	symbol := string(order.Symbol)
+
+	// Record cancellation result in total metrics
+	success := "true"
```
c9s: not used?

pkg/exchange/coinbase/metrics.go
```diff
@@ -0,0 +1,145 @@
...
+	symbol := string(order.Symbol)
+
+	// Record submission status
+	success := "true"
+	if err != nil {
+		success = "false"
+		// Categorize error code for error count metrics
+		var responseErr *requestgen.ErrResponse
+		statusCode := "200"
+		if errors.As(err, &responseErr) {
```
c9s: error casting should be done in. the caller frame

pkg/exchange/coinbase/metrics.go
dboyliao: No, it's used at L143

dboyliao: It's been refactored out.

pkg/exchange/coinbase/metrics.go
dboyliao: done.

PR #2123, opened 2025-07-31
FIX: [tradeService] update trade query to accommodate the binary order UUID

dboyliao:


pkg/service/trade.go
```diff
@@ -426,6 +427,25 @@ func queryTradesSQL(options QueryTradesOptions) string {
 	return sql
 }
 
+func (s *TradeService) getTradeSelectColumns() []string {
+	if s.DB.DriverName() != "mysql" {
+		return []string{"*"}
+	}
+	v := reflect.TypeOf(types.Trade{})
```
c9s: for type, we usually use "t" as the prefix of the variable.

tt (type of trade)

pkg/service/trade.go
```diff
@@ -426,6 +427,25 @@ func queryTradesSQL(options QueryTradesOptions) string {
...
+	var columns []string
+	for i := 0; i < v.NumField(); i++ {
+		field := v.Field(i)
+		if colname := field.Tag.Get("db"); colname != "" {
```
c9s: camalCase colName

pkg/service/trade.go
```diff
@@ -426,6 +427,25 @@ func queryTradesSQL(options QueryTradesOptions) string {
...
+			if colname == "order_uuid" {
```
c9s: maybe we can do: strings.HasSuffix(colName, "uuid") here

pkg/service/trade.go
```diff
@@ -426,6 +427,25 @@ func queryTradesSQL(options QueryTradesOptions) string {
...
+				columns = append(columns, "IF(order_uuid != '', BIN_TO_UUID(order_uuid, true), '') as order_uuid")
+			} else if colname != "-" {
```
c9s: let's skip "-" field directly before everything

pkg/service/trade.go
dboyliao: I prefer to be as specific as possible.
Or the future developers have to be aware of the side effect of naming in the column name with suffix `"uuid"`.

pkg/service/trade.go
dboyliao: done.

pkg/service/trade.go
dboyliao: done.

pkg/service/trade.go
dboyliao: done.

kbearXD (approved):

PR #2127, opened 2025-08-01
FEATURE: [xalign] live note support for critical balance discrepancy alert

dboyliao:


pkg/strategy/xalign/strategy.go
```diff
@@ -597,17 +603,22 @@ func (s *Strategy) align(ctx context.Context, sessions bbgo.ExchangeSessionMap)
...
+						BaseCurrency:      currency,
+						Delta:             q,
+						SustainedDuration: s.Duration.Duration(),
+						Price:             price,
+						Quantity:          quantity,
+						Amount:            amount,
+					},
+					livenote.Channel(s.LargeAmountAlert.Slack.Channel),
+					livenote.OneTimeMention(s.SlackNotifyMentions...),
+					livenote.Comment(s.Duration.Duration().String()),
```
c9s: add details in the comment?

c9s: describe how long does the alert sustained

dboyliao: I updated the comment.
Please have a look.

pkg/strategy/xalign/alert_cbd.go
```diff
@@ -93,3 +100,40 @@ func (m *CriticalBalanceDiscrepancyAlert) SlackAttachment() slack.Attachment {
 		},
 	}
 }
+
+func (m *CriticalBalanceDiscrepancyAlert) ObjectID() string {
+	if m.notifyDate.IsZero() {
```
c9s: since you don't store the alert object, so your notifyTime will always be zero

dboyliao: Oh, yes.
I fix it with a cache object.

PR #2154, opened 2025-08-16
refactor: use the built-in max/min to simplify the code

yinwenyu6:
use the built-in max/min to simplify the code

Inspired by https://github.com/c9s/bbgo/pull/2016  and replace all.


c9s (commented):
These are generated files, you need to update c9s/rockhopper

PR #2169, opened 2025-08-19
FIX: [coinbase] subscribe balance update for user data stream

dboyliao:


pkg/exchange/coinbase/exchange.go
```diff
@@ -120,6 +120,19 @@ func (e *Exchange) QueryAccountBalances(ctx context.Context) (types.BalanceMap,
 	return balances, nil
 }
 
+func (e *Exchange) QueryAccountID(ctx context.Context) (map[string]string, error) {
```
c9s: don't export this method since it's too specific

pkg/exchange/coinbase/stream_handlers.go
```diff
@@ -153,6 +153,35 @@ func (s *Stream) buildMapForUserDataStream(channel2LocalSymbolsMap map[types.Cha
 	// - order life cycle events: receive, open, done, change, activate(for stop orders)
 	// - order match
 	channel2LocalSymbolsMap[userChannel] = privateChannelLocalSymbols
+
+	// query account id for the symbols
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*20)
+	defer cancel()
+	markets, err := s.exchange.QueryMarkets(ctx)
+	if err != nil {
+		panic("fail to query markets")
```
c9s: could be dangerous here, any kinds of http error would cause panic

c9s (changes requested):

pkg/exchange/coinbase/stream_handlers.go
dboyliao: fixed

pkg/exchange/coinbase/exchange.go

pkg/exchange/coinbase/exchange.go
```diff
@@ -120,6 +120,19 @@ func (e *Exchange) QueryAccountBalances(ctx context.Context) (types.BalanceMap,
 	return balances, nil
 }
 
+func (e *Exchange) queryAccountIDs(ctx context.Context) (map[string]string, error) {
```
c9s: i feel like you need a queryAccounts method, and another "toAccountCurrencyIds" function to convert the accounts into the map type

pkg/exchange/coinbase/stream_handlers.go
```diff
@@ -153,6 +153,38 @@ func (s *Stream) buildMapForUserDataStream(channel2LocalSymbolsMap map[types.Cha
 	// - order life cycle events: receive, open, done, change, activate(for stop orders)
 	// - order match
 	channel2LocalSymbolsMap[userChannel] = privateChannelLocalSymbols
+
+	// query account id for the symbols
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*20)
+	defer cancel()
+	markets, err := s.exchange.QueryMarkets(ctx)
+	if err != nil {
+		logStream.WithError(err).Warn("fail to query markets")
```
c9s: you can add a FieldLogger i. the stream struct for reuse

dboyliao: done.

pkg/exchange/coinbase/exchange.go
dboyliao: Yes, I add a similar function, `queryAccountIDsBySymbols`, instead.
Internally `queryAccountIDsBySymbols` will call `queryAccountIDs` and convert the map accordingly.

I can't use `QueryAccount` directly is that it would lost the account id in the `types.Account`.
So I have to use the raw get balance request here.

pkg/exchange/coinbase/stream.go
```diff
@@ -37,6 +33,11 @@ type Stream struct {
 	passphrase string
 	secretKey  string
 
+	logger *logrus.Entry
```
c9s: use logrus.FieldLogger interface

c9s (approved):

pkg/exchange/coinbase/stream.go
dboyliao: done.

kbearXD (approved):

PR #2176, opened 2025-08-23
FIX: [xmaker] check and handle eof error

c9s:


pkg/exchange/max/exchange.go
```diff
@@ -691,19 +704,19 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
 			if responseErr.StatusCode >= 400 {
 				recoveredOrder, recoverErr := e.recoverOrder(ctx, o)
 				if recoverErr != nil {
-					return createdOrder, fmt.Errorf("unable to recover order, error: %w", recoverErr)
+					return nil, fmt.Errorf("unable to recover order, error: %w", recoverErr)
 				}
 
 				// note, recoveredOrder could be nil if the order is not found
-				return recoveredOrder, nil
+				return recoveredOrder, err
```
dboyliao: The `err` here is not nil and `recoveredOrder` can be either nil or not right?
I think this makes the return values by `.SubmitOrder` very confusing to the caller.

c9s: recoveredOrder is nil, because order not found

c9s: fixed, thanks

pkg/exchange/max/exchange.go
```diff
@@ -678,11 +679,26 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
...
+
+		if errors.Is(err, io.EOF) {
+			recoveredOrder, recoverErr := e.recoverOrder(ctx, o)
+			if recoverErr != nil {
+				return nil, fmt.Errorf("unable to recover order, error: %w", recoverErr)
+			} else if recoveredOrder == nil {
+				return nil, fmt.Errorf("unable to recover order, error: %w", err)
+			}
+
+			// note, recoveredOrder could be nil if the order is not found
```
dboyliao, line 695: remove this comment since it's no longer the case here.

PR #2179, opened 2025-08-23
FIX: [xalign] ensure notify date is truncated to the start of the day

dboyliao:


pkg/strategy/xalign/alert_cbd.go
```diff
@@ -122,7 +122,7 @@ func (m *CriticalBalanceDiscrepancyAlert) ObjectID() string {
 		cbdDateCache.notifyDate = now.Truncate(time.Hour * 24)
 	} else if time.Since(cbdDateCache.notifyDate) > time.Hour*24 {
 		// update the notify date cache if it's a new day
-		cbdDateCache.notifyDate = now
+		cbdDateCache.notifyDate = now.Truncate(time.Hour * 24)
 	}
 
 	dateString := cbdDateCache.notifyDate.Format(time.DateOnly)
```
c9s, line 128: you need to call truncate here?

PR #2182, opened 2025-08-23
IMPROVE: [order executor] improve batch place orders

anywhy:


pkg/bbgo/order_execution.go
```diff
@@ -325,26 +326,45 @@ type OrderCallback func(order types.Order)
...
 
-	var errIndexes []int
 	for i, submitOrder := range submitOrders {
-		createdOrder, err2 := exchange.SubmitOrder(ctx, submitOrder)
-		if err2 != nil {
-			err = multierr.Append(err, err2)
-			errIndexes = append(errIndexes, i)
-		}
+		go func(i int, submitOrder types.SubmitOrder) {
+			defer wg.Done()
```
c9s: use errgroup to collect errors from different goroutine

pkg/bbgo/order_execution.go
```diff
@@ -325,26 +326,45 @@ type OrderCallback func(order types.Order)
 func BatchPlaceOrder(
 	ctx context.Context, exchange types.Exchange, orderCallback OrderCallback, submitOrders ...types.SubmitOrder,
 ) (types.OrderSlice, []int, error) {
-	var createdOrders types.OrderSlice
-	var err error
+	results := make([]struct {
```
c9s: you can predefine a struct like

orderStatus or erroredOrders

something like that

pkg/bbgo/order_execution.go
anywhy: done

pkg/bbgo/order_execution.go
anywhy: done

pkg/bbgo/order_execution.go
```diff
@@ -321,30 +322,46 @@ func (c *BasicRiskController) ProcessOrders(
...
+	Err   error
+}
+
 // BatchPlaceOrder
 func BatchPlaceOrder(
 	ctx context.Context, exchange types.Exchange, orderCallback OrderCallback, submitOrders ...types.SubmitOrder,
 ) (types.OrderSlice, []int, error) {
-	var createdOrders types.OrderSlice
-	var err error
+	g, _ := errgroup.WithContext(ctx)
```
c9s: keep the cancel func, and defer the cancel call

pkg/bbgo/order_execution.go
```diff
@@ -321,30 +322,46 @@ func (c *BasicRiskController) ProcessOrders(
...
 	for i, submitOrder := range submitOrders {
-		createdOrder, err2 := exchange.SubmitOrder(ctx, submitOrder)
-		if err2 != nil {
-			err = multierr.Append(err, err2)
-			errIndexes = append(errIndexes, i)
-		}
+		i, submitOrder := i, submitOrder
+		g.Go(func() error {
+			order, err := exchange.SubmitOrder(ctx, submitOrder)
+			results[i] = batchOrderStatus{i, order, err}
```
c9s: use literal struct constructor syntax, don't fill fields by ordering.

pkg/bbgo/order_execution.go
```diff
@@ -321,30 +322,46 @@ func (c *BasicRiskController) ProcessOrders(
...
+
+	for _, result := range results {
+		if result.Err != nil {
+			err = multierr.Append(err, result.Err)
+			errIndexes = append(errIndexes, result.Index)
+		} else if result.Order != nil {
+			result.Order.Tag = submitOrders[result.Index].Tag
 			if orderCallback != nil {
-				orderCallback(*createdOrder)
+				orderCallback(*result.Order)
```
c9s: you need to call orderCallback immediately once you submitted the order, this is too far from the call.

pkg/bbgo/order_execution.go
```diff
@@ -321,30 +322,46 @@ func (c *BasicRiskController) ProcessOrders(
...
-	var errIndexes []int
+	results := make([]batchOrderStatus, len(submitOrders))
 	for i, submitOrder := range submitOrders {
-		createdOrder, err2 := exchange.SubmitOrder(ctx, submitOrder)
-		if err2 != nil {
-			err = multierr.Append(err, err2)
-			errIndexes = append(errIndexes, i)
-		}
+		i, submitOrder := i, submitOrder
+		g.Go(func() error {
```
c9s: I wonder if we can really benefits from this goroutine, cuz binance set the rate limiter 10 orders per second 

(Spot Order Rate Limit (for a single account) 100 orders per 10 seconds)

strategies like grid2 depends on this order execution, and the sequence of the orders are important.

would prefer a concurrent option for this, and apply the original behaivor by default.

pkg/bbgo/order_execution.go
```diff
@@ -321,30 +322,46 @@ func (c *BasicRiskController) ProcessOrders(
...
+			order, err := exchange.SubmitOrder(ctx, submitOrder)
+			results[i] = batchOrderStatus{i, order, err}
+			return nil
+		})
+	}
 
-		// if createdOrder is not nil and has a valid OrderID, we should add it to the createdOrders
-		if createdOrder != nil && createdOrder.OrderID > 0 {
-			createdOrder.Tag = submitOrder.Tag
+	_ = g.Wait()
```
c9s: please don't ignore the error, at least you should log it.

pkg/bbgo/order_execution.go
anywhy: On Binance Futures, you can submit 300 orders every 10 seconds. I’ve modified my strategy to use a concurrent mode, which has significantly improved submission efficiency. I added a switch to control whether concurrency is enabled, and it’s turned off by default.

pkg/bbgo/order_execution.go
anywhy: done

pkg/bbgo/order_execution.go
c9s: ok, thanks

c9s (approved):

pkg/exchange/okex/stream.go
```diff
@@ -202,7 +202,7 @@ func (s *Stream) subscribePrivateChannels(next func()) func() {
 		instType := s.exchange.getInstrumentType()
 		var subs = []WebsocketSubscription{
 			{Channel: ChannelAccount},
-			{Channel: "orders", InstrumentType: string(instType)},
+			{Channel: ChannelOrderTrades, InstrumentType: string(instType)},
```
c9s, line 205: not ChannelOrder here?

anywhy: I see that ``ChannelOrderTrades`` is defined as ``orders``.
https://github.com/c9s/bbgo/blob/981c703acf87c9de94b381745a6ae1a7e7d88844/pkg/exchange/okex/parse.go#L41

pkg/bbgo/order_execution.go
```diff
@@ -321,30 +327,59 @@ func (c *BasicRiskController) ProcessOrders(
...
-	var errIndexes []int
-	for i, submitOrder := range submitOrders {
-		createdOrder, err2 := exchange.SubmitOrder(ctx, submitOrder)
-		if err2 != nil {
-			err = multierr.Append(err, err2)
-			errIndexes = append(errIndexes, i)
+	submitOrder := func(i int, order types.SubmitOrder, execCtx context.Context) {
+		createdOrder, err := exchange.SubmitOrder(execCtx, order)
+		if orderCallback != nil && createdOrder != nil && createdOrder.OrderID > 0 {
+			createdOrder.Tag = order.Tag
```
gx578007, line 345: The behavior has been changed. The callback is required for Tag to be assigned. Is this indended?

c9s: I think it's because the .Tag is not persisted, and only exists in the types.SubmitOrder. so we need to copy the tag to the returned order (the created one)

anywhy: Yeah, Exactly. If don’t copy it, the Tag will be lost.

PR #2195, opened 2025-09-02
FEATURE: [fixedpoint] add expirable value wrapper

c9s:


pkg/exchange/max/exchange.go
```diff
@@ -43,17 +42,16 @@ type Exchange struct {
...
-	}
-
-	client := maxapi.NewRestClient(baseURL)
+	client := maxapi.NewRestClientDefault()
 	client.Auth(key, secret)
 	if subAccount != "" {
 		client.SetSubAccount(subAccount)
 	}
 
+	if err := client.Initialize(context.Background()); err != nil {
```
ycdesu, line 51: Could we specify a timeout context to control the initialization time, or allow the user to specify the context through middleware?

c9s: I wanted to, but the change could be huge, we should do it in a separate PR and get it well tested

dboyliao (approved):

pkg/exchange/max/exchange.go
```diff
@@ -43,17 +42,16 @@ type Exchange struct {
 }
 
 func New(key, secret, subAccount string) *Exchange {
-	baseURL := maxapi.ProductionAPIURL
-	if override := os.Getenv("MAX_API_BASE_URL"); len(override) > 0 {
-		baseURL = override
-	}
-
-	client := maxapi.NewRestClient(baseURL)
+	client := maxapi.NewRestClientDefault()
```
dboyliao, line 45: Does `max` no longer support `MAX_API_BASE_URL` env variable?

c9s: it's supported in the NewRestClientDefault constructor

ycdesu (approved):

PR #2202, opened 2025-09-09
FIX: [coinbase] fix `IsBuyer` for RESTful

dboyliao:


pkg/exchange/coinbase/convert.go
```diff
@@ -79,8 +80,8 @@ func toGlobalTrade(cbTrade *api.Trade) types.Trade {
 		Quantity:      cbTrade.Size,
 		QuoteQuantity: cbTrade.Size.Mul(cbTrade.Price),
 		Symbol:        toGlobalSymbol(cbTrade.ProductID),
-		Side:          toGlobalSide(cbTrade.Side),
-		IsBuyer:       cbTrade.Liquidity == api.LiquidityTaker,
+		Side:          side,
```
ycdesu, line 83: Don't we have to invoke `toGlobalSide` anymore?

kbearXD (approved):

pkg/exchange/coinbase/convert.go
dboyliao: Yes, we use `toGlobalSide` extensively.

PR #2203, opened 2025-09-10
IMPROVE: [xmaker] improve hedge market margin rebalance

c9s:


kbearXD (approved):

pkg/strategy/xmaker/repay.go
```diff
@@ -0,0 +1,61 @@
...
+	"golang.org/x/time/rate"
+
+	"github.com/c9s/bbgo/pkg/bbgo"
+	"github.com/c9s/bbgo/pkg/fixedpoint"
+	"github.com/c9s/bbgo/pkg/types"
+)
+
+var repayRateLimiter = rate.NewLimiter(rate.Every(30*time.Second), 1)
+
+func tryToRepayDebts(ctx context.Context, session *bbgo.ExchangeSession) bool {
```
dboyliao, line 16: Why do you decide to refactor `tryToRepayDebts` from a method to a standalone function?
The difference I see is replacing `s.logger` with module level `log`.
Why is that?

dboyliao (approved):

pkg/strategy/xmaker/repay.go
c9s: it's unrelated to a strategy, it's more like a function designed for session, so we don't have to depend on the strategy logger. however, I would like to keep this function simple, so we use global logger here

PR #2204, opened 2025-09-10
FEATURE: [log] replace the log rotation plugin with lumberjack

anywhy:
The ```rotatelogs.WithLinkName``` option in file-rotatelogs is not container-friendly.

### DIFF
--- a/pkg/cmd/root.go
+++ b/pkg/cmd/root.go
@@ -6,18 +6,17 @@ import (
 	"path"
 	"runtime/pprof"
 	"strings"
-	"time"
 	_ "time/tzdata"
 
 	"github.com/heroku/rollrus"
 	"github.com/joho/godotenv"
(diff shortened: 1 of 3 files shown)


pkg/cmd/root.go
```diff
@@ -248,14 +246,12 @@ func init() {
...
-			path.Join(logDir, "access_log.%Y%m%d"),
-			rotatelogs.WithLinkName("access_log"),
-			// rotatelogs.WithMaxAge(24 * time.Hour),
-			rotatelogs.WithRotationTime(time.Duration(24)*time.Hour),
-		)
-		if err != nil {
-			log.Panic(err)
+
+		writer := &lumberjack.Logger{
+			Filename:   "access.log",
```
c9s: logDir is not used? you need to setup the path?

anywhy: yes, need setup path. done

PR #2205, opened 2025-09-11
FIX: [binance] maintain and update risk balances from stream

c9s:


pkg/exchange/binance/stream.go
```diff
@@ -85,6 +93,10 @@ type Stream struct {
 
 	// depthBuffers is used for storing the depth info
 	depthBuffers map[string]*depth.Buffer
+
+	riskBalances          map[string]*RiskBalance
```
gx578007, line 97: using sync.Map?

c9s: want to avoid using sync.Map here since it requires type-casting

PR #2210, opened 2025-09-15
IMPROVE: [exchange] Improve query account add position risk info

anywhy:


pkg/exchange/okex/convert_futures.go
```diff
@@ -42,25 +44,28 @@ func toGlobalFuturesPositions(futuresPositions []okexapi.Position) types.Futures
...
-			AverageCost: futuresPosition.AvgPx,
-			Base:        futuresPosition.Pos,
-			Quote:       futuresPosition.Pos.Mul(futuresPosition.MarkPx),
-			PositionRisk: &types.PositionRisk{
-				Leverage: futuresPosition.Lever,
-			},
-			Symbol: symbol,
+		positionSide := toGlobalPositionSide(okexapi.PosSide(futuresPosition.PosSide))
+
+		posKey := fmt.Sprintf("%s:%s", symbol, positionSide)
```
c9s: why in this format?

pkg/types/account.go
```diff
@@ -288,6 +289,21 @@ func (a *Account) UpdateBalances(balances BalanceMap) {
...
+
+	if a.FuturesInfo == nil {
+		a.FuturesInfo = &FuturesAccount{}
+	}
+
+	if a.FuturesInfo.Positions == nil {
+		a.FuturesInfo.Positions = make(FuturesPositionMap)
+	}
+
+	maps.Copy(a.FuturesInfo.Positions, positions)
```
c9s, line 301: not merge? does this Copy override the existing positions ?

pkg/types/account.go
```diff
@@ -288,6 +289,21 @@ func (a *Account) UpdateBalances(balances BalanceMap) {
...
+func (a *Account) UpdateFuturesPositions(positions FuturesPositionMap) {
+	a.Lock()
+	defer a.Unlock()
+
+	if a.FuturesInfo == nil {
+		a.FuturesInfo = &FuturesAccount{}
+	}
+
+	if a.FuturesInfo.Positions == nil {
+		a.FuturesInfo.Positions = make(FuturesPositionMap)
```
c9s: you can initialize these fields in NewAccount?

pkg/exchange/okex/convert_futures.go
anywhy: Since the position mode involves both one-way and hedge (two-way) positions, in the case of hedge mode, there may be a position record for each direction.

pkg/types/account.go
anywhy: The merge using the maps.Copy function has been implemented.
``` go
func Copy[M1 ~map[K]V, M2 ~map[K]V, K comparable, V any](dst M1, src M2) {
	for k, v := range src {
		dst[k] = v
	}
}
```

c9s: ok looks good!

pkg/exchange/okex/convert_futures.go
c9s: can we define a struct for this key?

`type positionKey struct { symbol string ; side ; string }`

c9s (approved):

anywhy:
I had already joined Telegram before, but I rarely talk there.😄

PR #2216, opened 2025-09-18
UPDATE: [xalign] implement metrics tracking for balance deltas

dboyliao:
- tracking quantity/amount delta of assets
- Introduce new config `skipAlignOnAnyActiveTransfer`
  - If set true, the strategy will skip alignment if there is any active transfer of any currency

### DIFF
--- a/pkg/strategy/xalign/metrics.go
+++ b/pkg/strategy/xalign/metrics.go
@@ -5,6 +5,12 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
+type deltaGaugeKey struct {
+	currency string
+	side     string
+	typ      string
+}
+
 var quantityDeltaMetrics = promauto.NewGaugeVec(
 	prometheus.GaugeOpts{
 		Name: "xalign_balance_quantity_delta",
@@ -23,10 +29,17 @@ var amountDeltaMetrics = promauto.NewGaugeVec(
 
 func (s *Strategy) updateMetrics(currency string, side string, quantityDelta float64, amountDelta float64) {
 	var quantityGauge, amountGauge prometheus.Gauge
-	key := currency + "_" + side
-	qKey := key + "_quantity"
-	aKey := key + "_amount"
-	if g, ok := s.gaugeMetrics[qKey]; ok {
+	qKey := deltaGaugeKey{
+		currency: currency,
+		side:     side,
+		typ:      "quantity",
+	}
+	aKey := deltaGaugeKey{
+		currency: currency,
+		side:     side,
+		typ:      "amount",
+	}
+	if g, ok := s.deltaGaugesMap[qKey]; ok {
 		quantityGauge = g
 	} else {
 		quantityGauge = quantityDeltaMetrics.With(
@@ -35,9 +48,9 @@ func (s *Strategy) updateMetrics(currency string, side string, quantityDelta flo
 				"side":  side,
 			},
 		)
-		s.gaugeMetrics[qKey] = quantityGauge
+		s.deltaGaugesMap[qKey] = quantityGauge
 	}
-	if g, ok := s.gaugeMetrics[aKey]; ok {
+	if g, ok := s.deltaGaugesMap[aKey]; ok {
 		amountGauge = g
 	} else {
 		amountGauge = amountDeltaMetrics.With(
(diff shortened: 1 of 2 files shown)


pkg/strategy/xalign/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+	prometheus.GaugeOpts{
+		Name: "xalign_balance_amount_delta",
+		Help: "The balance delta of the quote amount of the currency",
+	},
+	[]string{"currency", "side"},
+)
+
+func (s *Strategy) updateMetrics(currency string, side string, quantityDelta float64, amountDelta float64) {
+	var quantityGauge, amountGauge prometheus.Gauge
+	key := currency + "_" + side
```
c9s: can we define a struct for the key?

type metricsKey struct { currency string; side string ; ... }

dboyliao: fixed.

pkg/strategy/xalign/metrics.go
```diff
@@ -0,0 +1,67 @@
...
+	prometheus.GaugeOpts{
+		Name: "xalign_balance_quantity_delta",
+		Help: "The balance delta of the asset",
+	},
+	[]string{"asset", "side"},
+)
+
+var amountDeltaMetrics = promauto.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Name: "xalign_balance_amount_delta",
```
c9s: use xalign_balance_value_delta ?  cuz we use "net asset value" terms like this for calculated asset values

pkg/strategy/xalign/metrics.go
```diff
@@ -0,0 +1,67 @@
+package xalign
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+type deltaGaugeKey struct {
+	asset string
```
c9s: call it "currency" to align with other metrics (we use currency in the balance metrics)

"asset" is more like you have more complex info collected for a specific token

pkg/strategy/xalign/metrics.go
```diff
@@ -0,0 +1,67 @@
...
+	var quantityGauge, amountGauge prometheus.Gauge
+	qKey := deltaGaugeKey{
+		asset: asset,
+		side:  side,
+		typ:   "quantity",
+	}
+	aKey := deltaGaugeKey{
+		asset: asset,
+		side:  side,
+		typ:   "amount",
```
c9s: would prefer "value" here

pkg/strategy/xalign/metrics.go
dboyliao: ok

pkg/strategy/xalign/metrics.go
dboyliao: done.

pkg/strategy/xalign/metrics.go
dboyliao: done.

pkg/strategy/xalign/strategy.go
```diff
@@ -495,51 +502,57 @@ func (s *Strategy) recordBalances(totalBalances types.BalanceMap, now time.Time)
...
+			bbgo.Notify("Found active %s deposit",
 				pendingDeposit.Asset,
 				pendingDeposit)
 		}
-
-		return false, nil
 	}
 
-	return true, nil
+	return activeTransferMap, nil
```
c9s: minor: can be just `activeTransfers`

c9s: since Map is type name

pkg/strategy/xalign/strategy.go
```diff
@@ -495,51 +502,57 @@ func (s *Strategy) recordBalances(totalBalances types.BalanceMap, now time.Time)
...
+		return nil, fmt.Errorf("unable to check active transfers (deposit): %w", err)
+	}
+
+	for _, pendingDeposit := range pendingDeposits {
+		log.Warnf("found active transfer (%f %s deposit)",
 			pendingDeposit.Amount.Float64(),
 			pendingDeposit.Asset)
 
 		s.resetFaultBalanceRecords(pendingDeposit.Asset)
+		activeTransferMap[pendingDeposit.Asset] = struct{}{}
```
c9s: maybe you just append the pendingDeposit into the map as values?

pkg/strategy/xalign/strategy.go
dboyliao: ok

pkg/strategy/xalign/strategy.go
dboyliao: In that case, the value type has to be `any`, since there can be withdraw or deposit.

dboyliao: fixed

PR #2234, opened 2025-09-29
FEATURE: [bybit] add futures position and leverage api

anywhy:


pkg/exchange/bybit/exchange.go
```diff
@@ -629,3 +659,27 @@ func (e *Exchange) GetAllFeeRates(ctx context.Context) (bybitapi.FeeRates, error
 func (e *Exchange) NewStream() types.Stream {
 	return NewStream(e.key, e.secret, e)
 }
+
+func (e *Exchange) GetCategory() bybitapi.Category {
```
c9s: make it private method here?

anywhy: done.

bailantaotao (approved):

PR #2238, opened 2025-09-30
FIX: [xmaker] minor fixes

c9s:
- do not recover order when code is 2016
- maker hedge interface consistent, pass hedgeDelta
- check last price and bind it to the market trade stream


pkg/exchange/max/exchange.go
```diff
@@ -652,6 +652,20 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
...
 			if responseErr.StatusCode >= 400 {
+				if responseErr.IsJSON() {
+					var maxError maxapi.ErrorResponse
+					if err := responseErr.DecodeJSON(&maxError); err != nil {
+						logger.WithError(err).Errorf("unable to decode max api error response: %s", responseErr.Response.Body)
+					}
+
+					switch maxError.Err.Code {
+					// 2006: The nonce has already been used by access key.
+					case 2016:
```
ycdesu: The code in the comment and case are different: 2006 vs 2016.

pkg/exchange/max/exchange.go
```diff
@@ -652,6 +652,20 @@ func (e *Exchange) SubmitOrder(ctx context.Context, order types.SubmitOrder) (cr
 			)
 
 			if responseErr.StatusCode >= 400 {
+				if responseErr.IsJSON() {
+					var maxError maxapi.ErrorResponse
+					if err := responseErr.DecodeJSON(&maxError); err != nil {
+						logger.WithError(err).Errorf("unable to decode max api error response: %s", responseErr.Response.Body)
```
ycdesu: Should we return here to avoid entering following switch case?

pkg/strategy/xmaker/strategy.go
```diff
@@ -1692,9 +1692,13 @@ func (s *Strategy) hedge(ctx context.Context, uncoveredPosition fixedpoint.Value
```
dboyliao, line 1692: Should add a return here?

pkg/exchange/max/exchange.go
c9s: You're right, let me do this

c9s: Fixed

pkg/strategy/xmaker/strategy.go
c9s: No, because spread maker will adjust the hedgeDelta

pkg/exchange/max/exchange.go
c9s: comment is updated

PR #2243, opened 2025-10-01
FIX: [xmaker] fix hedge market hedge behavior

c9s:
- call Cover before submitting the order
- trigger hedge when position delta channel received delta

### DIFF
--- a/pkg/strategy/xmaker/hedgeexecutor.go
+++ b/pkg/strategy/xmaker/hedgeexecutor.go
@@ -154,18 +154,29 @@ func (m *CounterpartyHedgeExecutor) clear(ctx context.Context) error {
 		return nil
 	}
 
-	if err := m.session.Exchange.CancelOrders(ctx, *m.hedgeOrder); err != nil {
-		m.logger.WithError(err).Errorf("failed to cancel order: %+v", m.hedgeOrder)
+	if o, ok := m.HedgeMarket.activeMakerOrders.Get(m.hedgeOrder.OrderID); ok {
+		m.hedgeOrder = &o
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/exchange/max/maxapi/restapi.go
```diff
@@ -62,16 +63,26 @@ func init() {
...
+	for maxTries := 10; maxTries > 0; maxTries-- {
 		select {
 		case <-ctx.Done():
 			logger.Error("unable to update server time offset due to timeout")
 			return
 		default:
 
 		}
+
 		serverTime, err := client.NewGetTimestampRequest().Do(context.Background())
```
bailantaotao: Is a use context timeout needed on Do?

pkg/exchange/max/maxapi/restapi.go
```diff
@@ -62,16 +63,26 @@ func init() {
...
+			var errResponse *requestgen.ErrResponse
+			if errors.Is(err, &net.DNSError{}) {
+				return
+			} else if errors.As(err, &errResponse) {
+				if errResponse.StatusCode >= 400 && errResponse.StatusCode < 500 {
+					return
+				}
+			}
+
 			logger.WithError(err).Error("unable to get max exchange server timestamp")
```
bailantaotao: log `maxTries` ?

c9s: added

pkg/exchange/max/maxapi/restapi.go
c9s: fixed, thanks!

PR #2247, opened 2025-10-03
FIX: [ProfitFixer] warning on no trades and update position notification string

dboyliao:


pkg/strategy/common/profit_fixer.go
```diff
@@ -103,8 +103,19 @@ func (f *ProfitFixer) Fix(
 	if err != nil {
 		return err
 	}
+	if len(allTrades) == 0 {
+		log.Warnf("no trades found for %s from %s to %s, skip profit fixing", symbol, since.String(), until.String())
```
c9s: use "no trades found between t1 and t2"?

pkg/strategy/common/profit_fixer.go
```diff
@@ -103,8 +103,19 @@ func (f *ProfitFixer) Fix(
...
+		return nil
+	}
 
-	return f.FixFromTrades(allTrades, stats, position)
+	err = f.FixFromTrades(allTrades, stats, position)
+	if err != nil {
+		return err
+	}
+	// update position open time and changed time
+	position.OpenedAt = since
```
c9s: should use the time from the first trade

pkg/strategy/common/profit_fixer.go
```diff
@@ -103,8 +103,19 @@ func (f *ProfitFixer) Fix(
...
+	position.ChangedAt = allTrades[len(allTrades)-1].Time.Time()
```
c9s: this field will be updated when a trade is added to the position?

pkg/strategy/common/profit_fixer.go
dboyliao: ok

pkg/strategy/common/profit_fixer.go
dboyliao: You're right.
Will remove it

pkg/strategy/common/profit_fixer.go
dboyliao: done.

c9s (approved):

gx578007 (approved):

PR #2249, opened 2025-10-07
FIX: [xmaker] handle NaN liquidity demand signal

dboyliao:


c9s (approved):

pkg/strategy/xmaker/signal/liqudity_demand.go
```diff
@@ -76,6 +76,9 @@ func (s *LiquidityDemandSignal) CalculateSignal(_ context.Context) (float64, err
 	}
 	score := (last - s.mean) / s.std
 	sig := math.Tanh(score) * 2 // scale to [-2, 2]
+	if math.IsNaN(sig) {
```
c9s, line 79: You need to handle Inf for score as well?

dboyliao: If score is ±Inf, the signal would be ±2, which is by design of the signal.

c9s: is score ±Inf expected?

c9s: which possibly means s.std is zero

c9s (approved):

PR #2252, opened 2025-10-08
FEATURE: [ProfitFixer] implement token fee map building for profit calculations

dboyliao:


pkg/strategy/common/profit_fixer.go
```diff
@@ -95,6 +105,113 @@ func (f *ProfitFixer) aggregateAllTrades(ctx context.Context, symbol string, sin
...
+	for _, trade := range trades {
+		// skip trade if fee currency is USD*
+		if strings.HasPrefix(trade.FeeCurrency, "USD") {
+			continue
+		}
+		// skip trade if we do not have market info
+		if _, ok := markets[trade.Exchange]; !ok {
+			continue
+		}
+		market := markets[trade.Exchange][trade.Symbol]
```
c9s, line 147: you can pass ExchangeSession into this method, so that you can just do session.Market(symbol) to get the market?

c9s (approved):

pkg/strategy/common/profit_fixer.go
dboyliao: In order to do that, I have to modify the method signature of `f.AddExchange()`.
Then I have to update all places where `f.AddExchange()` is applied.
With defensive programming in mind, I'd rather not to do it.
Instead, I just do type casting and resolve token fee prices whenever it's possible.

ycdesu (approved):

PR #2259, opened 2025-10-12
FEATURE: [exchange][hyperliquid] init hyperliquid exchange

anywhy:


pkg/exchange/hyperliquid/hyperapi/client.go
```diff
@@ -0,0 +1,287 @@
...
+			return nil, fmt.Errorf("expiresAfter cannot be negative: %d", *expiresAfter)
+		}
+		data = append(data, 0x00)
+		data = appendUint64(data, uint64(*expiresAfter))
+	}
+
+	return data, nil
+}
+
+func (c *Client) newAuthenticatedRequest(
```
c9s: perhaps just replace NewAuthenticatedRequest ?

pkg/exchange/hyperliquid/hyperapi/client.go
```diff
@@ -0,0 +1,287 @@
...
+
+	return SignatureResult{
+		R: hexutil.EncodeBig(r),
+		S: hexutil.EncodeBig(s),
+		V: v,
+	}, nil
+}
+
+// buildActionData constructs the data for action hashing
+func (c *Client) buildActionData(action any, nonce uint64, expiresAfter *int64) ([]byte, error) {
```
c9s: pass vaultAddress from the func parameters so that this can be a pure function

pkg/exchange/hyperliquid/hyperapi/client.go
```diff
@@ -0,0 +1,287 @@
...
+		source = "a" // mainnet
+	}
+
+	return map[string]any{
+		"source":       source,
+		"connectionId": hash,
+	}
+}
+
+func (c *Client) sign(typedData apitypes.TypedData) (SignatureResult, error) {
```
c9s: ditto, can be pure function if this does not depends on the Client object

pkg/exchange/hyperliquid/hyperapi/client.go
```diff
@@ -0,0 +1,287 @@
...
+				{Name: "chainId", Type: "uint256"},
+				{Name: "verifyingContract", Type: "address"},
+			},
+		},
+		PrimaryType: "Agent",
+		Message:     phantomAgent,
+	})
+}
+
+func (c *Client) PhantomAgent(hash []byte) map[string]any {
```
c9s, line 166: ditto, can be pure function

pkg/exchange/hyperliquid/hyperapi/client.go
```diff
@@ -0,0 +1,287 @@
...
+				payload["vaultAddress"] = nil
+			}
+		} else {
+			// For struct types, we need to use reflection or type assertion
+			// For now, assume it's not usdClassTransfer
+			payload["vaultAddress"] = c.vaultAddress
+		}
+	}
+
+	body, err := json.Marshal(payload)
```
c9s: can be `return json.Marshal(payload)` directly.

pkg/exchange/hyperliquid/hyperapi/client.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Add("Accept", "application/json")
+	return req, nil
+}
+
+func (c *Client) castPayload(action any) ([]byte, error) {
```
c9s: seems like this method is not just castPayload, maybe call it buildPayload?

pkg/exchange/hyperliquid/hyperapi/place_order_request.go
```diff
@@ -0,0 +1,62 @@
...
+
+	metaType InfoReqType `param:"type" default:"order" validValues:"order"`
+
+	orders []struct {
+		asset         string  `param:"a"`
+		isBuy         bool    `param:"b"`
+		size          string  `param:"s"`
+		price         string  `param:"p"`
+		reduceOnly    bool    `param:"r"`
+		clientOrderID *string `param:"c"`
```
c9s: I think this is not supported -- param tag inside the field with slice

pkg/exchange/hyperliquid/hyperapi/client.go
anywhy: done.

pkg/exchange/hyperliquid/hyperapi/client.go
anywhy: done.

pkg/exchange/hyperliquid/hyperapi/client.go
anywhy: done.

pkg/exchange/hyperliquid/hyperapi/client.go
anywhy: done.

pkg/exchange/hyperliquid/hyperapi/client.go
anywhy: done.

pkg/exchange/hyperliquid/hyperapi/place_order_request.go
anywhy: Yeah, it really doesn’t support that. Any idea how to deal with this kind of param?

c9s: wrap it with a struct, and add String() method to marshal the internal payload into json string? I am not sure

is it a nested json data in the request?

pkg/exchange/hyperliquid/hyperapi/get_account_balance_request.go
```diff
@@ -0,0 +1,37 @@
...
+	EntryNtl fixedpoint.Value `json:"entryNtl"`
+}
+
+type Account struct {
+	Balances []Balance `json:"balances"`
+}
+
+//go:generate GetRequest -url "/info" -type GetAccountBalanceRequest -responseDataType Account
+type GetAccountBalanceRequest struct {
+	client requestgen.APIClient
```
c9s, line 23: why not using requestgen.AuthenticatedAPIClient here?

anywhy: The Hyperliquid ```/info``` API doesn't require any signatures to access; it's just public data on the chain.

c9s:
@anywhy please ping me on telegram when you think it's ready

pkg/exchange/hyperliquid/hyperapi/get_account_balance_request.go
c9s: ok

PR #2450, opened 2026-04-09
refactor: remove obsolete //+build comments

stringsbuilder:
The plusbuild analyzer suggests a fix to remove obsolete build tags of the form:

        //+build linux,amd64

in files that also contain a Go 1.18-style tag such as:

        //go:build linux && amd64

(It does not check that the old and new tags are consistent;
that is the job of the 'buildtag' analyzer in the vet suite.)

### DIFF
--- a/pkg/backtest/assets_dummy.go
+++ b/pkg/backtest/assets_dummy.go
@@ -1,5 +1,4 @@
 //go:build !web
-// +build !web
 
 package backtest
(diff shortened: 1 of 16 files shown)


dboyliao (approved):

dboyliao:
Nice work!
Thanks @stringsbuilder.

PR #2458, opened 2026-04-17
FEATURE: [xfundingv2] implement market selection

dboyliao:


pkg/strategy/xfundingv2/market_data.go
```diff
@@ -0,0 +1,65 @@
...
+	Data []struct {
+		Asset          string           `json:"asset"`
+		CollateralRate fixedpoint.Value `json:"collateralRate"`
+		UpdateTime     int64            `json:"updateTime"`
+		Virtual        bool             `json:"virtual"`
+		DiscountFactor fixedpoint.Value `json:"discountFactor"`
+	} `json:"data"`
+}
+
+func queryPortfolioModeCollateralRates(ctx context.Context, assets []string) (map[string]fixedpoint.Value, error) {
```
c9s, line 35: put this method to binance.Exchange?

pkg/strategy/xfundingv2/market_selector.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Step 3: Filter candidates
+	candidates := make([]MarketCandidate, 0, len(indices))
+	// query the volume data
+	takerQuoteVolumes := make(map[string]fixedpoint.Value)
+	for _, symbol := range symbols {
+		takerVals, err := m.binanceFutures.QueryTakerBuySellVolumes(
```
c9s, line 86: you may hit the rate limit here if you get too many symbols

c9s (approved):

pkg/strategy/xfundingv2/market_data.go
dboyliao: Since this method is using the undocumented Binance API (`bapi`), which may subject to changes without notifications or such, it can be relatively unstable.
It seems to me it's not in the right position to put it into `binance.Exchange`.

pkg/strategy/xfundingv2/market_selector.go
dboyliao: The request weight for this API is 0:
https://developers.binance.com/docs/derivatives/usds-margined-futures/market-data/rest-api/Taker-BuySell-Volume

PR #2460, opened 2026-04-19
FEATURE: [binance] add the normal order ID for the algo order

anywhy:


pkg/exchange/binance/parse.go
```diff
@@ -1362,6 +1368,7 @@ func (e *AlgoOrderUpdateEvent) OrderFutures() (*types.Order, error) {
 			TimeInForce:   types.TimeInForce(e.AlgoOrder.TimeInForce),
 		},
 		OrderID:          uint64(e.AlgoOrder.AlgoId),
+		ActualOrderId:    actualOrderId,
```
c9s, line 1371: i think we have the field for the original order id

pkg/types/order.go
```diff
@@ -339,6 +339,9 @@ type Order struct {
 	// ExecutedQuantity is how much quantity has been executed
 	ExecutedQuantity fixedpoint.Value `json:"executedQuantity" db:"executed_quantity"`
 
+	// ActualOrderId is the order id of the executed normal order on exchange for algo order
+	ActualOrderId uint64 `json:"actualOrderId,omitempty" db:"-"`
```
c9s, line 343: well i think the db migration needs to be updated as well

pkg/exchange/binance/parse.go
anywhy: done.

pkg/types/order.go
anywhy: done.

migrations/sqlite3/20260421214100_orders_add_original_order_id.sql
```diff
@@ -0,0 +1,9 @@
+-- +up
+-- +begin
+ALTER TABLE orders ADD COLUMN original_order_id INTEGER NOT NULL DEFAULT 0;
```
c9s: sorry my fault, I found binance uses "actual", maybe we should use "actual" not "original" ,

anywhy: ok, updated

c9s (approved):

PR #2462, opened 2026-04-21
FEATURE: [xfundingv2] implement cost estimation for entry and exit

dboyliao:


pkg/strategy/xfundingv2/cost_tracker.go
```diff
@@ -9,3 +14,145 @@ func AnnualizedRate(fundingRate fixedpoint.Value, fundingIntervalHours int) fixe
 	numFundingPeriods := int64(annualFundingHours / fundingIntervalHours)
 	return fundingRate.Mul(fixedpoint.NewFromInt(numFundingPeriods))
 }
+
+type CostTracker struct {
+	market types.Market
+
+	// futuresPosition is the open position size in futures
+	// the open position size in spot is always the same but with opposite sign
+	futuresPosition fixedpoint.Value
```
c9s: not tracking spot position?

pkg/strategy/xfundingv2/cost_tracker.go
```diff
@@ -9,3 +14,145 @@ func AnnualizedRate(fundingRate fixedpoint.Value, fundingIntervalHours int) fixe
...
+
+	futuresOrderBook, spotOrderBook *types.StreamOrderBook
+
+	futuresFeeRate FeeRate
```
c9s: should apply from session.MakerFeeRate or session.TakerFeeRate

pkg/strategy/xfundingv2/cost_tracker.go
```diff
@@ -9,3 +14,145 @@ func AnnualizedRate(fundingRate fixedpoint.Value, fundingIntervalHours int) fixe
...
+type TransactionCost struct {
+	FuturesPosition fixedpoint.Value
+	SpotFee         fixedpoint.Value
+	FuturesFee      fixedpoint.Value
+	SpreadPnL       fixedpoint.Value
+}
+
+// EstimateEntryCost calculates the cost of entering the position
+// Note that the cost is based on the current order book, the cost is estimated by the time this function is called.
+func (c *CostTracker) EstimateEntryCost(isMaker bool) (TransactionCost, error) {
```
c9s: CostTracker should focus on the cost tracking? I mean the actual costs,

estimation could be a standalone function? cuz they need different params.

pkg/strategy/xfundingv2/strategy.go
```diff
@@ -21,6 +22,11 @@ func init() {
 	bbgo.RegisterStrategy(ID, &Strategy{})
 }
 
+type FeeRate struct {
```
c9s: we have defined a type called `ExchangeFee`

pkg/strategy/xfundingv2/cost_tracker.go
dboyliao: As noted in the comments, the spot position is assumed to be in the same size but with opposite sign.
For example, if `futurePosition` is `-3`, the spot size is then `3`.

pkg/strategy/xfundingv2/strategy.go
dboyliao: ok

pkg/strategy/xfundingv2/cost_tracker.go
dboyliao: it's removed

pkg/strategy/xfundingv2/strategy.go
dboyliao: done.

pkg/strategy/xfundingv2/cost_tracker.go
dboyliao: I was thinking that this struct will leverage the stream order book and keeps "track" of the update-to-date entry/exit cost.
That's why I named it `CostTracker`.

I will implement the actual PnL tracking in another struct, maybe named as `ArbitragePosition`, which will also keep track of the state of the position (opening, closing, ready, or closed)

How do you think?
Maybe we just need to rename `CostTracker` for something else?

pkg/strategy/xfundingv2/cost_tracker.go
dboyliao: I rename this field to `targetPosition`.

c9s (approved):

pkg/strategy/xfundingv2/cost_tracker.go
dboyliao: Rename it to `CostEstimator`.

PR #2464, opened 2026-04-23
FEATURE: [xfundingv2] update cost estimation types and live account updates

dboyliao:
- type renaming: `EstimatedCost`
- minor renaming for market selection

### DIFF
--- a/pkg/strategy/xfundingv2/market_selector.go
+++ b/pkg/strategy/xfundingv2/market_selector.go
@@ -36,7 +36,11 @@ type MarketSelectionConfig struct {
 type MarketCandidate struct {
 	Symbol string `json:"symbol"`
 
-	LastFundingRate        fixedpoint.Value `json:"lastFundingRate"`
+	LastFundingRate fixedpoint.Value `json:"lastFundingRate"`
+	// FundingIntervalHours is the funding interval in hours (e.g., 8 for 8h funding)
+	// It's normally 8 for the most Binance perpetuals, but some may have different intervals like 4h.
+	// See https://www.binance.com/en/futures/funding-history/perpetual/real-time-funding-rate
+	FundingIntervalHours   int              `json:"fundingIntervalHours"`
 	AnnualizedRate         fixedpoint.Value `json:"annualizedRate"`
 	TakerBuyQuoteVolume24h fixedpoint.Value `json:"takerBuyQuoteVolume24h"`
 	InRangeDepth           fixedpoint.Value `json:"inRangeDepth"`
@@ -46,17 +50,14 @@ type MarketCandidate struct {
 
 // MarketSelector selects the best market based on funding rate and liquidity
 type MarketSelector struct {
-	*MarketSelectionConfig
+	MarketSelectionConfig
 	binanceFutures *binance.Exchange
 	orderBooks     map[string]*types.StreamOrderBook
 	logger         logrus.FieldLogger
 }
 
 // NewMarketSelector creates a new MarketSelector
-func NewMarketSelector(config *MarketSelectionConfig, exchange *binance.Exchange, orderBooks map[string]*types.StreamOrderBook, logger logrus.FieldLogger) *MarketSelector {
-	if config == nil {
-		config = &MarketSelectionConfig{}
-	}
+func NewMarketSelector(config MarketSelectionConfig, exchange *binance.Exchange, orderBooks map[string]*types.StreamOrderBook, logger logrus.FieldLogger) *MarketSelector {
 	return &MarketSelector{
 		MarketSelectionConfig: config,
 		binanceFutures:        exchange,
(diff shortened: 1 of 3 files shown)


pkg/strategy/xfundingv2/strategy.go
```diff
@@ -157,31 +170,83 @@ func (s *Strategy) CrossRun(
...
-		s.arbitrage(trade.Time.Time())
+		s.tick(trade.Time.Time())
 	}))
 	s.futuresSession.MarketDataStream.OnMarketTrade(types.TradeWith(s.TickSymbol, func(trade types.Trade) {
-		s.arbitrage(trade.Time.Time())
+		s.tick(trade.Time.Time())
 	}))
 
+	// query initial balance snapshots via REST API
+	spotAccount, err := s.spotSession.Exchange.QueryAccount(ctx)
```
c9s: you can call s.spotSession.GetAccount() to get the realtime account info.

pkg/strategy/xfundingv2/strategy.go
```diff
@@ -157,31 +170,83 @@ func (s *Strategy) CrossRun(
...
+	if err != nil {
+		return fmt.Errorf("failed to query spot account: %w", err)
+	}
+	s.spotAccount = spotAccount
+
+	futuresAccount, err := s.futuresSession.Exchange.QueryAccount(ctx)
```
c9s: ditto, you don't have to maintain the account object by yourself,
in most of cases, session.GetAccount() is sufficient

pkg/strategy/xfundingv2/strategy.go
```diff
@@ -157,31 +170,83 @@ func (s *Strategy) CrossRun(
...
+		return fmt.Errorf("failed to query futures account balances: %w", err)
+	}
+	s.futuresAccount = futuresAccount
+
+	// listen for balance events
+	s.spotSession.UserDataStream.OnBalanceSnapshot(func(balances types.BalanceMap) {
+		s.mu.Lock()
+		defer s.mu.Unlock()
+		s.spotAccount.UpdateBalances(balances)
+	})
```
c9s: session is already doing this update for account.

pkg/strategy/xfundingv2/strategy.go
```diff
@@ -157,31 +170,83 @@ func (s *Strategy) CrossRun(
...
+	s.spotSession.UserDataStream.OnBalanceUpdate(func(update types.BalanceMap) {
+		s.mu.Lock()
+		defer s.mu.Unlock()
+		s.spotAccount.UpdateBalances(update)
+	})
```
c9s: ditto

pkg/strategy/xfundingv2/strategy.go
dboyliao: done.

pkg/strategy/xfundingv2/strategy.go
dboyliao: done.

pkg/strategy/xfundingv2/strategy.go
dboyliao: done.

pkg/strategy/xfundingv2/strategy.go
dboyliao: done.

pkg/strategy/xfundingv2/market_selector.go
c9s: this can be just value field, so every field will be set to zero

pkg/strategy/xfundingv2/strategy.go
```diff
@@ -157,31 +167,47 @@ func (s *Strategy) CrossRun(
 
 	// initialize depth books for model selection
 	for _, symbol := range candidateSymbols {
-		book := types.NewStreamBook(symbol, s.futuresSession.ExchangeName)
-		book.BindStream(s.futuresSession.MarketDataStream)
-		s.depthBooks[symbol] = book
+		futuresBook := types.NewStreamBook(symbol, s.futuresSession.ExchangeName)
+		futuresBook.BindStream(s.futuresSession.MarketDataStream)
```
c9s: this won't work because you need to send the subscription, you will get an empty book in this case

c9s (approved):

pkg/strategy/xfundingv2/strategy.go
dboyliao: I do subscribe to the book channel in `CrossSubscribe`.
I think it should work in this case?

pkg/strategy/xfundingv2/market_selector.go
dboyliao: done.

pkg/strategy/xfundingv2/strategy.go
dboyliao: fixed

PR #2467, opened 2026-04-24
FEATURE: [xfundingv2] most profitable market selection

dboyliao:
- Add `FuturesDirection`
    - when it is `"short"`, the strategy will only short on the futures.
    - when it is `"long"`, the strategy will only long on the futures.
- select market to start a new round based on profitability
    - the most profitable markets = the market that can break-even with the entry+exit cost fastest

### DIFF
--- a/pkg/strategy/xfundingv2/market_selector.go
+++ b/pkg/strategy/xfundingv2/market_selector.go
@@ -64,8 +64,8 @@ func NewMarketSelector(config MarketSelectionConfig, exchange *binance.Exchange,
 	}
 }
 
-// RankMarkets returns a ranked list of market candidates
-func (s *MarketSelector) RankMarkets(ctx context.Context, symbols []string) ([]MarketCandidate, error) {
+// SelectMarkets returns a list of market candidates that meet the selection criteria
+func (s *MarketSelector) SelectMarkets(ctx context.Context, symbols []string) ([]MarketCandidate, error) {
 	// Step 1: Query premium indexes for each symbol
 	indices, err := queryFundingRates(ctx, s.binanceFutures, s.logger, symbols)
 	if err != nil {
@@ -148,8 +148,6 @@ func (s *MarketSelector) RankMarkets(ctx context.Context, symbols []string) ([]M
 		})
 	}
 
-	// TODO: compute composite score and sort candidates
-
 	return candidates, nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/strategy/xfundingv2/market_selector.go
```diff
@@ -10,8 +11,27 @@ import (
...
 )
 
+type FuturesDirection string
+
+const (
+	FuturesDirectionLong  FuturesDirection = "long"
+	FuturesDirectionShort FuturesDirection = "short"
+)
+
+func (d FuturesDirection) Sign() int {
```
c9s: why not use types.PositionType?

pkg/strategy/xfundingv2/market_selector.go
```diff
@@ -23,29 +43,50 @@ type MarketSelectionConfig struct {
...
-	TakerBuyQuoteVolume24h fixedpoint.Value `json:"takerBuyQuoteVolume24h"`
-	InRangeDepth           fixedpoint.Value `json:"inRangeDepth"`
+	FundingIntervalHours   int
+	AnnualizedRate         fixedpoint.Value
+	TakerBuyQuoteVolume24h fixedpoint.Value
+	InRangeDepth           fixedpoint.Value
 
-	Score fixedpoint.Value `json:"score"` // maybe later
+	// MinHoldingInterval is the estimated minimum holding interval to break even for this market candidate
+	MinHoldingInterval time.Duration
```
c9s: Should be named as MinHoldingDuration?

pkg/strategy/xfundingv2/strategy.go
```diff
@@ -83,13 +99,11 @@ func (s *Strategy) Initialize() error {
 	}
 	s.futuresOrderBooks = make(map[string]*types.StreamOrderBook)
 	s.spotOrderBooks = make(map[string]*types.StreamOrderBook)
+	s.costEstimators = make(map[string]*CostEstimator)
```
c9s: costEstimator should be only used by one time, isn't it?

c9s (approved):

pkg/strategy/xfundingv2/strategy.go
dboyliao: The cost estimator has the access to the stream orderbooks to keep track of the latest order book status.
The caller can also use `.SetTargetPosition()` to dynamically update the target position and calculate the entry/exit cost accordingly.
Hence, my design is ok to reuse the cost estimator.

pkg/strategy/xfundingv2/market_selector.go
dboyliao: fixed.

pkg/strategy/xfundingv2/market_selector.go
dboyliao: done.

pkg/strategy/xfundingv2/strategy.go
```diff
@@ -293,3 +340,120 @@ func (s *Strategy) filterMarketCollateralRate(ctx context.Context, symbols []str
...
+	bestCandidate := &sortedCandidates[0]
+	targetPosition := targetPositions[bestCandidate.Symbol]
+	// set the estimated min holding interval for the selected candidate
+	numHoldingHours := breakevenIntervals[bestCandidate.Symbol].Int() * bestCandidate.FundingIntervalHours
+	bestCandidate.MinHoldingDuration = time.Duration(numHoldingHours) * time.Hour
+	return bestCandidate, targetPosition
+}
+
+func (s *Strategy) calculateMinHoldingIntervals(candidate MarketCandidate, bestPrice, targetPosition fixedpoint.Value) (fixedpoint.Value, error) {
+	costEstimator := s.costEstimators[candidate.Symbol]
```
c9s: what if costEstimator not found? do you have a fallback alllocator?

dboyliao: I refactor the cost estimator and now all symbols share the same estimator.
So it is no longer the issue the estimator would be missing.
It's guaranteed to be initialized when strategy starts.

pkg/strategy/xfundingv2/strategy.go
dboyliao: I remove the stream order books and the caller is now responsible to provide the order book snapshots for cost estimation.

c9s (approved):

PR #2470, opened 2026-04-29
STRATEGY: [xhedgegrid] basic skeleton with hedge simulator

c9s:


pkg/strategy/xhedgegrid/hedge.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// If timestamp is after all k-lines, use the last close price
+	if len(s.KLines) > 0 {
+		s.currentPrice = s.KLines[len(s.KLines)-1].Close
+		s.lastKLineIndex = len(s.KLines) - 1
+	}
+	return nil
+}
+
+func (s *HedgeSimulator) Hedge(uncovered, delta fixedpoint.Value) error {
```
dboyliao, line 127: Looks like `uncovered` is not used?

pkg/strategy/xhedgegrid/metrics.go
```diff
@@ -0,0 +1,213 @@
...
+			Help: "the price of filled grid order",
+		},
+		append([]string{
+			"exchange", // exchange name
+			"symbol",   // symbol of the market
+			"side",
+		}, extendedLabels...),
+	)
+
+	metricsNumOfOpenOrders = prometheus.NewGaugeVec(
```
dboyliao, line 173: I think you can use `promauto` so you won't need `registerMetrics`?

dboyliao: And move the `metricsRegistered` check into this function will do.

dboyliao (approved):

pkg/strategy/xhedgegrid/hedge.go
c9s: It's the Hedge method interface signature used in xmaker, so I wanted to align the method signature

pkg/strategy/xhedgegrid/metrics.go
c9s: Can do, this part is copied from grid2, we can adjust this later.

PR #2473, opened 2026-05-05
FEATURE: [xfundingv2] implement round opening/closing logics

dboyliao:
- A new round is opening if
    - there is a profitable market
    - there is **no** active round
    - the estimated break-even holding intervals is no greater than configured threshold (e.g 48 hrs)
- A round will be set to closing if
    - the round is expired: the total holding funding intervals is greater than estimated min holding intervals
    - the last funding rate is not favorable

### DIFF
--- a/pkg/strategy/xfundingv2/arb_round.go
+++ b/pkg/strategy/xfundingv2/arb_round.go
@@ -325,10 +325,6 @@ func (r *ArbitrageRound) FuturesSymbol() string {
 	return r.futuresWorker.Symbol()
 }
 
-func (r *ArbitrageRound) FuturesTargetPosition() fixedpoint.Value {
-	return r.spotWorker.FilledPosition().Neg()
-}
-
 func (r *ArbitrageRound) State() RoundState {
 	return r.state
 }
--- a/pkg/strategy/xfundingv2/arb_round.go
+++ b/pkg/strategy/xfundingv2/arb_round.go
@@ -294,6 +294,8 @@ func (r *ArbitrageRound) HandleSpotTrade(trade types.Trade, currentTime time.Tim
 		return
 	}
 
+	r.spotWorker.AddTrade(trade)
+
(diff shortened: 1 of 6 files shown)


pkg/strategy/xfundingv2/arb_round.go
```diff
@@ -79,6 +86,43 @@ func NewArbitrageRound(
 	}
 }
 
+func (r *ArbitrageRound) StartTime() time.Time {
+	return r.startTime
+}
+
+func (r *ArbitrageRound) IsExpired(currentTime time.Time) bool {
```
c9s: IsCloseable?

Expired is a bit confusing

pkg/strategy/xfundingv2/limiter.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+// Limit defines the rate of token replenishment (tokens per second).
+type Limit float64
+
+// Inf represents an infinite rate (never throttled).
+const Inf = Limit(math.MaxFloat64)
+
+// Every converts a minimum time interval between events to a Limit.
+func Every(interval time.Duration) Limit {
```
c9s: Why implement another Limiter here?

dboyliao: Because xfundingv2 is a tick-based strategy.
I need a tick-based rate limiter as well.

pkg/strategy/xfundingv2/arb_round.go
dboyliao: Ok, fixed.

pkg/strategy/xfundingv2/limiter.go
dboyliao: I just find out that I can use `.AllowN` of the `rate.Limiter`.
I'll replace it with `rate.Limiter`.

c9s (approved):

PR #2489, opened 2026-05-12
FIX: [xfundingv2] fix deadlock, TWAP taker price, transfer logic, and typos

dboyliao:
- Fix deadlock (`HasOrder`)
- Fix TWAP taker price calculation
- Make transfer retry tick non-blocking
- Fix market selection to use the ratio-adjusted quote balance for depth price lookup
- Fix transfer direction when futures has surplus fee asset (transfer out instead of no-op)
- Simplify holding interval calculation
- Return error instead of silently succeeding when estimated funding fee is zero
- Fix typos

### DIFF
--- a/pkg/strategy/xfundingv2/strategy.go
+++ b/pkg/strategy/xfundingv2/strategy.go
@@ -215,7 +215,7 @@ func (s *Strategy) CrossRun(
 		return fmt.Errorf("spot session %s not found", s.SpotSession)
 	}
 	if futuresEx, ok := s.futuresSession.Exchange.(types.FuturesExchange); !ok {
-		return fmt.Errorf("sessioin %s does not support futures", s.futuresSession.Name)
+		return fmt.Errorf("session %s does not support futures", s.futuresSession.Name)
 	} else if !futuresEx.GetFuturesSettings().IsFutures {
 		return fmt.Errorf("session %s is not configured for futures trading", s.futuresSession.Name)
 	}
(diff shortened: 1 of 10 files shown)


c9s (approved):

pkg/strategy/xfundingv2/strategy.go
```diff
@@ -904,6 +921,35 @@ func (s *Strategy) handleClosedRound(ctx context.Context, round *ArbitrageRound,
 		feeAvgCost := executor.Position().AverageCost
 		round.SetAvgFeeCost(s.FeeSymbol, feeAvgCost)
 	}
+	// clean up open orders if there is any
+	var spotOpenOrders, futuresOpenOrders []types.Order
+	for _, order := range round.spotWorker.Executor().AllOrders() {
```
c9s: You need activeOrderbook to get open orders

pkg/strategy/xfundingv2/strategy.go
```diff
@@ -904,6 +921,35 @@ func (s *Strategy) handleClosedRound(ctx context.Context, round *ArbitrageRound,
...
+		if !order.GetRemainingQuantity().IsZero() {
+			if order.IsFutures {
+				futuresOpenOrders = append(futuresOpenOrders, order)
+			} else {
+				spotOpenOrders = append(spotOpenOrders, order)
+			}
+		}
+	}
+	if len(spotOpenOrders) > 0 {
```
c9s, line 932: same here, AllOrders includes filled orders

pkg/strategy/xfundingv2/strategy.go
dboyliao: Fixed.
I add `.OpenOrders()` which reads the orders from the `ActiveMakerOrders()`.

pkg/strategy/xfundingv2/strategy.go
dboyliao: fixed.
We cancel orders returned by `.OpenOrders()`.

PR #2490, opened 2026-05-13
FIX: [xfundingv2] implement cleanup process for closed rounds and handle remaining positions

dboyliao:
- Non-blocking cleanup for closed rounds
  - Retry up to 3 times; send Slack notification on persistent failure
  - Block new rounds from opening on the same symbol until cleanup succeeds
- Close remaining futures positions safely after round ends
  - Implement `Cleanup` method using reduce-only market orders verified via `QueryPositionRisk`
  - Add `ReduceOnly` flag to TWAP orders when closing positions to prevent opening new positions
  - cancel open orders using general executor `GracefulCancelActiveOrderBook` method.
- Detect orphaned positions on startup
  - Query all futures position risks and verify each has a corresponding active round
  - Fail startup if mismatched positions are found
- Restore open orders into executor state on startup
  - Query open orders for each symbol and add them to `ActiveMakerOrders`
- Add Slack attachment support for round status reporting
  - `SlackAttachment` on `ArbitrageRound` and `CloseRoundTask` with color-coded state display
- Refactor: extract `getMidPrice` utility and `canOpenRound` helper

### DIFF
--- a/pkg/strategy/xfundingv2/arb_round.go
+++ b/pkg/strategy/xfundingv2/arb_round.go
@@ -246,21 +246,6 @@ func (r *ArbitrageRound) String() string {
 }
 
 func (r *ArbitrageRound) SlackAttachment() slack.Attachment {
-	var color string
-	switch r.syncState.State {
-	case RoundPending:
-		color = "#9C9494"
-	case RoundOpening:
-		color = "#6FCF97"
-	case RoundReady:
-		color = "#56CCF2"
-	case RoundClosing:
-		color = "#F2994A"
-	case RoundClosed:
-		color = "#B10202"
-	default:
-		color = "#9C9494"
-	}
 	title := fmt.Sprintf("Arbitrage Round %s (%s)", r.SpotSymbol(), r.syncState.State)
 	fields := []slack.AttachmentField{
 		{
@@ -323,9 +308,28 @@ func (r *ArbitrageRound) SlackAttachment() slack.Attachment {
 
 	return slack.Attachment{
 		Title:  title,
-		Color:  color,
+		Color:  r.stateColor(),
 		Fields: fields,
 	}
+}
+
+func (r *ArbitrageRound) stateColor() string {
+	var color string
+	switch r.syncState.State {
+	case RoundPending:
+		color = "#9C9494"
+	case RoundOpening:
+		color = "#6FCF97"
+	case RoundReady:
+		color = "#56CCF2"
+	case RoundClosing:
+		color = "#F2994A"
+	case RoundClosed:
+		color = "#B10202"
+	default:
+		color = "#9C9494"
+	}
+	return color
 }
 
 func (r *ArbitrageRound) TotalFundingIncome() fixedpoint.Value {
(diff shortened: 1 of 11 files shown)


pkg/strategy/xfundingv2/arb_round.go
```diff
@@ -242,6 +245,89 @@ func (r *ArbitrageRound) String() string {
...
+		color = "#6FCF97"
+	case RoundReady:
+		color = "#56CCF2"
+	case RoundClosing:
+		color = "#F2994A"
+	case RoundClosed:
+		color = "#B10202"
+	default:
+		color = "#9C9494"
+	}
```
c9s: pull out to "stateColor()"

pkg/strategy/xfundingv2/arb_round.go
```diff
@@ -523,12 +658,8 @@ func (r *ArbitrageRound) Tick(currentTime time.Time, spotOrderBook types.OrderBo
 
 	defer func() {
```
c9s: would suggest not to use defer for this case.

too many dependencies

pkg/strategy/xfundingv2/strategy.go
```diff
@@ -233,6 +248,28 @@ func (s *Strategy) CrossRun(
...
+	if err != nil {
+		return fmt.Errorf("failed to query position risk from futures session exchange: %w", err)
+	}
+	risksMap := make(map[string]types.PositionRisk)
+	for _, risk := range risks {
+		risksMap[risk.Symbol] = risk
+	}
+	// if there is any open position, it should have a corresponding active round which is loaded via LoadState.
+	// Otherwise, it is a mismatch and should raise an error to stop the strategy from running.
+	var misMatchSymbols []string
```
c9s: mismatchSymbols i think

pkg/strategy/xfundingv2/strategy.go
```diff
@@ -482,14 +547,37 @@ func (s *Strategy) tick(ctx context.Context, tickTime time.Time) {
 		s.transitRoundState(ctx, round, tickTime)
 	}
 
-	// remove closed active rounds
+	// enque closed active rounds
 	for _, round := range s.activeRounds {
 		if round.State() == RoundClosed {
```
c9s: would prefer NOT op here for this if statement so that you can reduce an indent

dboyliao: fixed.

pkg/strategy/xfundingv2/strategy.go
dboyliao: fixed.

pkg/strategy/xfundingv2/arb_round.go
dboyliao: done.

pkg/strategy/xfundingv2/arb_round.go
dboyliao: done.

c9s (approved):

PR #2494, opened 2026-05-15
FEATURE: [xfundingv2] add prometheus metrics for tracking round performance

dboyliao:
- `fundingRateMetrics`: the funding rate time series
- `annualizedFundingRateMetrics`: annualized funding rate
- `roundAnnualizedTriggerRateMetrics`: annualized rate of the triggering funding rate
- `roundHoldingIntervalMetrics`: the number of holding intervals of the round
- `roundNetPnLMetrics`: round net PnL
- `roundPositionFilledRatioMetrics`: round position filled ratio. it's designed to increasing up to 1 when the round is opening position and decreasing down to 0 when closing.

### DIFF
--- a/pkg/strategy/xfundingv2/metrics.go
+++ b/pkg/strategy/xfundingv2/metrics.go
@@ -1,0 +1,57 @@
+package xfundingv2
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// TODO: integrate `dynamic.InitializeConfigMetrics` for automatic metric labels generation based on the config struct fields.
+// see xmaker for example usage
+
+var fundingRateMetrics = promauto.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Name: "xfundingv2_funding_rate",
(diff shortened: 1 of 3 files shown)


pkg/strategy/xfundingv2/metrics.go
```diff
@@ -0,0 +1,54 @@
...
+	prometheus.GaugeOpts{
+		Name: "xfundingv2_round_holding_interval",
+		Help: "Holding interval of the arbitrage round in seconds",
+	},
+	[]string{"strategy", "strategyInstance", "symbol"},
+)
+
+var roundNetPnLMetrics = promauto.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Name: "xfundingv2_round_net_pnl",
```
c9s, line 45: how to distinguish the PnL of a specific round we want to see? or it's the latest round?

dboyliao: Since the `canOpenRound()` method guarantees that there is at most one round per symbol at any time.
We can use just symbol and the time to identify PnL for a specific round.

c9s (approved):

PR #2516, opened 2026-06-13
feat(dashboard): improve order flow widget rendering and add price aggregation

ttw225:
### Summary
Improves the order flow dashboard's readability, adds a configurable
price-aggregation unit, and scales the price axis proportionally.

### Changes
- **Price aggregation (`--aggregate` flag):** New `--aggregate` flag sets the
  minimum price unit for grouping trades into levels (e.g. `0.1`, `5`). Prices
  are floored to the nearest unit before being bucketed; `0` (default) disables
  aggregation.
- **Proportional price axis:** Each level's vertical position maps linearly to
  its price within the visible range, so the gap between rows mirrors the real
  price distance instead of being evenly spaced. Levels whose prices round to
  the same row are de-overlapped (kept one row apart) rather than dropped, so
  every visible level — the POC included — always gets a row.
- **Sorted levels:** Price levels are rendered in descending order (high → low),
  matching how a typical order book / footprint chart reads.
- **Display limit:** Caps the visible rows to at most 10 levels above and below
  the POC (point of control). A `hidden ↑N ↓N` indicator shows how many levels
  are windowed off each side; the window shrinks with terminal height and the
  POC stays visible even when the terminal is too short to fit every level.
- **Aligned columns:** Fixed-width price / delta / volume columns with padded
  strings so values line up cleanly regardless of length.
- **Thread safety:** `UpdateTrade` (fed from the trade stream) and `Draw`
  (render loop) share a mutex with a sorted snapshot, fixing concurrent
  map/slice access.

### Usage Examples

```bash
go run ./cmd/bbgo dashboard orderflow --exchange binance --symbol BTCUSDT
go run ./cmd/bbgo dashboard orderflow --exchange binance --symbol BTCUSDT --aggregate 0
go run ./cmd/bbgo dashboard orderflow --exchange binance --symbol BTCUSDT --aggregate 0.1
```

### Tests
- `TestAggregatePrice` — flooring behavior across units (zero/negative disable, sub-unit, integer, fractional steps).
- Widget tests — per-price/side aggregation, descending sort, POC windowing + hidden counts, empty draw, and a `-race` concurrency test for `UpdateTrade`/`Draw`.

### DIFF
--- a/pkg/cmd/dashboard.go
+++ b/pkg/cmd/dashboard.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/c9s/bbgo/pkg/cmd/widget"
 	"github.com/c9s/bbgo/pkg/exchange"
+	"github.com/c9s/bbgo/pkg/fixedpoint"
 	"github.com/c9s/bbgo/pkg/types"
 	ui "github.com/gizak/termui/v3"
 	"github.com/spf13/cobra"
@@ -19,6 +20,7 @@ func init() {
 	orderFlowCmd.MarkFlagRequired("exchange")
 	orderFlowCmd.Flags().String("symbol", "", "trading pair symbol")
 	orderFlowCmd.MarkFlagRequired("symbol")
+	orderFlowCmd.Flags().String("aggregate", "0", "aggregate price unit; 0 disables aggregation")
 	dashboardCmd.AddCommand(orderFlowCmd)
 	RootCmd.AddCommand(dashboardCmd)
 }
@@ -52,6 +54,15 @@ var orderFlowCmd = &cobra.Command{
 		if len(exName) == 0 || len(symbol) == 0 {
 			return errors.New("--exchange and --symbol are required")
 		}
+		aggregate, err := cmd.Flags().GetString("aggregate")
+		if err != nil {
+			return err
+		}
+		aggregateUnit, err := fixedpoint.NewFromString(aggregate)
+		if err != nil {
+			return err
+		}
+
 		// create a new public trade stream
 		exMin, err := exchange.NewWithEnvVarPrefix(types.ExchangeName(exName), strings.ToUpper(exName))
 		ex, ok := exMin.(types.Exchange)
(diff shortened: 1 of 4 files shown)


pkg/cmd/dashboard.go
```diff
@@ -19,6 +20,7 @@ func init() {
 	orderFlowCmd.MarkFlagRequired("exchange")
 	orderFlowCmd.Flags().String("symbol", "", "trading pair symbol")
 	orderFlowCmd.MarkFlagRequired("symbol")
+	orderFlowCmd.Flags().String("aggregate", "0", "aggregate price unit; 0 disables aggregation")
```
dboyliao: You may use `.Int()` instead of `.String()`.

pkg/cmd/dashboard.go
```diff
@@ -52,6 +54,15 @@ var orderFlowCmd = &cobra.Command{
 		if len(exName) == 0 || len(symbol) == 0 {
 			return errors.New("--exchange and --symbol are required")
 		}
+		aggregate, err := cmd.Flags().GetString("aggregate")
```
dboyliao: As the comment above, you can use `.GetInt("aggregate")` here.

pkg/cmd/dashboard.go
```diff
@@ -52,6 +54,15 @@ var orderFlowCmd = &cobra.Command{
...
+		if err != nil {
+			return err
+		}
+		aggregateUnit, err := fixedpoint.NewFromString(aggregate)
```
dboyliao: Following the comments above, you can use `fixedpoint.NewFromInt()`, which won't raise error making it more suitable for your use case.

pkg/cmd/dashboard.go
ttw225: Thanks for the review! Agreed that String + NewFromString introduced an unnecessary error path.

However, this feature is designed to support fractional units (0.1, 2.5, etc.) — the tests cover these cases — so switching to Int would silently break them. I went with a Float64 flag + fixedpoint.NewFromFloat() instead: same benefit of dropping the error branch, while keeping fractional aggregation intact.

pkg/cmd/dashboard.go
ttw225: Thanks, same as above.

pkg/cmd/dashboard.go
ttw225: Thanks, same as above.

pkg/cmd/dashboard.go
dboyliao: Oh, you are right.
Good catch.

dboyliao (approved):

PR #2524, opened 2026-06-23
FIX: [xfundingv2] bug fix and add position metrics

dboyliao:
- Fix deviation ratio calculation
    - prevent zero division when closing (spot filled may be zero)
    - add `TriggeredTargetPosition` to get the original target position for deviation ratio calculation
- Adjust the timing of futures position sync
   - sync futures position ONLY after a successful transfer
   - To prevent the risk of liquidation

### DIFF
--- a/pkg/strategy/xfundingv2/metrics.go
+++ b/pkg/strategy/xfundingv2/metrics.go
@@ -55,3 +55,11 @@ var roundPositionFilledRatioMetrics = promauto.NewGaugeVec(
 	},
 	[]string{"strategy_type", "strategy_id", "symbol", "accountType"},
 )
+
+var roundPositionMetrics = promauto.NewGaugeVec(
+	prometheus.GaugeOpts{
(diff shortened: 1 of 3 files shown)


c9s (approved):

pkg/strategy/xfundingv2/metrics.go
```diff
@@ -55,3 +55,11 @@ var roundPositionFilledRatioMetrics = promauto.NewGaugeVec(
 	},
 	[]string{"strategy_type", "strategy_id", "symbol", "accountType"},
 )
+
+var roundPositionMetrics = promauto.NewGaugeVec(
+	prometheus.GaugeOpts{
+		Name: "xfundingv2_round_position",
+		Help: "Spot position of the arbitrage round",
+	},
+	[]string{"strategy_type", "strategy_id", "symbol", "accountType"},
```
c9s, line 64: It's only for xfunding so you don't need "strategy_type" ?

PR #2536, opened 2026-06-29
FIX: [order] add selected columns for MySQL order query generation

dboyliao:


c9s (approved):
sounds like fix not refactor?

PR #2539, opened 2026-07-01
FIX: [xfundingv2] strategy state persistency and renaming

dboyliao:
- export persistent fields for the strategy struct
- minor method/variable renaming
- debug logging message
- dryrun mode

### DIFF
--- a/pkg/strategy/xfundingv2/strategy.go
+++ b/pkg/strategy/xfundingv2/strategy.go
@@ -334,7 +334,6 @@ func (s *Strategy) CrossRun(
 	if len(candidateSymbols) == 0 {
 		return errors.New("no candidate symbols after filtering")
 	}
-
 	// market checking and setup the general order executors
 	// NOTE: the executors should be created first before anything else to ensure the executors get updated first.
 	// The twap executors rely on the state of the general order executors to determine the filled quantity.
@@ -604,6 +603,12 @@ func (s *Strategy) CrossRun(
 			}
 		}
 	})
+
+	bbgo.Notify("✅ Strategy %s is up and running with %d candidate symbols: %v",
+		s.InstanceID(),
+		len(s.candidateSymbols),
+		s.candidateSymbols,
+	)
 
 	// Register shutdown handler to persist state
 	bbgo.OnShutdown(ctx, func(ctx context.Context, wg *sync.WaitGroup) {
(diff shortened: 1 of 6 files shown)


pkg/strategy/xfundingv2/strategy.go
```diff
@@ -187,10 +189,21 @@ func (s *Strategy) Defaults() error {
...
 func (s *Strategy) Initialize() error {
-	s.logger = logrus.WithFields(logrus.Fields{
-		"strategy":    ID,
-		"strategy_id": s.InstanceID(),
-	})
+	if os.Getenv("DEBUG_XFUNDINGV2") != "" {
+		logger := logrus.New()
+		logger.SetFormatter(logrus.StandardLogger().Formatter)
+		logger.SetOutput(logrus.StandardLogger().Out)
+		logger.SetLevel(logrus.DebugLevel)
```
c9s: can you add a newDebugLogger for this?

pkg/strategy/xfundingv2/strategy.go
```diff
@@ -187,10 +189,21 @@ func (s *Strategy) Defaults() error {
...
+		s.logger = logger.WithFields(logrus.Fields{
+			"strategy":    ID,
+			"strategy_id": s.InstanceID(),
+		})
```
c9s: Another pattern is to add a debug method that takes (msg string, va ...any)

more concise

pkg/strategy/xfundingv2/twap_order_executor.go
```diff
@@ -170,6 +176,12 @@ func (o *TWAPExecutor) PlaceOrder(quantity fixedpoint.Value, side types.SideType
 		return nil, fmt.Errorf("order is of dust quantity: %s", quantity)
 	}
 
+	if o.dryRun {
+		msg := fmt.Sprintf("[TWAPExecutor] dry run mode, would have submitted order: %+v", order)
+		o.logger.Info(msg)
+		return nil, errors.New(msg)
```
c9s, line 182: used Info but returned as error? i don't get it

i guess you need to apply Warnf

and return nil error?

dboyliao: I think dry-run is benign and will do no harm to anything.
A `info` message should be sufficient.
But at a second thought, maybe change it to `warn` is better.

pkg/strategy/xfundingv2/strategy.go
dboyliao: ok

pkg/strategy/xfundingv2/strategy.go
dboyliao: In that case, I need to enable the debug level for the logger.
If I change the log level without creating a new logger, it will change the global log level for `logrus`.
All other debug messages will be printed on the console which is not desirable.
Hence, I prefer to create a new logger than a debug method.

c9s (approved):

PR #2554, opened 2026-07-13
fix: supertrend Subscribe panics on disabled linearRegression config

rs-assistant:
## What

`setupIndicators()` nils `LinearRegression` when `window == 0` or the interval is empty, but `Subscribe()` dereferences `s.LinearRegression.Interval` unconditionally.

A config like:

```yaml
exchangeStrategies:
- on: binance
  supertrend:
    symbol: BTCUSDT
    interval: 1h
    window: 50
    linearRegression:
      window: 0
```

panics with a nil-interval subscribe (and a block without an interval panics inside `session.Subscribe` on the empty kline interval).

## Fix

Guard both panic paths in `Subscribe()`, consistent with the `setupIndicators()` disable semantics. Degenerate configs that set an interval with `window: 0` keep their historical (unused) subscription, so existing backtest replay inputs are unchanged.

## Tests

Added `TestStrategySubscribe_LinearRegressionNilSafe` covering nil block / window-zero block / valid block (separate interval) — the first two panicked before this fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018hiceFQWejhHyHnLayt1dE

### DIFF
--- a/pkg/strategy/supertrend/strategy.go
+++ b/pkg/strategy/supertrend/strategy.go
@@ -129,7 +129,14 @@ func (s *Strategy) Validate() error {
 
 func (s *Strategy) Subscribe(session *bbgo.ExchangeSession) {
 	session.Subscribe(types.KLineChannel, s.Symbol, types.SubscribeOptions{Interval: s.Interval})
-	session.Subscribe(types.KLineChannel, s.Symbol, types.SubscribeOptions{Interval: s.LinearRegression.Interval})
+	// LinearRegression is optional: setupIndicators() nils it for window==0 or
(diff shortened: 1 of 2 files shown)


c9s (approved):

c9s:
build failed

andycheng123:
@c9s This failure also occurs on main itself, roughly 20% of the time: a goroutine timing race triggers an extra SubmitOrder call that exceeds the gomock Times() limit.

andycheng123:
@c9s re-ran then failed test and passed

PR #2561, opened 2026-07-18
FEATURE: [position] PnL mode support

dboyliao:


pkg/types/position.go
```diff
@@ -754,6 +776,109 @@ func (p *Position) AddTrade(td Trade) (profit fixedpoint.Value, netProfit fixedp
...
+		p.Base = p.Base.Sub(quantity)
+		p.Quote = p.Quote.Add(quoteQuantity)
+	}
+	if p.IsDust(td.Price) {
+		p.OpenedAt = td.Time.Time()
+	}
+	return realizedPnL, realizedNetPnL, true
+}
+
+func (p *Position) UsePnLMode(mode PnLMode) {
```
c9s: can we expand the Mode to separate methods?

UseExcludeFeeFromCostMode()
SetExcludeFeeFromCostMode()

dboyliao: fixed

pkg/types/position.go
```diff
@@ -619,6 +631,23 @@ func (p *Position) calculateFeeInQuote(td Trade) fixedpoint.Value {
...
+	p.Lock()
+	if p.tradeHandler == nil {
+		switch p.PnLMode {
+		case PnLModeClassic:
+			p.tradeHandler = p.addTradeClassic
+		case PnLModeExcludeFeeFromCost:
+			p.tradeHandler = p.addTradeExcludeFeeFromCost
+		default:
+			p.tradeHandler = p.addTradeClassic
+		}
```
c9s: You can define a method that returns a function by the given PnL mode. 

and then you can call that method in this method

dboyliao: fixed.

c9s (approved):

PR #2584, opened 2026-08-06
FEATURE: [xmaker] support quote quantity mode

dboyliao:
- Add `EnableQuoteQuantity` config option to xmaker, when enabled:
    - allowing `quantityScale` to be interpreted as a **quote-currency** amount per layer (e.g. USDT) instead of base-currency quantity. When enabled, the scaled value is divided by the layer's quote price to derive the actual base-currency order quantity.
    - `quantity` will be interpreted as a **quote-currency** amount and the multiplier is applied to the quote amount. 
    - `depthQuantity` for the pricing is in **quote-currency** under this mode

### DIFF
--- a/pkg/strategy/xmaker/strategy.go
+++ b/pkg/strategy/xmaker/strategy.go
@@ -856,7 +856,7 @@ func (s *Strategy) getInitialLayerQuantity(i int, price fixedpoint.Value) (fixed
 	if s.QuantityMultiplier.Sign() > 0 && i > 0 {
 		q = fixedpoint.NewFromFloat(
 			q.Float64() * math.Pow(
-				s.QuantityMultiplier.Float64(), float64(i+1),
+				s.QuantityMultiplier.Float64(), float64(i),
 			),
 		)
 	}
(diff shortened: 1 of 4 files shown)


pkg/strategy/xmaker/strategy.go
```diff
@@ -202,6 +202,11 @@ type StrategyConfig struct {
 	// QuantityScale helps user to define the quantity by layer scale
 	QuantityScale *bbgo.LayerScale `json:"quantityScale,omitempty"`
 
+	// EnableQuoteQuantity enables the quantity calculation base on quote amount.
+	// when true, the `quantityScale` will be interpreted as quote currency instead of base currency
+	// the `quantity` will also be interpreted as quote currency
+	EnableQuoteQuantity bool `json:"enableQuoteQuantity"`
```
c9s: Maybe "Use" would be better than "Enable" in this case?

c9s (approved):

pkg/strategy/xmaker/strategy.go
dboyliao: ok

c9s (approved):

PR #2588, opened 2026-08-12
FIX: [binance] fix binance futures balance translation behavior

c9s:


c9s:
cc @anywhy  in case you are using it

dboyliao (approved):

PR #2597, opened 2026-08-20
FIX: [futures] fix Slack attachment error for futures positions

dboyliao:


pkg/types/account.go
```diff
@@ -365,14 +365,11 @@ func (p *FuturesPosition) SlackAttachment() slack.Attachment {
 	if p.PositionRisk == nil {
 		return slack.Attachment{
 			Color:    "#f48536",
-			Title:    fmt.Sprintf("%s Perp", p.Symbol),
+			Title:    fmt.Sprintf("%s Perp Position", p.Symbol),
```
c9s: use "FuturesPosition"?

pkg/types/account.go
```diff
@@ -391,7 +388,7 @@ func (p *FuturesPosition) SlackAttachment() slack.Attachment {
 		side = string(PositionClosed)
 	}
 
-	title := fmt.Sprintf("%s %s Perp %s %sX", p.Symbol, side, marginType, p.PositionRisk.Leverage)
+	title := fmt.Sprintf("%s %s Perp Position %s %sX", p.Symbol, side, marginType, p.PositionRisk.Leverage)
```
c9s: use "FuturesPosition"?

c9s (approved):

pkg/types/account.go
dboyliao: ok

pkg/types/account.go
dboyliao: ok

PR #2604, opened 2026-08-24
FIX: [xfundingv2] revise round prometheus metrics

dboyliao:
- unify round-specific metrics collection
    - Add `.RecordMetrics()` method


c9s:
check slack DM

c9s (approved):

PR #2621, opened 2026-09-01
FIX: [fixedpoint] correct Trunc to truncate toward zero for negative values

Tyagiquamar:
## Problem

`fixedpoint.Trunc` truncated negative values toward **negative infinity** instead of toward zero. For example `trunc(-1.5)` produced `-2` where the expected toward-zero semantics give `-1`. Since this type underpins price/quantity math, the floor-vs-trunc discrepancy is financially meaningful (it can silently inflate a negative quantity by one unit).

## Fix

Correct the negative path to truncate toward zero (1-line change in `pkg/fixedpoint/convert.go`).

## Test

`Test_Trunc` covers positive, negative, sub-unit, and boundary cases (`trunc(-1.5) = -1`, `trunc(-0.5) = 0`, `trunc(-123.456) = -123`, etc.). 5 negative cases failed before the fix; all 13 pass after.

`go test ./pkg/fixedpoint/ -run Trunc -count=1 -v`

Note: happy to adjust if any existing strategy intentionally relied on floor semantics for negatives. I did not find one, but it is worth a maintainer confirm.

### DIFF
--- a/pkg/fixedpoint/convert.go
+++ b/pkg/fixedpoint/convert.go
@@ -36,7 +36,7 @@ const (
 
 // Trunc returns the integer portion (truncating any fractional part)
 func (v Value) Trunc() Value {
-	return NewFromFloat(math.Floor(v.Float64()))
+	return NewFromFloat(math.Trunc(v.Float64()))
 }
 
 func (v Value) Round(r int, mode RoundingMode) Value {
--- a/pkg/fixedpoint/convert_test.go
+++ b/pkg/fixedpoint/convert_test.go
@@ -122,3 +122,31 @@ func Test_FormatString(t *testing.T) {
 		})
 	}
 }
+
+func Test_Trunc(t *testing.T) {
+	cases := []struct {
+		input    string
+		expected string
+	}{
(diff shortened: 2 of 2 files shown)


c9s (approved):

Tyagiquamar:
The failing `build (6.2, 1.25)` job is `TestSyntheticHedge_MarketOrderHedge` in `pkg/strategy/xmaker` — a mock timing flake (`Unexpected call to SubmitOrder ... has already been called the max number of times`), unrelated to this change. `xmaker` never calls `fixedpoint.Value.Trunc`; the only strategy callers of `Trunc` are grid and techsignal.

I reproduced the same failure locally on pristine `main` (`e44ec94`) under `-race -count 10`, so it is not introduced here. Could you re-run the job when convenient? Thanks!

Tyagiquamar:
@c9s The empty commit 34b09df2 successfully retriggered CI, but both workflows are currently held as action_required because they come from the fork. Could you approve the Go and golang-lint workflow runs, then merge the PR once they pass? Auto-merge is already enabled.

PR #2622, opened 2026-09-01
docs: correct cumulative PNL command labels

colinaumaty:
Correct cumulative PNL labels and the session initialization note across the four strategy draw helpers.


c9s (approved):

c9s:
thanks!

dboyliao (approved):

PR #2653, opened 2026-09-15
FEATURE: [xmaker] add no-borrowable cooldown

dboyliao:
#### Binance Exchange
- **Handle Insufficient Asset Error (`-3045`)**: Updated [`QueryMarginAssetMaxBorrowable`](file:///pkg/exchange/binance/exchange.go#L447-L455) to return `fixedpoint.Zero, nil` when Binance reports error code `-3045` (system does not have enough asset), treating it as having no borrowable capacity rather than failing.
- **Error Code Helper**: Added [`isBinanceErrorCode`](file:///pkg/exchange/binance/exchange.go#L465-L479) helper function to check API error codes and refactored [`configureFuturesOptions`](file:///pkg/exchange/binance/exchange.go#L1782-L1788) to use it.

#### XMaker Strategy
- **Borrowable Cooldown Support**: Added `NoBorrowableCooldown` to [`StrategyConfig`](file:///pkg/strategy/xmaker/strategy.go#L244-L248) and `cooldownAt` timestamp tracking to [`Strategy`](file:///pkg/strategy/xmaker/strategy.go#L363-L367).
- **Cooldown Gate in Quoting**: Added a check at the start of [`updateQuote`](file:///pkg/strategy/xmaker/strategy.go#L1014-L1028) to pause order quoting while the cooldown period is active.
- **Borrowability Check**: Added a check after canceling maker orders in [`updateQuote`](file:///pkg/strategy/xmaker/strategy.go#L1061-L1071) to verify hedge base asset borrowability and trigger `NoBorrowableCooldown` when unavailable.


pkg/strategy/xmaker/strategy.go
```diff
@@ -244,6 +244,8 @@ type StrategyConfig struct {
 	AuthTimeout types.Duration `json:"authTimeout,omitempty"`
 
 	CircuitBreaker *circuitbreaker.BasicCircuitBreaker `json:"circuitBreaker"`
+
+	NoBorrowableCooldown types.Duration `json:"noBorrowableCooldown,omitempty"`
```
c9s, line 248: use disableXxxxx

c9s (approved):

pkg/strategy/xmaker/strategy.go
dboyliao: It's the cooldown period when the base asset is not borrowable.
I think it's better to use current naming.

PR #2662, opened 2026-09-21
feat(types): add exchange cancel-replace contract

zenixls2:
## Summary

- Add `CancelReplace` to the shared exchange trade contract.
- Add generated mock support and shared cancel-then-create fallback semantics.
- Use Binance Spot native cancel-replace with strict partial-result validation.
- Add fallback implementations for exchanges without a native endpoint and for backtest.
- Update the MAX withdrawal example to the current `maxapi/v3` client so the latest upstream tree compiles.

The fallback path is explicitly non-atomic:

```text
cancel old order -> create replacement order
```

## Validation

- `go test ./... -run '^$' -count=1`
- `go test ./pkg/types/... ./pkg/exchange/... ./pkg/backtest -count=1`
- `go test -race ./pkg/types ./pkg/exchange/binance ./pkg/backtest -count=1`
- `gofmt` and `git diff --check`

All passed on top of `c9s/bbgo` latest `main` at `4a4a18a08897579d157c8fd3b412309cd4954852`.


c9s (approved):

pkg/backtest/cancel_replace.go
```diff
@@ -0,0 +1,16 @@
...
+
+	"github.com/c9s/bbgo/pkg/types"
+)
+
+var _ types.ExchangeTradeService = (*Exchange)(nil)
+
+// CancelReplace uses the same cancel-then-create semantics as a venue without
+// a native atomic cancel-replace endpoint.
+func (e *Exchange) CancelReplace(ctx context.Context, cancelReplaceMode types.CancelReplaceModeType, order types.Order) (*types.Order, error) {
+	_ = cancelReplaceMode
```
dboyliao, line 14: What is this line for?

dboyliao (approved):

